Issue #105, opened 2013-04-10
Caja 1.6.x: Rendering of Desktop Icon Text is broken in case "No Background" is configured


diva4:
After upgrade from Mate 1.4 to Mate 1.6 the text below the Desktop Icons are rendered with a wrong colour after login:
![Icon-text-wrong](https://f.cloud.github.com/assets/1687892/364759/75bc32f6-a228-11e2-9131-f2f571da1f4c.png)
The rendering gets corrected by selecting the Desktop and pressing F5 or by invoking "mate-appearance-properties -p background":
![Icon-text-correct](https://f.cloud.github.com/assets/1687892/364765/b4a609b0-a228-11e2-9bda-8b5f2f74f1f0.png)
This issue occurs only if "no background" is configured.
It looks like a similiar issue as #81

PR #285, opened 2014-04-28
Quick fix for #105

bl0ckeduser:
Enclosed is a trivial one-line change that successfully fixes bug #105.

(A more sophisticated fix might be possible by inspecting this one.
At least, I have figured out exactly where the problem manifests itself.)

Here is a "before" screenshot with incorrect font color on logon:

And here is an "after" (after the fix) screenshot with correct font color:


raveit65:
This fix causes a new issue.
https://github.com/mate-desktop/caja/issues/438

PR #309, opened 2014-08-18
fixed: now caja can make use of the default manual layout option

tsowa:
fixed: now caja can make use of the default manual layout option
(which can be set by using gsettings or dconf editor)

fixed: caja crashes when the default manual layout option was either:
'directory' or 'atime'

(patch copied from 1.8 branch)


NiceandGently:
Can you please say which gsettings key you mean?

tsowa:
I have described it in the last post:
http://forums.mate-desktop.org/viewtopic.php?f=2&t=3093

lukefromdc:
In GTK 3.22, in fact every time I've played with the manual layout option in a GTK3 build of Caja it has worked as-is.

raveit65:
I can't reproduce it, so closing both PRs.

PR #366, opened 2014-12-31
fixed: Adjust Default _zoom level always get the same old metadata value in icon/compact and list view.

lenky0401:
Always get the old metadata value update the metadata value, lead to adjusting default zoom level failure.


lenky0401:
You are right, thanks.

PR #370, opened 2015-01-10
use mate_gdk_spawn_command_line_on_screen like other MATE apps do

monsta:


src/caja-window-menus.c
```diff
@@ -576,31 +580,9 @@ action_caja_manual_callback (GtkAction *action,
...
-#if GTK_CHECK_VERSION (2, 24, 0)
-        gdk_spawn_command_line_on_screen(gtk_window_get_screen(GTK_WINDOW(window)), "mate-help", &error);
-#else
-        g_spawn_command_line_async("mate-help", &error);
-#endif
-#endif
+        mate_gdk_spawn_command_line_on_screen(gtk_window_get_screen(GTK_WINDOW(window)),
+                                              "mate-help",
+                                              &error);
     }
```
infirit: mate-help is no more for some time now. The intention I believe is to show the general help when we are on the desktop.

We should replace the whole thing with a simple ```gtk_show_uri ``` like the else part but then pointing to ```"help:mate-user-guide"```. Can you split this and put this in a new commit please?

monsta: Well, the ```else``` part already points to ```help:mate-user-guide/goscaja-1``` - should I change something there?

infirit: The else part is ok, it points to the __caja__ user-guide, just the first part needs an update to point to the user-guide index :smile:

monsta:
Not sure if I got it right, but I've made it to show mate-user-gude in both cases. It still gives error on that (help isn't available) even though I have mate-user-guide from git branch installed. Am I missing something?

infirit:
mate-user-guide does not have a caja part yet. The following should get you the index ```yelp "help:mate-user-guide"```

monsta:
LOL, I didn't have yelp installed! :) Maybe the error message should give a hint about that?
Once I installed yelp, it showed me the index when I've pressed F1 on the desktop.
As for the caja part of the guide, indeed, it gives an error - no such page.

PR #385, opened 2015-02-19
Make sorting by display name case-insensitive

Zaffy:
Hey guys,

I find very annoying that filenames with Capitals letters are sorted before the small ones.  
For example when you have to go and find something starting with 'W', 'w', you don't actually know if it'll be in the middle (capital case) or at the end.

This small change should deal with the issue. It's just a "feature request", although I have already made it - I think it's a better way than a bug report.  

Merge it or not, it's up to you.

### DIFF
--- a/libcaja-private/caja-file.c
+++ b/libcaja-private/caja-file.c
@@ -2876,7 +2876,7 @@ compare_by_display_name (CajaFile *file_1, CajaFile *file_2)
 	} else {
 		key_1 = caja_file_peek_display_name_collation_key (file_1);
 		key_2 = caja_file_peek_display_name_collation_key (file_2);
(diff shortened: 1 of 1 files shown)


flexiondotorg:
@stefano-k Please review this for inclusion in MATE 1.10

Zaffy:
Bump...

monsta:
I'm not sure what is fixed here...
I have this kind of sorting in Caja:

caja sorting

It looks pretty much case-insensitive to me.

Zaffy:
Hmm... it works for me now too! I don't know what happened, but I justkept myself up-to-date... Meybe it could be due to a broken locale?

monsta:
@flexiondotorg: does it work for you?

Issue #382, opened 2015-02-11
GTK3: missing css class 'view' in FMIconContainer


NiceandGently:
Caja has a missing css class 'view' in FMIconContainer which make theming very weird.
Screenshot without theming caja view with GreenLaguna.
![fmiconcontainer1](https://cloud.githubusercontent.com/assets/961604/6155918/26e17efa-b236-11e4-836e-c44ad17d8491.png)
Here you see a white background which comes from theme basecolor and in gtk-inspector you see that FMIconContainer has no css class 'view'.
Now i add the missing class with gtk-inspector.
![fmiconcontainer2](https://cloud.githubusercontent.com/assets/961604/6155995/90266e16-b236-11e4-972a-6b045dc6ce7a.png)
And write a  theme entry in gtk-inspector.
![fmiconcontainer3](https://cloud.githubusercontent.com/assets/961604/6156035/de318424-b236-11e4-8533-d824f0eefccb.png)
Et voila, caja view has a colored background.
Note, nautilus or nemo has this css class.
Currently i worked around with that entry
```
CajaView .entry {
	background-color: shade(@base_color, 1.06)
}
```
But this is complete wrong and makes it hard for other theme artist to write a theme for caja.
Hope this is comprehensible and if you want to reproduce it disable the entry 'CajaView .entry' in our themes.

PR #392, opened 2015-03-04
GTK3: add missing style class view to FMIconContainer

NiceandGently:
fixes https://github.com/mate-desktop/caja/issues/382


NiceandGently:
taken from nemo

Issue #398, opened 2015-04-03
caja automounts USB flash drives and CD/DVD drives while session is locked


sunweaver:
This issue reflects what has been reported to DebianBug#781608 [1].

Quoting DebBug#781608:

"""
To avoid auto-run drive-by attacks by a physically proximate attacker on
the system from USB auto-mounting screen is locked, the desktop should
delay automounting until the screen is unlocked (to not interfere with
the case of sitting back down at your system, plugging in a device,
and then unlocking your screen).

This is similar to how gnome-keyring flushes all keys the when locking
the screen.

http://www.net-security.org/secworld.php?id=10544

This was fixed in Gnome upstream and in Ubuntu:

https://bugs.launchpad.net/ubuntu/+source/nautilus/+bug/714958
https://bugs.launchpad.net/ubuntu/+source/nautilus/+bug/724285

The attached patch likely needs the dbus names changed to, e.g.,
"org.mate.ScreenSaver".

Thanks!

-Kees
"""

A patch has been provided in Debian/Ubuntu for Nautilus. I personally think this issue is worth assigning a CVE.

A side-effect of this is:

  o login as user A
  o insert a movie DVD
  o change user to user B
  o try to play that DVD

What you will experience is ons and offs when accessing the DVD, because both session (of user A and user B) try to access that DVD medium. The DVD drive motor will go on and off, it is merely impossible to eject that DVD medium, etc.

The Debian bug ships a patch from nautilus. I took a look at it, but it seems that it interferes with Caja's module concept and must be rewritten. I'd be happy if someone from upstream could pick up this task soon.

[1] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=781608

sunweaver:
@stefano-k, @monsta: I have just requested a CVE ID for this issue as recommended by Moritz Mühlenhoff from the Debian Security team.

monsta:
I see the second part ("side effect") is already mentioned in https://github.com/mate-desktop/caja/issues/362...

thassan:
perfect.. this was annoying as well

PR #400, opened 2015-04-06
Fix automount on locked screen

tarakbumba:
This pull request should fix https://github.com/mate-desktop/caja/issues/398 without causing g_bus_unwatch_name_assertion. Merely based on nautilus commits. Please review. Thanks

### DIFF
--- a/src/caja-application.c
+++ b/src/caja-application.c
@@ -120,7 +120,10 @@ static void     mount_added_callback              (GVolumeMonitor            *mo
 static void     volume_added_callback              (GVolumeMonitor           *monitor,
         GVolume                  *volume,
         CajaApplication      *application);
-static void     drive_connected_callback           (GVolumeMonitor           *monitor,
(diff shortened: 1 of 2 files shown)


sunweaver:
@monsta, @stefano-k: this fix should get into 1.10 ASAP IMHO. A 1.8.x backport patch will also be highly welcome! As the issue is really bugging people here, I am looking forward to providing a fix for this soon in Debian. But not without MATE upstream consent. THANKS!

sunweaver:
@tarakbumba: thanks for working on thist!

monsta:
@tarakbumba: thanks. But doesn't it also forbid automount when just a screensaver is active (no screen locking, only screensaver)?

monsta:
Ok, looks like it delays automounting even when screensaver is active without the lock.
I can't come up with any possible consequences of that, so I think it's ok to merge.

PR #401, opened 2015-04-07
U fixes for file management preferences

NiceandGently:
Setting a default windows size fixes that the preferences window display always to huge with GTK3.
Expanding list colums avoid to use a scrollbar for a few entries.


flexiondotorg:
Tested on Arch Lnux with MATE 1.10 built against GTK2 and GTK3.

PR #409, opened 2015-04-25
No daemon mode on other desktop environments

tarakbumba:
Caja should exit when last window closed on non MATE desktops.
Otherwise it keeps running altough its daemon mode is not needed
KDE, XFCE and others don't need caja's automount features so
Caja should behave like other file managers out there on these
environments. Please review this pull request.

### DIFF
--- a/src/caja-main.c
+++ b/src/caja-main.c
@@ -504,7 +504,11 @@ main (int argc, char *argv[])
     caja_global_preferences_init ();
 
     /* exit_with_last_window being FALSE, caja can run without window. */
-    exit_with_last_window = g_settings_get_boolean (caja_preferences, CAJA_PREFERENCES_EXIT_WITH_LAST_WINDOW);
+    /* But we should need this still set as TRUE on other desktop environments. Otherwise caja wont quit */
+    if (g_strcmp0 (g_getenv ("XDG_CURRENT_DESKTOP"), "MATE") == 0)
+    {
+        exit_with_last_window = g_settings_get_boolean (caja_preferences, CAJA_PREFERENCES_EXIT_WITH_LAST_WINDOW);
+    }
 
     application = NULL;
 
--- a/src/caja-main.c
+++ b/src/caja-main.c
@@ -505,7 +505,9 @@ main (int argc, char *argv[])
 
     /* exit_with_last_window being FALSE, caja can run without window. */
     /* But we should need this still set as TRUE on other desktop environments. Otherwise caja wont quit */
-    if (g_strcmp0 (g_getenv ("XDG_CURRENT_DESKTOP"), "MATE") == 0)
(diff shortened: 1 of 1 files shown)


monsta:
Does automount feature also work when caja doesn't manage the desktop (```caja --no-desktop```)?

tarakbumba:
@monsta, this pull request only disables to reading value for "exit-with-last-window" setting in org.mate.caja.preferences for other DEs. As you can see that this setting is already true elsewhere in other parts of the code. Our default gsettings value is set to "false" which is correct for MATE. But this should be set to "true" on other DEs. 

So, being triggered with "--no-desktop" option or not and managing/not managing automounts; caja doesn't exist if exit-with-last-window is not set to true.

monsta:
I meant, maybe it's not needed when caja runs with ```--no-desktop```? It should not interfere with other file managers in this case.

tarakbumba:
Altough caja runs with the "--no-desktop" argument, it still doesn't exit when last window is closed and it must be killed. This is the case with other desktops. Of course caja doesn't manage automounts but it acts like that.

NiceandGently:
@tarakbumba 
PR needs to be adjusted to use XDG_CURRENT_DESKTOP and XDG_SESSION_DESKTOP and DESKTOP_SESSION
Ie. fedora use DESKTOP_SESSION

tarakbumba:
I merely take this idea from https://github.com/mate-desktop/caja/commit/23b75dadd35f039a8cd93111528a8a66dde8888e and that commit also uses only XDG_CURRENT_DESKTOP ?

monsta:
Yes, @flexiondotorg pointed out that it should be corrected as well.

NiceandGently:
> Altough caja runs with the "--no-desktop" argument, it still doesn't exit when last window is closed and it must be killed. This is the case with other desktops. Of course caja doesn't manage automounts but it acts like that.

Same happends if you run caja as root from a terminal, caja won't close until you close the terminal.

tarakbumba:
> Same happends if you run caja as root from a terminal, caja won't close until you close the terminal.
> Or when you use the gksu extension.

I can not confirm this if this pull request is applied to caja. Here, with patched caja it cleanly exits here; if it is started via gksu-polkit (Mageia doesn't use gksu but gksu-polkit)

NiceandGently:
Cool, the updated PR works like a charm, caja closed now fine if run as root.
Enough reason for me to use it for fedora.
@monsta 
@flexiondotorg 
Please test and merge

monsta:
Weird... works when I run ```gksudo caja``` but not with ```gksu caja``` :confused:

monsta:
Great. ```XDG_CURRENT_DESKTOP=MATE``` with gksu.

monsta:
Ok, I made it work for me: https://github.com/mate-desktop/caja/pull/439
Please test it guys.

PR #428, opened 2015-06-18
GTK3: editable-label : only clear the selection if we change to insensitive

NiceandGently:
this fixes non selected text if renaming a file

taken from:
https://git.gnome.org/browse/nautilus/commit/?h=gnome-3-0&id=d2c3410


NiceandGently:
works here with gtk2

monsta:
@flexiondotorg: works fine in Debian Testing with both GTK+ 2 and 3. When I press F2, the filename gets selected as it should.

monsta:
ah wait, I forgot I can merge here myself... need more coffee I guess :D

Issue #423, opened 2015-06-08
[for discussion] drag-n-drop behavior in bookmarks section of sidebar: what to do?


monsta:
@stefano-k @flexiondotorg @NiceandGently @sunweaver @willysr @infirit 

Some history:

1. Some time ago, drag-n-drop of files to the bookmarked folders didn't work (https://github.com/mate-desktop/caja/issues/345).
2. Later it has been fixed by @NiceandGently in https://github.com/mate-desktop/caja/commit/910b9141ac634a86d8f53fda534e33787a160efb.
3. I've filed [this Debian bug report](https://bugs.debian.org/786395), asking to make a patch for Debian out of that commit.
4. During the patch testing, @sunweaver found out that adding new folders via drag-n-drop to bookmarks list does not work anymore.

The choice:

- if we revert https://github.com/mate-desktop/caja/commit/910b9141ac634a86d8f53fda534e33787a160efb, we'll be able to add new folders to bookmarks via drag-n-drop. Like it's currently done in Caja 1.8: drag some folder to the empty space in bookmarks section, then a horizontal line appears, then you can drop the folder there, and it will become a new bookmark. But we'll also reintroduce https://github.com/mate-desktop/caja/issues/345 back.
- if we leave it as is, we'll be able to drag-n-drop files into some existing bookmarked folder, but won't be able to add new folders to bookmarks via drag-n-drop.

Current situation:

The patch isn't accepted in Debian (and Ubuntu), and we have Caja 1.8 there, so currently https://github.com/mate-desktop/caja/issues/345 affects these distros.
Once 1.10 makes it there, https://github.com/mate-desktop/caja/issues/345 will be fixed, but the ability to add new folders to bookmarks via drag-n-drop will be lost.

Upstream situation:

@NiceandGently found out that Nautilus now has both behaviors - that is, it allows both drag-n-drop of files into the existing bookmarks and adding new folders to bookmarks. It shows which action will be done so the user could distinguish between them and choose the one he wants.
It had been implemented somewhere between 3.8 and 3.10.

**Bad news**: during the development of Nautilus 3.10, they made their own [GtkPlacesSidebar widget](https://developer.gnome.org/gtk3/stable/GtkPlacesSidebar.html) and switched Nautilus to it. So we don't have any upstream code to cherry-pick - we need to analyze GtkPlacesSidebar code and try to implement the same logic in Caja to fix our issue. This is not easily and quickly done...

[...]

Issue #345, opened 2014-11-10
drag and drop to bookmarks in side pane does not work


denispr:
dragging and dropping files to devices in side pane does work, but not to bookmarked folders.
(caja 1.8.1 on Linux Mint 17)

NiceandGently:
https://github.com/mate-desktop/caja/commit/910b9141ac634a86d8f53fda534e33787a160efb

monsta:
@stefano-k: how do you think, should that commit be ported to 1.8?

monsta:
Cherry-picked to 1.8 branch: https://github.com/mate-desktop/caja/commit/ec0c8f4143d900db6968b5a73a45a08a925b6f06

flexiondotorg:
The https://github.com/mate-desktop/caja/commit/910b9141ac634a86d8f53fda534e33787a160efb has been reverted for the 1.10.1 release because it introduces a usability regression. The MATE team need some time to determine the best solution to this issue.

PR #429, opened 2015-06-18
places-sidebar: enable both creating bookmarks and dropping files

monsta:
- drag a folder to the top or bottom of the bookmark list to add
a new bookmark. a horizontal line will be shown and the mouse
pointer will have a '+' at its right side.
- if there are no bookmarks (and that section isn't shown), drag
a folder over the network section's heading. this will add a new
bookmark and make the bookmarks list visible.
- drag a file or a folder over some existing bookmark to move it
into the bookmarked folder. the mouse pointer will have an arrow
at its right side.
- if you want to copy a file or a folder there instead, perform
the same operation but hold ctrl during that. the mouse pointer
will have a '+' at its right side. no horizontal line will be
shown, so you will be able to distinguish this operation from
adding a new bookmark. :)

adapted from https://github.com/linuxmint/nemo/commit/f18c4dd842be0998cbe6a519d9904bfd9993ca3e
(thanks to @mtwebster)

fixes https://github.com/mate-desktop/caja/issues/345
closes https://github.com/mate-desktop/caja/issues/423

@stefano-k @flexiondotorg @NiceandGently @sunweaver @willysr @infirit @denispr
please test it, if all goes well, I'll merge it.

### DIFF
--- a/src/caja-places-sidebar.c
+++ b/src/caja-places-sidebar.c
@@ -1134,7 +1134,6 @@ compute_drop_position (GtkTreeView *tree_view,
         /* never drop on headings, but special case the bookmarks heading,
          * so we can drop bookmarks in between it and the first item.
          */
-
         gtk_tree_path_free (*path);
         *path = NULL;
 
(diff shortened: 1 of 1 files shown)


NiceandGently:
works well here with both gtk+ versions.
+1

PR #431, opened 2015-06-21
icon-container: reset the double click counter after a double click

NiceandGently:
If we don't do this, we ignore any other double click event that happen
during the next 'gtk-double-click-time' interval after the first double
click.

taken from:
https://git.gnome.org/browse/nautilus/commit/?id=a8539a6


NiceandGently:
Note, i noticed this issue with gtk3 build

Issue #406, opened 2015-04-16
caja in compact view, screws focus with long name files 


freed00m:
Demo: http://youtu.be/lUqo938Zjgg

Minor annoyance, hope it'll be fixed someday.

monsta:
I've found an old [Nautilus bug report](https://bugzilla.gnome.org/show_bug.cgi?id=585511) about this issue (which we obviously inherited from it).
Unfortunately, gnome devs fixed the issue in their usual fashion - they've (removed)[https://bugzilla.gnome.org/show_bug.cgi?id=676842] the compact view from Nautilus completely. No feature - no bug. So we'll have to find our own way to fix the problem.

freed00m:
@monsta - wow they really did so :O 

Is there any way I can contribute more? Only thing I can do now is report and test bugs.

monsta:
That's good already :)
Of course we accept bugfixes via PRs as well.

PR #434, opened 2015-06-24
compact view: don't auto-scroll to the right on long filenames

monsta:
however, now it will auto-scroll to the left if you manually
scroll to the right and then click on any filename. I presume
this behavior is more desired than the previous one. if it's
not so, let me know (in the comments).

adapted from
https://github.com/linuxmint/nemo/commit/72c2214c82637ab622438d46c28e67289a49d472
(thanks to @glebihan)

fixes https://github.com/mate-desktop/caja/issues/406

@flexiondotorg @NiceandGently @stefano-k 
It works fine in GTK+2 build in Debian Testing, but I'd like to get it tested more before merging (e.g. GTK+3 build, other distros, etc.) :smile:


monsta:
Ok, it also works in GTK+3 build in Debian Testing.

PR #485, opened 2015-12-01
GTK3: fix desktop flashes

lukefromdc:
Two commits, both can be used or just the first.

The first commit stops the flashes when clicking between the desktop and open windows, in all themes. Remaining flashes turn black.  Text on desktop rename labels (unselected text)  had to be hardcoded to black (sane default), themes can override by using the #fmdesktopiconview>.view>* selector.  This commit overrides the .view and .view:backdrop selectors that differ in themes like Adwaita and cause desktop flashes when clicking between the desktop and open windows.
 
The second commit turns the remaining flashes transparent, so they are never seen except by momentary intensification of shadows when changing themes or during heavy disk activity. By forcing transparency for the desktop icon view window, the flashes become transparent and thus almost invisible instead of black. Testing with gtk3.19 I could not reproduce the reported crash on changing backgrounds with fade option enabled. 

The second commit did introduce one new compiz issue, compiz must be restarted to switch to a solid color background. Not necessary for gradient or picture backgrounds.

There are compiz issues with current git master compiz that did not come from these commits. These are a requirement to enable the fade option to swtich to a solid color background, and an ugly failure to redraw the desktop on moving a window when a soild color background is used with icons on the desktop disabled. I tested these issues with and without these commits to be sure.

### DIFF
--- a/src/caja-desktop-window.c
+++ b/src/caja-desktop-window.c
@@ -138,6 +138,11 @@ caja_desktop_window_new (CajaApplication *application,
 
     /* Special sawmill setting*/
     gtk_window_set_wmclass (GTK_WINDOW (window), "desktop_window", "Caja");
+    
+#if GTK_CHECK_VERSION (3, 0, 0)      /*turn remaining gtk3.16 and later flashes transparent */   
+    GdkVisual *visual = gdk_screen_get_rgba_visual(screen);
+    gtk_widget_set_visual(GTK_WIDGET(window), visual);
+#endif    
 
     g_signal_connect (window, "delete_event", G_CALLBACK (caja_desktop_window_delete_event), NULL);
 
--- a/src/file-manager/fm-desktop-icon-view.c
+++ b/src/file-manager/fm-desktop-icon-view.c
@@ -836,6 +836,23 @@ fm_desktop_icon_view_create (CajaWindowSlotInfo *slot)
     view = g_object_new (FM_TYPE_DESKTOP_ICON_VIEW,
                          "window-slot", slot,
                          NULL);
+                         
+     /*fix most of the nasty gtk3.16 and later flashes*/
+#if GTK_CHECK_VERSION (3, 0, 0) 
+    gtk_widget_set_name(view, "fmdesktopiconview");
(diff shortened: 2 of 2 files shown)


lukefromdc:
Closing this because gtk3.19.4 git master up to date 12-13-2015 breaks this. The new version passes theme information differently and any reference to :backdrop causes the flash. Somehow it does not in Adwaita, but this patch now works ONLY in Adwaita. The window transparency patch by itself now works fine in themes that do not set view:backdrop but does not work in Adwaita because that selector is present. More work needed if all post-3.14 GTK 3 versions to work

raveit65:
Well, major prob with this PR was that i always noticed a redraw of desktop icons (crash/and restart of caja) after transition end if destop background fade option is enabled, during wallpaper change.
Noticed with every gtk+-3.x version.
I use the fixes for my inofficial gtk3 repo for fedora which is experimental, here i can disable the fade option with a gsettings override.
But this isn't acceptable for official repos.
Sadly to hear that gtk+ (damage, inc. by trouble maker) busted that workaround.

lukefromdc:
I'm working on another version right now, but it looks like getting both Adwaita and 
all other themes to work will be very difficult. I can get one or the other to work 
now, but not both. If I can find out why this still works in Adwaita and only Adwaita
I have my fix for now. Still, we need to find a way to entirely stop the desktop background
from reloading at all until it is all the way ready to reload quickly

raveit65:
Forget Adwaita, Mate use mate-themes.
....I've still start on port BlueMenta for 3.20.

lukefromdc:
I am trying to make this work on a theme-independant basis so that no theme creates flashes. My own theme and Adwaita are the only gtk3.20 themes I have now, I tested on all the MATE themes when I was on 3.18.  I just got results with one new line added to one file, will have it up shortly but when I deleted the PR branch the commits on it got added to my master branch and I have to find a way to revert them first before making a new branch.  Otherwise I have no way to post any clean PR's and I don't want to delete the whole repo again now that it has been used

lukefromdc:
New PR https://github.com/mate-desktop/caja/pull/486

PR #486, opened 2015-12-15
Flashfix5

lukefromdc:
New version of previous PR for fixing desktop flashes, this version works both before and after the commit to gtk3.19.4 that broke the previous effort. The first commit here, as before, stops most flashes. A single additional line was needed to hardcode #fmdesktopiconview (the icon view) as well as #fmdesktopiconview .view (the icon container within) to stop flashes on themes other than Adwaita. This branch now works again,  though the previously reported issue with crashes on changing backgrounds with fade enabled will probably be unchanged. Also the memory leak on a flash is still present, this covers up the symptoms rather than fixing the underlying problem. This is a "good enough for now" fix but not the RIGHT fix.

### DIFF
--- a/src/caja-desktop-window.c
+++ b/src/caja-desktop-window.c
@@ -138,6 +138,11 @@ caja_desktop_window_new (CajaApplication *application,
 
     /* Special sawmill setting*/
     gtk_window_set_wmclass (GTK_WINDOW (window), "desktop_window", "Caja");
+    
+#if GTK_CHECK_VERSION (3, 0, 0)      /*turn remaining gtk3.16 and later flashes transparent */   
+    GdkVisual *visual = gdk_screen_get_rgba_visual(screen);
+    gtk_widget_set_visual(GTK_WIDGET(window), visual);
+#endif
 
     g_signal_connect (window, "delete_event", G_CALLBACK (caja_desktop_window_delete_event), NULL);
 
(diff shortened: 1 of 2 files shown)


raveit65:
I noticed same problem with changing wallpapers if 'background-fade' is enabled.
I appreciate your efforts to fix this annoying issue, but we can't break the background-fade function.
We need to find out th real cause of the flashes, imo it is trigger when icon-cache is updated.
Well, we should leave this report open for that everybody can download your good workaround.

Sorry

lukefromdc:
I suspect that finding the ultimate cause would require building Caja 1.10, which
flashed only with gtk3.16 or higher, then with it in place an indirect GIT bisect of 
every change to gtk3.15 until the offending commit is found, thus revealing the 
offending code. This would be quite beyond me, I am seeing dozens and dozens
of new commits to gtk3.19 every day anmd am sure gtk3.15 was the same.

My guess is that commits concerning GtkWindow would be at fault, as that version
of GTK also brough in all those compiz/transparancy issues. They just broke the 
position of the calendar today but I suspect that will be fixed within a couple days
as bugs are going in and out of gtk3.19.5 as radical changes come in and are
debugged. At least this way if something comes in and is NOT fixed I will know 
exactly when it came it and can look for it.

monsta:
Now that the flashes seem to be fixed by https://github.com/mate-desktop/caja/pull/498, what should we do about this PR?

lukefromdc:
Close it-it has been superseded by something better. I will do that myself as
we are done with this kludgy workaround.

lukefromdc:
https://github.com/mate-desktop/caja/pull/498 fixed the flashes right, and without the side effects of this workaround. Closing this as a superseded kludge that worked when it was needed but now we've got something far better.

Issue #438, opened 2015-07-13
Wrong text color in Caja, Mint MATE 17.1/2


ingeva:
From LinuxMint MATE 17.1, caja sometimes switches text color to very light grey, making it impossible to read. I'm using a standard Mint-X theme with slightly changed colors but it's not possible to change foreground/background color in window. I've tried to install other themes but haven't been able to make it work. Instructions don't match reality. If I change background or text color in theme, text reappears, only to disappear again shortly after.
Clicking on the right side after F3 makes text reappear, but it tends to occur again very soon. I've tried Nautilus and Nemo but they lack caja's nice F3 feature.
Screenshot follows.
![screenshot-phpbb](https://cloud.githubusercontent.com/assets/1525307/8644048/cd48e680-293b-11e5-9f95-a7db5e411973.png)

flexiondotorg:
Please report this issue to Linux Mint, the MATE team do no maintain the Mint-X theme.

ingeva:
I did. They told me to report it here. I believe it's a caja problem, not a theme problem, but I need confirmation. They could not reproduce the error, and I can't NOT reproduce the error.

flexiondotorg:
I've never encountered this issue on Arch Linux or on Ubuntu MATE. If you can provide steps to reproduce the issue using the Menta theme then we can reopen this issue and investigate ☺

ingeva:
As I mentioned, I believe it's a caja problem, not a theme problem, but I'm open to anything as long as I can make it work. It's too annoying as it is.
There are several directories in all themes. Which one do I place in ~/.themes?

ingeva:
For now I have switched to the Mint-X Blue theme, and just changed the window border, no colors. I haven't seen the error yet. If it doesn't happen again I'll just stick to this theme. As long as no one else reports the error, we'll assume it's a local problem only. For me the theme is unimportant. I just want a working system with good colors (contrast) and visible text. :)

infirit:
Please also experiment with the default mental theme.

ingeva:
The default theme is Mint-X and its variants, and I've tried to install other themes with no luck.

iugostudio:
I have similar problem. 
![lightgrey_309](https://cloud.githubusercontent.com/assets/14885632/11241690/8751967a-8e05-11e5-8a9a-e1021b270f70.png)

ingeva:
[...]

Issue #81, opened 2013-02-20
caja doesn't use the the right font color in iconview with dark themes


raveit65:
If i use use a dark desktop theme caja-1.5.x doesn't displayed the expected white font color for folder names if i used 'iconview' when caja starts.
I see the folder names with black color on dark background.
I can toogle to 'listview' with crtl+2 and the font color is white, than toogle back to 'iconview' with crtl+1 and the fontcolor has the expecected value 'white'.
All sreenshots with gnomishdark
- after caja starts with iconview
https://dl.dropbox.com/u/49862637/Mate-desktop/Bugs/caja-1.png
- switch to listview
https://dl.dropbox.com/u/49862637/Mate-desktop/Bugs/caja-2.png
- back to iconview
https://dl.dropbox.com/u/49862637/Mate-desktop/Bugs/caja-3.png

Also i  notice, if i switch back to a white theme and a caja window is open, the font color stays on white color with a white caja background in iconview.

This issue happes with all dark themes which i know, (gnomishdark, delorean-dark, anewstartblood, drackfire dreamblack) so i don't think it's a theme issue.
All tested themes working well with caja-1.4.x.
@jasmineaura
Does #72 fix this perhaps?

sunweaver:
This has also been reported as  debbug#770760.

axalix:
I still have this problem. Caja 1.8.2. Linux Mint 17

dcharlespyle:
Still happening in mate-desktop 1.8.2, in Fedora 21.  As a workaround I used the eyedropper to set the background to the same color as the dark theme (BlackMATE) I am using.

Ralnoc:
This is still seen in Fedora 22 running xfce4 with Caja 1.10.2. 

A easy workaround is to apply a color background that doesn't conflict with the font coloring through 'Backgrounds and Emblems -> Colors'. You drag a non-conflicting color to the file list and this will address it. 

However we should not have to implement a workaround. Caja should honor the correct theme colors provided by the WM.

luispabon:
Still an issue on Ubuntu 15.10 + Mate 1.8 (Caja 1.10.4).

gamax92:
Still an issue, font color is black and after changing theme and back, it goes to proper text color. Then after a while I notice it's back to the wrong color.

raveit65:
fixed for gtk3

raveit65:
@sunweaver 
I've cherry-picked that commit to 1.12 too.
[...]

PR #495, opened 2016-01-06
Fix for #438

bl0ckeduser:
This should fix the issue and keep the old bugfix still good
Screen shot: http://www.cs.mcgill.ca/~gcormi5/bugfixcaja2016.png

Thank you
Happy new year 2016

Gabriel

### DIFF
--- a/libcaja-private/caja-icon-container.c
+++ b/libcaja-private/caja-icon-container.c
@@ -9434,7 +9434,9 @@ setup_label_gcs (CajaIconContainer *container)
     }
     else
     {
-        if (container->details->use_drop_shadows || eel_background_is_dark (background))
+        if (container->details->use_drop_shadows ||
+	     (eel_background_is_dark (background) && eel_background_is_set(background))
+	   )
         {
             setup_gc_with_fg (container, LABEL_COLOR, 0xEFEFEF);
             setup_gc_with_fg (container,


raveit65:
Happy new year and thank you.
merged https://github.com/mate-desktop/caja/commit/61800829da41036da5719b0e39e6f015cbd75f9a

monsta:
Thanks! That was an annoying issue.

I guess we're not reverting https://github.com/mate-desktop/caja/pull/285 then?

raveit65:
> I guess we're not reverting #285 then?

 no, only if you want to intoduced the old bug again :smile: 
But it seems that there is a general failure with this logic because we have the issue with dark themes. https://github.com/mate-desktop/caja/issues/81
I tested local https://git.gnome.org/browse/nautilus/commit/?id=ef74d171892ed7189b4a52695ae8de575d2dac1a
which replace eel-background usage and is used by nemo too, here the issue with dark themes is gone.
But only applying this commit causes other problems.

monsta:
@bl0ckeduser: damn, I can still reproduce it with custom background using steps from https://github.com/mate-desktop/caja/issues/438#issuecomment-169078577. Only it's not light on light, it's dark on dark...

E.g. I have black font in my GTK+ theme, then I set a dark background using Emblems & Backgrounds dialog, then open three tabs using the mentioned steps to reproduce. Text on the third tab is black on dark background... :confused:

monsta:
Note: this is **without** changes from https://github.com/mate-desktop/caja/pull/501.

monsta:
Ok, now fixed with https://github.com/mate-desktop/caja/pull/526.

Issue #457, opened 2015-09-09
use middle button to close tabs


sc0w:
Feature request:

To use middle button to close tabs, like firefox.

thanks

raveit65:
merged https://github.com/mate-desktop/caja/commit/ed977d25e487400f547997f95802d139e4fce26b

PR #497, opened 2016-01-08
Middle-click tab closing (#457)

bl0ckeduser:


raveit65:
@bl0ckeduser 
Do you think it's poosible to add a gsettings key for disable/enable it?
I can imagine that some users don't like the feature.

bl0ckeduser:
will investigate

PR #499, opened 2016-01-12
Small additional fix for #457

bl0ckeduser:
For the existing feature code to work the current selected tab must be correctly update when a middle click is made to a tab label. otherwise, middle click will always close the currently open tab and not the clicked one.

I am investigating making a gsettings parameter for this too as requested, it should be trivial.

### DIFF
--- a/src/caja-notebook.c
+++ b/src/caja-notebook.c
@@ -218,7 +218,7 @@ button_press_cb (CajaNotebook *notebook,
     tab_clicked = find_tab_num_at_pos (notebook, event->x_root, event->y_root);
 
     if (event->type == GDK_BUTTON_PRESS &&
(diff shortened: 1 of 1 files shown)


bl0ckeduser:
i can make a gui option too to go with the gsettings once i figure it out ;-)

monsta:
I think gsettings key is enough. I actually think it's more than enough as middle mouse button did nothing before, so I can't imagine someone would change the key to make the button do completely nothing again :smile:

Well... maybe except for accidental middle button clicks...

raveit65:
merged https://github.com/mate-desktop/caja/commit/4c7f59163c415140d76330a1b3c57b1efe35ca14
Thank you

PR #501, opened 2016-01-15
fix fg-color of dark themes

raveit65:
fixes https://github.com/mate-desktop/caja/issues/81

### DIFF
--- a/libcaja-private/caja-icon-container.c
+++ b/libcaja-private/caja-icon-container.c
@@ -9567,7 +9567,7 @@ setup_label_gcs (CajaIconContainer *container)
                           "frame_text", &frame_text,
                           NULL);
 
-    if (frame_text /* || !eel_background_is_set(background) */)
+    if (frame_text || !caja_icon_container_get_is_desktop (container))
     {
         setup_gc_with_fg (container, LABEL_COLOR,
                           eel_gdk_color_to_rgb (&style->text[GTK_STATE_NORMAL]));
@@ -9577,8 +9577,7 @@ setup_label_gcs (CajaIconContainer *container)
     }
     else
     {
-        if (container->details->use_drop_shadows ||
-	     (eel_background_is_dark (background) && eel_background_is_set(background)))
+        if (container->details->use_drop_shadows || eel_background_is_dark (background))
         {
             setup_gc_with_fg (container, LABEL_COLOR, 0xEFEFEF);
             setup_gc_with_fg (container,


raveit65:
@monsta @bl0ckeduser 
please test, in my test this does not break previous fixes for
https://github.com/mate-desktop/caja/issues/438
https://github.com/mate-desktop/caja/issues/105

bl0ckeduser:
No regressions for me with the foregoing patch

monsta:
Thanks, I don't see https://github.com/mate-desktop/caja/issues/81 anymore here :smile:

But I see it reverts the change from https://github.com/mate-desktop/caja/pull/495 - hope it's ok.

raveit65:
As i said this fixes all 3 issues for me.

monsta:
Ok good, can be merged in both master and 1.12 then :smile:

raveit65:
merged https://github.com/mate-desktop/caja/commit/1ff346876afd37a957aa490be25bdd2fd90e4773

monsta:
Well there's a regression after all... When using custom background color (you can set it via Emblems & Background dialog), font color isn't changed. E.g. if I select dark bg color and font of my GTK+ theme is black, it will remain black (and thus unreadable).

Tested in GTK+2 build.

@bl0ckeduser: can anything be done about that?

bl0ckeduser:
It is a quite complex issue. It seems eel_background_is_set and eel_background_is_dark routines are not functioning properly and returning bogus values on repeated calls. It might be a bug in mate-desktop, or in both. It would need some pretty detailed debugging to track down

bl0ckeduser:
I am investigating some more. When the bug happens, for some reason update_label_color() fails to be called whereas in previous instances it is called succesfully. Before the call, eel_background_is_dark returns false which leads to incorrect theming while after the call it returns true.

monsta:
Hmm... but I can reproduce it every time. No matter which background color I choose, the font stays black.

raveit65:
@bl0ckeduser 
any chance that you find a fix before 1.14 release end of march?

bl0ckeduser:
I don't think, sorry

monsta:
@alexandervdm: maybe you know what could cause the problem? See my description above. And if we'll revert https://github.com/mate-desktop/caja/commit/1ff346876afd37a957aa490be25bdd2fd90e4773 (which fixes the issue), we'll get https://github.com/mate-desktop/caja/issues/81 back, which isn't nice as well.

monsta:
Ok, now fixed with https://github.com/mate-desktop/caja/pull/526.

PR #507, opened 2016-01-22
GTK3: fix deprecated GtkMisc

raveit65:
@monsta 
@lukefromdc 
Please test.
Here a review is needed.
Most changes correct only deprecated gtk_misc_set_alignment for labels from pevious fix.
But i fixes also deprecated gtk_misc_{set/get}_padding in a few files.
Hope that i didn't break anything:)

### DIFF
--- a/eel/eel-labeled-image.c
+++ b/eel/eel-labeled-image.c
@@ -1102,8 +1102,13 @@ labeled_image_update_alignments (EelLabeledImage *labeled_image)
 
         if (labeled_image->details->fill)
         {
+#if GTK_CHECK_VERSION (3, 16, 0)
+            x_alignment = gtk_label_get_xalign (GTK_LABEL (labeled_image->details->label));
+            y_alignment = gtk_label_get_yalign (GTK_LABEL (labeled_image->details->label));
+#else
             gtk_misc_get_alignment (GTK_MISC (labeled_image->details->label),
                                     &x_alignment, &y_alignment);
+#endif
 
             /* Only the label is shown */
             if (!labeled_image_show_image (labeled_image))
@@ -1139,9 +1144,9 @@ labeled_image_update_alignments (EelLabeledImage *labeled_image)
 
             }
 
-#if GTK_CHECK_VERSION (3, 14, 0)
-            gtk_widget_set_halign (labeled_image->details->label, GTK_ALIGN_CENTER);
-            gtk_widget_set_valign (labeled_image->details->label, GTK_ALIGN_CENTER);
+#if GTK_CHECK_VERSION (3, 16, 0)
+            gtk_label_set_xalign (GTK_LABEL (labeled_image->details->label), x_alignment);
+            gtk_label_set_yalign (GTK_LABEL (labeled_image->details->label), y_alignment);
 #else
             gtk_misc_set_alignment (GTK_MISC (labeled_image->details->label),
                                     x_alignment,
@@ -1157,8 +1162,13 @@ labeled_image_update_alignments (EelLabeledImage *labeled_image)
 
         if (labeled_image->details->fill)
         {
+#if GTK_CHECK_VERSION (3, 0, 0)
+            x_alignment = gtk_widget_get_halign (labeled_image->details->image);
+            y_alignment = gtk_widget_get_valign (labeled_image->details->image);
+#else
             gtk_misc_get_alignment (GTK_MISC (labeled_image->details->image),
                                     &x_alignment, &y_alignment);
+#endif
 
             /* Only the image is shown */
             if (!labeled_image_show_label (labeled_image))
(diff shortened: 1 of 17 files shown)


raveit65:
Note, for such a line
gtk_misc_set_alignment (GTK_MISC (label), 0.0, 0.5);
i set only the x value for gtk3.
It looks like that 0.0/ 0.5 is a default, i most examples gome upstream set only the x value in this case.
So i followed them it it works in my test.

lukefromdc:
I found that the icons in the file properties/emblems tab don't line up but that that is already the case in mate-desktop/master, no change there

emblems_in_properties

Other than this, no obvious problems in my test and nothing unique to this commit

monsta:
> Note, for such a line
gtk_misc_set_alignment (GTK_MISC (label), 0.0, 0.5);
i set only the x value for gtk3.
It looks like that 0.0/ 0.5 is a default, in most examples gome upstream set only the x value in this case.

According to [docs](https://developer.gnome.org/gtk3/stable/GtkMisc.html), both ```xalign``` and ```yalign``` properties have 0.5 as the default value...

src/caja-location-bar.c
```diff
@@ -263,8 +267,17 @@ style_set_handler (GtkWidget *widget, GtkStyle *previous_style)
     pango_layout_get_pixel_size (layout, &width2, NULL);
     width = MAX (width, width2);
 
+#if GTK_CHECK_VERSION (3, 0, 0)
+    gtk_widget_get_margin_start (widget);
+    gtk_widget_get_margin_end (widget);
+    margin_top = gtk_widget_get_margin_top (widget);
+    margin_bottom = gtk_widget_get_margin_bottom (widget);
+    xpad = margin_start + margin_end;
+    ypad = margin_top + margin_bottom;
```
monsta: Minor thing: ```ypad```, ```margin_top```, ```margin_bottom``` are only assigned here and never used elsewhere - so they can be safely removed.

Major thing: ```margin_start``` and ```margin_end``` are not initialized before use; did you mean something like this instead?

```C
margin_start = gtk_widget_get_margin_start (widget);
margin_end = gtk_widget_get_margin_end (widget);
```

eel/eel-labeled-image.c
```diff
@@ -1102,8 +1102,13 @@ labeled_image_update_alignments (EelLabeledImage *labeled_image)
 
         if (labeled_image->details->fill)
         {
+#if GTK_CHECK_VERSION (3, 16, 0)
+            gtk_label_get_xalign (GTK_LABEL (labeled_image->details->label));
```
monsta: Both ```gtk_label_get_xalign``` and ```gtk_label_get_yalign``` actually return values. Given that this code block should do the same thing as ```gtk_misc_get_alignment``` does with GTK+ < 3.16, I think it should look like this:

```C
x_alignment = gtk_label_get_xalign (GTK_LABEL (labeled_image->details->label));
y_alignment = gtk_label_get_yalign (GTK_LABEL (labeled_image->details->label));
```

eel/eel-labeled-image.c
```diff
@@ -1157,8 +1162,13 @@ labeled_image_update_alignments (EelLabeledImage *labeled_image)
 
         if (labeled_image->details->fill)
         {
+#if GTK_CHECK_VERSION (3, 0, 0)
+            gtk_widget_get_halign (labeled_image->details->image);
```
monsta: Same issue as with ```gtk_label_get_xalign``` / ```gtk_label_get_yalign``` above.

monsta:
@lukefromdc: not major thing, but I see broken vertical alignment of checkboxes and icons on your screenshot... normally, top of checkbox and top of icon should be at the same horizontal level.
I don't see this issue without this PR (GTK+ is 3.18.6).

monsta:
Oops, not at exactly the same level actually... top of icon is a bit higher than top of checkbox.

emblems

src/caja-location-bar.c
raveit65: This way it is fixed in marco too, taken from metacity

raveit65: just taken from marco fix for it.
Please correct all this
... i'm tired of of working on caja after 4 weeks

eel/eel-labeled-image.c
raveit65: Please fix it , i'm tired of working on caja after 4 weeks

eel/eel-labeled-image.c
raveit65: Please fix it , i'm tired of working on caja after 4 weeks

monsta:
Ok, fixed the mentioned issues, rebased and force-pushed. Please test it guys.

lukefromdc:
OK, built this and installed it, everything seems to work.  BTW, I looked at that
emblems tab in "properties" in GtkInspector, looks like the emblems are being
centered in a box whose size is determined by the label.

Sorry I could not test earlier but I am mostly active afternoon and night. Believe me
all that work on Caja is appreciated, I thought that desktop flash bug might never
be found as I never was able to find the root cause on my own. That fix was good 
enough I put it on a friend's machine running MATE with gtk3, since they've got
GTK3.16. No complaints from them about new bugs at all.

I don't know how many people run a whole locally build MATE environment from 
GIT master or ahead as their main DE but I do. I have backup DE's and a whole
backup partition as well if I get into trouble because either my code or somene else's
doesn't work.

PR #513, opened 2016-02-09
Gtk3.20 icon view theme support

lukefromdc:
Gtk3.19 /3.20: Add .view style class on icon view and compact view scrolled windows only

Applying the .view class to the icon container no longer works in gtk3.19, so apply it to the scrolled window instead.  Apply it in both icon view and compact view, as both need it.  Doing it from this file prevents it from being applied to the scrolled window under a treeview, thus preventing the exposed corner above the scrollbar in a treeview from showing the view background color.

Tested this build in gtk3.19.8 with good results. Tested in my own theme (UbuntuStudio_Legacy),Adwaita,  BlackMATE, Green-Submarine, and BlueMenta, all still work right. Adwaita now shows the proper white view color, and my own theme gives a white icon view without also showing
white squares over the treeview scrollbars. That was an unwanted side effect of applying the view color in the gtk theme to all the scrolled windows containing list, icon, or compact views as there is no way to tell them apart from the theme only.  This approach keeps the scrolled window under the treeview transparent but themes them under icon or compact views.

Using .view rather than a custom class enables themes not written for MATE to set the Caja background to match expected view colors, just as it did on the icon container in gtk3.18 and earlier. Leave it on the icon container too, as it may be needed in some themes to theme elements within the icon view.

Also, I found no ill effects from the added style class on gtk3.18. In that case, the .view style class on the icon container should cover over anything in the scrolled window. Since the style class is not applied to the scrolled window used by a list view, no ill effect there in gtk 3.19 or gtk3.18.  Applying to gtk3.19 only builds for now just in case


raveit65:
merged https://github.com/mate-desktop/caja/commit/e5c29655e5301b6d4ba19a3d4970745804039c0e
Thank you

PR #519, opened 2016-02-29
fix loading multiple python extensions (regression since 1.9.3)

alexandervdm:
My extension management code (GSoC2014) caused a regression that stops multiple Python extensions from loading correctly. This issue was reported as [issue 3](https://github.com/mate-desktop/python-caja/issues/22) on the python-caja tracker. This commit fixes the immediate problem.


monsta:
Works nice, thanks :smile:

Issue #520, opened 2016-03-01
Python extensions might cause duplicates of python-caja in extensions list


monsta:
@alexandervdm: there's an interesting side effect visible after your fix in https://github.com/mate-desktop/caja/commit/8a826a675862d56b6d01c8467282bde9171f8cf8...

When I add more than one Python extension into ```/usr/share/caja-python/extensions/``` or ```~/.local/share/caja-python/extensions/```, a duplicate of python-caja appears in the extensions list. In other words, every Python extension, except for the first one, adds a duplicate in the list. For example, here's how it looks with 3 Python extensions added:

![caja-python-duplicates](https://cloud.githubusercontent.com/assets/5138986/13422804/253cb5d8-dfa8-11e5-89f1-6163d481684e.png)

Enabling/disabling one of these identical items enables/disables all Python extensions (it can be seen when opening the dialog next time).

monsta:
Whoops. I forgot that in some of these VMs I had some extensions in ```~/.local/share/caja-python/extensions/``` as well. :smile:

So...
- 1 Python extension - no duplicates in the list
- 2 Python extensions - 1 duplicate
- 3 Python extensions - 2 duplicates

and so on...

alexandervdm:
The duplicates appear because the extension management system registers the libcaja-python.so module instead of the child python extension that is loaded through that module. 

I've prepared [a commit](https://github.com/alexandervdm/caja/commit/897eec276cc2065f5ebdbb032cf5bb45a4d6f7d2) that adds a check preventing libcaja-python.so from being registered multiple times. This fixes the duplicates problem, but still does not allow python extensions from being managed individually. 

For this to happen we need to expose information like the filename of the actual python file to caja. This will require changes to both caja and python-caja. I'm working on it and aim to have this patch ready by tomorrow night.

alexandervdm:
The two PR's referenced above fix the issue. Note that Python extensions do not expose detailed information (author, year, description and such) yet, but I'll follow up with that improvement when I can.

monsta:
Fixed with https://github.com/mate-desktop/caja/commit/2ca207ea1a91178d28583e2bdab5c3d6aac06d87, https://github.com/mate-desktop/caja/commit/c0bfecbb6f70bc732d35b477c57d538730c358bd and https://github.com/mate-desktop/python-caja/commit/6925f41da64b5aebee5d452a40c541303dd252bf.

Issue #22, opened 2012-06-05
Caja Critical errors when starting up.


rsantellan:
Using caja on linux mint 13 mate, doing caja --quit show this on the .xsession-errors file

Shutting down caja-open-terminal extension

--- Hash table keys for warning below:
--> application/xml
--> inode/directory
--> application/x-desktop
--> l2049
--> rodrigo
--> Rodrigo Santellan

(caja:2768): Eel-WARNING **: "unique eel_ref_str" hash table still has 6 elements at quit time (keys above)

(caja:2768): Eel-WARNING *_: "caja-directory.c: directories" hash table still has 4 elements at quit time
Initializing caja-open-terminal extension
*_ Message: Initializing gksu extension...

*\* (caja:2816): CRITICAL **: mate_bg_crossfade_start: assertion `fade->priv->fading_pixmap != NULL' failed

Firefox.PlacesItemSource "Firefox Places" encountered an error in UpdateItems: System.InvalidCastException: Cannot cast from source type to destination type.
  at Mono.Data.Sqlite.SqliteDataReader.VerifyType (Int32 i, DbType typ) [0x00000] in <filename unknown>:0 
  at Mono.Data.Sqlite.SqliteDataReader.GetString (Int32 i) [0x00000] in <filename unknown>:0 
  at Firefox.PlacesItemSource+<LoadPlaceItems>c__Iterator3.MoveNext () [0x00000] in <filename unknown>:0 
  at System.Collections.Generic.List`1[Firefox.PlaceItem].AddEnumerable (IEnumerable`1 enumerable) [0x00000] in <filename unknown>:0 
  at System.Collections.Generic.List`1[Firefox.PlaceItem]..ctor (IEnumerable`1 collection) [0x00000] in <filename unknown>:0 
  at System.Linq.Enumerable.ToArray[PlaceItem](IEnumerable`1 source) [0x00000] in <filename unknown>:0 
  at Firefox.PlacesItemSource.UpdateItems () [0x00000] in <filename unknown>:0 
  at Do.Universe.Safe.SafeItemSource.UpdateItems () [0x00000] in <filename unknown>:0 .

PR #521, opened 2016-03-07
Fix for duplicate Python extensions in management interface

alexandervdm:
This commit fixes issue #520. Note that python-caja [pull-request 24](https://github.com/mate-desktop/python-caja/pull/24) is also required for this solution.

### DIFF
--- a/libcaja-extension/caja-extension-types.h
+++ b/libcaja-extension/caja-extension-types.h
@@ -64,6 +64,7 @@ extern "C" {
     void caja_module_shutdown   (void);
     void caja_module_list_types (const GType **types,
                                  int          *num_types);
+    void caja_module_list_pyfiles (GList     **pyfiles);
 
 #ifdef __cplusplus
 }
--- a/libcaja-private/caja-extensions.c
+++ b/libcaja-private/caja-extensions.c
@@ -34,7 +34,7 @@ static GList *caja_extensions = NULL;
 
 
 static Extension *
-extension_new (gchar *filename, gboolean state, GObject *module)
+extension_new (gchar *filename, gboolean state, gboolean python, GObject *module)
 {
     Extension *ext;
     GKeyFile *extension_file;
@@ -50,7 +50,7 @@ extension_new (gchar *filename, gboolean state, GObject *module)
     ext->website = NULL;
     ext->state = state;
     ext->module = module;
-
+    
     extension_file = g_key_file_new ();
     extension_filename = g_strdup_printf(CAJA_DATADIR "/extensions/%s.caja-extension", filename);
     if (g_key_file_load_from_file (extension_file, extension_filename, G_KEY_FILE_NONE, NULL))
@@ -65,6 +65,12 @@ extension_new (gchar *filename, gboolean state, GObject *module)
     }
     g_key_file_free (extension_file);
     g_free (extension_filename);
+    
+    if (python)
+    {
+        ext->name = g_strconcat("Python: ", filename, NULL);
+        ext->description = "Python-caja extension";
+    }
 
     return ext;
 }
@@ -193,11 +199,16 @@ void
 caja_extension_register (gchar *filename, GObject *module)
 {
     gboolean ext_state = TRUE; // new extensions are enabled by default.
+    gboolean ext_python = FALSE;
     gchar *ext_filename;
     GList *l;
     
     ext_filename = g_strndup (filename, strlen(filename) - 3);
     ext_state = caja_extension_get_state (ext_filename);
+    
+    if (g_str_has_suffix (filename, ".py")) {
+        ext_python = TRUE;
+    }
 
     /* Do not attempt to register already registered extensions */
     for (l = caja_extensions; l != NULL; l = l->next)
@@ -208,7 +219,7 @@ caja_extension_register (gchar *filename, GObject *module)
         }
     }
 
-    Extension *ext = extension_new (ext_filename, ext_state, module);
+    Extension *ext = extension_new (ext_filename, ext_state, ext_python, module);
     caja_extensions = g_list_append (caja_extensions, ext);
 }
 
--- a/libcaja-private/caja-extensions.c
+++ b/libcaja-private/caja-extensions.c
@@ -192,13 +192,23 @@ caja_extensions_get_list (void)
 void
 caja_extension_register (gchar *filename, GObject *module)
 {
-    gboolean state = TRUE; // new extensions are enabled by default.
-    gchar *extname;
+    gboolean ext_state = TRUE; // new extensions are enabled by default.
+    gchar *ext_filename;
+    GList *l;
     
-    extname = g_strndup (filename, strlen(filename) - 3);
-    state = caja_extension_get_state (extname);
+    ext_filename = g_strndup (filename, strlen(filename) - 3);
+    ext_state = caja_extension_get_state (ext_filename);
 
-    Extension *ext = extension_new (extname, state, module);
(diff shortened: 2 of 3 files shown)


monsta:
Thanks - but it doesn't build here :disappointed: 

```
  CC       caja-module.lo
caja-module.c: In function ‘add_module_objects’:
caja-module.c:173:31: error: ‘pyfiles’ undeclared (first use in this function)
         module->list_pyfiles(&pyfiles);
                               ^
caja-module.c:173:31: note: each undeclared identifier is reported only once for each function it appears in
```

Also I thought that it would be nice to have some protection against loading older version of ```libpython-caja.so```. When ```list_pyfiles``` function isn't found in the module, there should be no attempt to call it...

alexandervdm:
Fixed the missing variable initialisation. Sorry about that. I manually merged my 1.10 based work against master.. so I was asking for it. I agree with your other comment as well, will have to look into it once I get off work. Thanks for the feedback!

monsta:
Builds fine now, thanks. I've tested it with git master version of ```python-caja```, and as I suspected, Caja crashes right on startup (when python extension is enabled), then session manager restarts it, then it crashes again, etc. So the protection is indeed required here. I'll wait for your fix :smile:

Regarding the first commit: as I understand it, it doesn't require changes in python-caja?
I'm asking because I'd like to put https://github.com/mate-desktop/caja/commit/8a826a675862d56b6d01c8467282bde9171f8cf8 and the first commit from here into 1.12 branch and make a new 1.12.x release. The fixes are small and they should be enough for 1.12.x users.
Other commits from here are bigger and they also add a new feature, so they should go into 1.14 only. :smile:

alexandervdm:
Thanks for testing - I'll get the protection code to you as soon as I can. :)

You are correct in that 897eec276cc2065f5ebdbb032cf5bb45a4d6f7d2 and 8a826a675862d56b6d01c8467282bde9171f8cf8 together fix this issue and the one described in [#22](https://github.com/mate-desktop/python-caja/issues/22). The only downside of including only these two is that python extensions cannot be managed individually, but a bugfix release that includes these commits would still improve the python extension situation a lot.

monsta:
BTW, if you don't mind, I can merge https://github.com/mate-desktop/caja/commit/897eec276cc2065f5ebdbb032cf5bb45a4d6f7d2 already, and then you can rebase your branch (and probably rename this PR :smile:).

flexiondotorg:
@alexandervdm Thanks for working on this ☺️ Do you have an ETA for the protection code? I need a fix for this quite urgently and my offer to reward your efforts still stands.

alexandervdm:
d0bd3f4a2a970049e4f5cf8a50a95ea3162031a1 adds protection code around the module loading. Tested OK on original and the new python-caja. If there's any issues, please let me know.

alexandervdm:
451cc95f2abafd5342d92496ba2914388fdaf1ec is an optional addition that sets a default name/description for python modules. [Example.](http://i.imgur.com/ibipu66.png)

monsta:
Works good, thanks! :smile:

monsta:
> Python extension don't include the .caja-extension files that expose details (name, description, author, etc) for C extensions shown in the interface above. I'll look into a Python variation for this mechanism. It will require a general change to caja and changes to individual python extensions to support this. Is this a patch you'd accept still for 16.04, or is this a next release kind of patch either way?

My opinion is that this new feature can go into next release.
I'm not sure if it should be done via .caja-extension files: we might use some special names in the comments at the start of .py files, for example.

monsta:
@raveit65 @XRevan86 @lukefromdc
Please test it if you can guys. For the test you'll need to have python-caja installed and have some Python extensions (more than one!) in ```/usr/share/caja-python/extensions/``` or ```~/.local/share/caja-python/extensions/```. You can get a few extensions from ```examples``` dir of python-caja source tree.

What should happen:
- no duplicate "Python" entries in the extensions list (https://github.com/mate-desktop/caja/issues/520 should be fixed)

What should not happen:
- any crashes or warnings :smile:

What should happen if you also apply https://github.com/mate-desktop/python-caja/pull/24 to python-caja:
- every Python extension is listed as a separate entry in the extensions list

monsta:
@alexandervdm: I see a bit of inconsistency here though... when some Python extensions are installed, there's no separate entry for python-caja itself in the list, so you basically can't disable it via UI.

What do you think about it guys?

lukefromdc:
I've never had any python extensions installed for Caja and have had plenty of
issues with python in the past with BS like checkinstall wanting to put 
everything in the wrong place.

Maybe tomorrow I can play with this, but I've yet to even build python-caja and
will have to get it working first, probably with the example extensions mentioned,
Have a major video editing projectnto finish first.

monsta:
@alexandervdm: dammit... https://github.com/mate-desktop/caja/commit/897eec276cc2065f5ebdbb032cf5bb45a4d6f7d2, applied without later commits, actually prevents https://github.com/mate-desktop/caja/commit/8a826a675862d56b6d01c8467282bde9171f8cf8 from working :disappointed: 
Check 1.12 branch. Looks like I've made the release too soon...

lukefromdc:
My test was not sucessful, in a build of python-caja and both these
changes everything built and installed, but not one of the example
extensions showed up at all after copying them over to 
/usr/share/caja-python/extensions/ . Going into the preferences 
dialog/extensions after restarting Caja showed only the extensions
that were already there, not the new Python ones.

Could this be another GTK 3 issue? I got a configuration warning for
python-caja about --with-gtk being not a rcognized option, rather 
like with Mozo. Mozo never worked with GTK3 until the code was
changed to support it, now it is GTK3 only and works just fine.

monsta:
Ok, looks like I've missed an important thing during my earlier testing... indeed, this PR alone (without https://github.com/mate-desktop/python-caja/pull/24) eliminates the duplicates from the extensions list, but multiple Python extensions don't work either.

I think https://github.com/mate-desktop/caja/commit/897eec276cc2065f5ebdbb032cf5bb45a4d6f7d2 prevents them from working once again. :confused:

monsta:
@lukefromdc: well, ```python-caja``` works fine for me in GTK+3 build. I actually see one Python extension loaded and showing a right-click menu item (it's ```folder-color``` extension).

monsta:
@alexandervdm: so, here are more test results:

- this PR alone:

  - as is: no duplicates, multiple extensions don't work;

  - with 897eec276cc2065f5ebdbb032cf5bb45a4d6f7d2 reverted: duplicates are present, multiple extensions work;

- this PR + https://github.com/mate-desktop/python-caja/pull/24:

  - as is: no duplicates, multiple extensions work. Also every Python extension is listed, but python-caja itself isn't (see https://github.com/mate-desktop/caja/pull/521#issuecomment-196002618);

  - with 897eec276cc2065f5ebdbb032cf5bb45a4d6f7d2 reverted: same. O_o

With that said, I think 897eec276cc2065f5ebdbb032cf5bb45a4d6f7d2 isn't needed and can be removed.

monsta:
Merged, except for 897eec276cc2065f5ebdbb032cf5bb45a4d6f7d2:
- 5639999463b0df7966e84e1c848289f51c192463 + 5fd73ddb561bab119623df39e20b801c5e3d9a1a + d0bd3f4a2a970049e4f5cf8a50a95ea3162031a1 as 2ca207ea1a91178d28583e2bdab5c3d6aac06d87
- 451cc95f2abafd5342d92496ba2914388fdaf1ec as https://github.com/mate-desktop/caja/commit/c0bfecbb6f70bc732d35b477c57d538730c358bd

PR #524, opened 2016-03-09
GTK 3.20: Fix desktop window rendering in all WM's

lukefromdc:
https://git.gnome.org/browse/gtk+/commit/?id=cdc580463e409a9b7e5f1480afa68d875ff3ff65 does not get along with
gtk_window_set_resizable false in a fullscreened, undecorated window. Put it behind a gtk3.18 and earlier only switch (don't risk breaking older GTK versions for now), and the desktop renders normally again. Tested and Works in Compiz, Marco, Mutter, and IceWM, didn't test any other WM's.


lukefromdc:
Closing this in favor of a https://github.com/mate-desktop/caja/pull/525 , which fixes this the same way Nautilus devs are doing by overriding the default window size which is the actual cause of the problem.

PR #526, opened 2016-03-17
fix font color issues in GTK+2 build

monsta:
@raveit65 @bl0ckeduser @alexandervdm @XRevan86

Hopefully this should fix all the font color issues for good. Please test it if you can. GTK+3 build shouldn't be affected by these changes due to ```#if```'s, so it should be enough to test GTK+2 build.

Please check whether **icon view** always has the right font color, and no regressions occur:
- desktop:
  - desktop font color should be white with black shadow, no matter which background color, gradient, or a wallpaper is chosen (https://github.com/mate-desktop/caja/issues/105 shouldn't happen);

- Caja window:
  - when no custom background is set and you switch between light and dark themes, font color should be black with light themes and white with dark themes (https://github.com/mate-desktop/caja/issues/81 shouldn't happen);

  - when custom background (either solid color or some pattern) is set, font color should be black with light background and white with dark background (the problem described in https://github.com/mate-desktop/caja/pull/501#issuecomment-172861877 shouldn't happen);

  - random changes of font color (https://github.com/mate-desktop/caja/issues/438) should not happen, no matter which theme or background is set. There are various steps to reproduce described in that report, for me the most reliable one (reproducible 100%) is described at https://github.com/mate-desktop/caja/issues/438#issuecomment-169078577.

### DIFF
--- a/eel/eel-background.c
+++ b/eel/eel-background.c
@@ -355,8 +355,6 @@ eel_background_ensure_realized (EelBackground *self)
 
     free_background_surface (self);
 
-    /* Calls mate_bg_set_color, which sets "ignore-pending-change" to false,
-       and queues emission of changed signal if it's still false */
     set_image_properties (self);
 
     window = gtk_widget_get_window (self->details->widget);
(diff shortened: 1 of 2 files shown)


raveit65:
Works well here with all test cases.
I rebased the commits for 1.12 which can be tested in
royal-pain-in-the-ass-1.12 branch.
Looks good here too.

bl0ckeduser:
It works for me, thank you

monsta:
@bl0ckeduser: just in case - did you check all the test cases from the list? I just don't want to introduce another regression... :smile:

bl0ckeduser:
yes

monsta:
Ok thanks :smile:

raveit65:
@monsta 
looks like this needs to be done for gtk3 too after new background change PRs

PR #533, opened 2016-04-15
GTK+-3 theming: add a CSS file for caja, and initialize it at startup

raveit65:
inspired by:
https://git.gnome.org/browse/nautilus/commit/?id=dad5d12

### DIFF
--- a/data/Makefile.am
+++ b/data/Makefile.am
@@ -33,6 +33,7 @@ cajadata_DATA = \
 	browser.xml \
 	caja-extras.placeholder  \
 	caja-suggested.placeholder \
+	caja.css \
 	$(NULL)
 
 # app data file
@@ -43,6 +44,7 @@ appdata_DATA = $(appdata_in_files:.xml.in=.xml)
 EXTRA_DIST = $(cajadata_DATA) \
 	caja.desktop \
 	caja.desktop.in \
+	caja.css \
 	freedesktop-dbus-interfaces.xml \
 	$(xml_in_files) \
 	$(desktop_in_files) \
--- a/data/caja.css
+++ b/data/caja.css
@@ -1,0 +1,23 @@
+
+
+.caja-canvas-item {
+    border-radius: 3px;
+}
+
+/* desktop mode */
+.caja-desktop.caja-canvas-item {
+    color: #ffffff;
+    text-shadow: 1px 1px alpha (#000000, 0.8);
+}
+
+.caja-desktop.caja-canvas-item:selected,
+.caja-desktop.caja-canvas-item:active,
+.caja-desktop.caja-canvas-item:hover {
+    text-shadow: none;
+}
+
+.caja-desktop EelEditableLabel.entry {
+   border-image: none;
+   border-width: 1px;
+   text-shadow: none;
+}
--- a/data/caja.css
+++ b/data/caja.css
@@ -16,8 +16,38 @@
     text-shadow: none;
 }
 
-.caja-desktop EelEditableLabel.entry {
+/* EelEditableLabel (icon labels) */
+.caja-desktop.view .entry,
+.caja-desktop.view .entry:focus,
+.caja-desktop.view .entry:backdrop,
+.caja-navigation-window .view .entry,
+.caja-navigation-window .view .entry:active,
+.caja-navigation-window .view .entry:focus,
+.caja-navigation-window .view .entry:backdrop {
    border-image: none;
+   border-style: solid;
    border-width: 1px;
+   border-color: #000000;
+   box-shadow: none;
+   border-radius: 3px;
    text-shadow: none;
+   background-image: -gtk-gradient(linear,
+                                   left top, left bottom,
+                                   from       (shade(rgba(255,255,255,1), 0.86)),
+                                   color-stop (0.15, shade(rgba(255,255,255,1), 0.96)),
+                                   color-stop (0.50, shade(rgba(255,255,255,1), 0.98)),
+                                   to         (shade(rgba(255,255,255,1), 1.00)));
 }
+
+.caja-desktop.view .entry:selected,
+.caja-desktop.view .entry:focus:selected,
+.caja-desktop.view .entry:backdrop:selected
+.caja-navigation-window .view .entry:active,
+.caja-navigation-window .view .entry:selected,
+.caja-navigation-window .view .entry:focus:selected,
+.caja-navigation-window .view .entry:backdrop:selected {
+    background-color: @theme_selected_bg_color;
+    color: @theme_selected_fg_color;
+    text-shadow: none;
+}
+
--- a/src/caja-application.c
+++ b/src/caja-application.c
@@ -1015,6 +1015,30 @@ desktop_changed_callback_connect (CajaApplication *application)
     return FALSE;
 }
 
+#if GTK_CHECK_VERSION (3, 0, 0)
+static void
+init_css (void)
+{
+    GtkCssProvider *provider;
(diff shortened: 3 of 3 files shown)


raveit65:
We need to be sure that canvas item use the right fg color with themes which don't use a styling for caja, ie. adwaita theme.
Before the PR desktop icons use a black fg color with adwaita theme, with the PR you see a white fg color with a black text shadow, which is readble for all kind of backgrounds.
We can also add more style settings if we think more stuff needs to be set.

Note: this works for all gtk+3 versions as we only use style classes, btw. EelEditableLabel is the css name with gtk+-3.20.
Gtk2 build is not affected.
@monsta 
@lukefromdc 
Please test.

raveit65:
Hope i added init_css (); at the right place, which is differ from orign commit.
Adding this line to https://github.com/mate-desktop/caja/blob/dev-css/src/caja-application.c#L334
gave me a build error.

lukefromdc:
I just tested this with Adwaita over GTK 3.21 (git master from a few days ago) and
it finally brought the correct text color to the desktop. This solves the "if it
does not work in Adwaita is doesn't really work in GTK3" issue. Had to restart
once with Adwaita preselected to get it working, but not sure if I had restarted
Caja after installing the new version. After that I could start it up on my normal
theme and Adwaita worked fine,

raveit65:
Of course caja needs to be restarted after installing the new code to initialize the css.
Simply adding only the css file does not work :)

lukefromdc:
OK, I just confirmed that you can switch to Adwaita after that first restart
to load the new version of Caja. When I thought I had first restarted it last
night I must have done something else. Anyway, it works.

monsta:
Looks ok in 3.18. BTW, 3.20.3 was pushed to Debian Unstable repos a couple of days ago, so if no critical bugs will be uncovered, it will reach Debian Testing repos in about 5 days, and I'll be able to test it. :smile:

raveit65:
merged.
https://github.com/mate-desktop/caja/commit/e8be955849664584f1fb697fd904ae9f85945a5e
https://github.com/mate-desktop/caja/commit/355a7dc4209ab6b0ddfdd5597c99d9a2ae63c406

lukefromdc:
An update to this created issues:
https://github.com/mate-desktop/caja/commit/355a7dc4209ab6b0ddfdd5597c99d9a2ae63c406
was needed for renaming files in Adwaita but cannot be overridden by any theme. Thus themes with certain basic color schemes like BlackMATE lose the unselected text, and my own theme loses the shadows on eel-editable-label in navigation windows. A gtk.css file in ~/.config/gtk-3.0 is the only way to override this. Reverting the commit isn't a solution as that breaks Adwaita. In BlackMATE the selected text is fine, it's the unselected text that disappears.

Doing something like this for mate-panel (menu problems in Adwaita) would break my theme rather badly and possibly others too unless some way is found to let the theme override the CSS file. I tried using "THEME" and "FALLBACK" priorities, but then the css file was ignored and Adwaita was broken again.

raveit65:
Well, with a transparent bg  the text entry at the desktop isn' t readable , depends on what wallpaper is used. For this reason i added a background image (1) for EelEditableLabel.
But i forgot to hard code the text color (2) with #000000 . This should bring back the unselected text for any theme. OK?
The 'box-shadow: none' i can remove, np. I thought it looks better without.
I stumbled about this during work on GreenLaguna, currently i didn't set anything in mate-applications.css.

The only problem that i see is not to hard code this values for text selection.
    background-color: @theme_selected_bg_color;
    color: @theme_selected_fg_color;

If a theme don't use this basic color names the selected text may be wrong.
But hard coding those values is a problem for all other themes.
I think (1) and (2) is necessary but i don't think we can match all possible situations in themes.
Any suggestions?

I saw that gnome-panel already use css file, btw. they use 2, one for Adwaita and one for HighContrast. But I'm not sure to do the same because we have more than 2 themes.
So, only settings without colors are save to use for all themes.

In general a theme artist should try to style mate-panel or caja or any other application.
We have added selectors for this. If a theme artist don't do that than his theme isn't suitable for mate desktop.
The idea from gtk+ devs to do not style any app is only theoretical. In practical this does not work., ie. mate-panel, firefox, and.... and....and.........., and especially with gtk+-3.20.
Every application is special and needs improvements, if you do more than the very simple adwaita styling.

raveit65:
OK, using this settings solves the issue with dark themes for me.
```
/* EelEditableLabel (icon labels) */
.caja-desktop.view .entry,
.caja-desktop.view .entry:focus,
.caja-desktop.view .entry:backdrop,
.caja-navigation-window .view .entry,
.caja-navigation-window .view .entry:active,
.caja-navigation-window .view .entry:focus,
.caja-navigation-window .view .entry:backdrop {
    border-image: none;
    border-style: solid;
    border-width: 1px;
    border-color: #000000;
    border-radius: 3px;
    color: #000000;
    text-shadow: none;
    background-image: -gtk-gradient(linear,
                                    left top, left bottom,
                                    from       (shade(rgba(255,255,255,1), 0.86)),
                                    color-stop (0.15, shade(rgba(255,255,255,1), 0.96)),
                                    color-stop (0.50, shade(rgba(255,255,255,1), 0.98)),
                                    to         (shade(rgba(255,255,255,1), 1.00)));
}
```
Edit /usr/share/caja/caja.css and do a 'killall caja'.
Any suggestions before i commit this

lukefromdc:
I just tested this, it works

lukefromdc:
Thanks for not killing the box shadow used in my theme for eel-editable-label,
this works quite nicely now

lukefromdc:
Here are some thoughts on the mate-panel/adwaita situation, which is
much more complex to resolve but nearly as ugly as the Caja situation
was.

If we start overriding themes for mate-panel, it will force me to move my
very complex panel menu theming to ~/.config/gtk-3.0/gtk.css or possibly
to move panel custom theming out of the GTK theme entirely, as is done
in Cinnamon. This could be done by having the panel look for  a theme file
somewhere that if it exists is loaded. In some tests last fall I made that work
but got segfaults if the file did not exist due to lack of testing for it first. 
In the "does not exist" case the default css could then load to do exactly
the same job.

This would fix the ugly menu problems in Adwaita but force the panel theming 
for many of the MATE themes into the new custom theming mode, possibly 
using separate files in the theme exactly the way Cinnamon themes are now
supported.  The good thing is that this sort of explicit loading would fix the 
problem of the default css overriding the theme css, since only one would 
be loaded.

raveit65:
> Thanks for not killing the box shadow used in my theme for eel-editable-label, this works quite nicely now

Np, i don't want break a theme from a team member ;)
I can also remove border-image: none;
But during porting of Menta and submarine themes i saw that an border-image goes inside of this entry for some reasons, which looks really bad.
Sadly, using padding and margin is ignored here.

> If we start overriding themes for mate-panel, it will force me to move my very complex panel menu theming to ~/.config/gtk-3.0/gtk.css or possibly to move panel custom theming out of the GTK theme entirely, as is done in Cinnamon.

I don't want to do that, because i do not really care about adwaita theme.
We are MATE...........and we we are the DE with the most working themes.
If someone want to use another theme he has to live with possible issues or copy our theme settings for the panel in his theme, which is easy if the theme don't use a gresource file ;)
Btw, i don't see a menu problem here with adwaita and gtk+- 3.20.3.
It looks OK but boring here.

raveit65:
fixed with https://github.com/mate-desktop/caja/commit/0b88f6fa5d043687c82a326dd5dcae4aed582644

PR #539, opened 2016-05-08
GTK3.21:fix sidebar segfault from (deprec)GtkStyle

lukefromdc:
Stop Caja from segfaulting with GTK 3.21 after commit 0f116135f4a5033ce4e9dfa19f10624701fa615c 
Avoid emitting ::style-set by name

To stop these segfaults,  put all instances of GtkStyle in caja-sidebar-title-c behind selectors keeping them out of GTK 3.21 or later builds. It worked and caused no visible problems in any theme, button at top of sidebars and sidebars themselves (all 5 options) look unchanged. Might need a more sophisticated fix  but this works for now.  For some reason only this file was affected.

No change to GTK 3.20 or earlier builds to avoid regressions and having to test every one of them


lukefromdc:
Closing in favor of https://github.com/mate-desktop/caja/pull/554, which ports the offending functions over to GtkStyleContext for all GTK3 versions

PR #558, opened 2016-05-27
GTK3:  Fix two deprecations, stop 3.21 segfault

lukefromdc:
In GTK 3.21, the use of GtkStyle in function style_set in caja-sidebar-title.c results in a segfault if the sidebar is showing, even though it is used only when the "information" sidebar is selected. GtkStyle also appears in update_title_font  in the same file.

The second function resizes the bold headline label font in the information sidebar when either the length of the text or the width of the sidebar changes. Port it to GtkStyleContext and keep it. 

The first function (style_set) does two things: It invokes the second function when the style is set up, so port its second input variable (which seems to receive only NULL anyway) to GtkStyleContext. The rest of the function is supposed to set the font for the "more information" label, but mostly seems to block updating that font with the system font and cause the size of the font to be different between GTK 3.20 or earlier and GTK 3.21. Disable that portion entirely in GTK3 builds, as that way the font is consistant across GTK3 versions(an appropriate size in all cases tested), updates with changes in the system font, and cannot segfault in GTK 3.21. Porting it to GtkStyleContext has been tested and stops the segfaults but leaves the other two problems mentioned. Disable that portion of style_set in GTK3 and be done with it.

### DIFF
--- a/src/caja-sidebar-title.c
+++ b/src/caja-sidebar-title.c
@@ -74,8 +74,13 @@ static GtkWidget *         sidebar_title_create_more_info_label (void);
 static void		   update_all 				(CajaSidebarTitle      *sidebar_title);
 static void		   update_more_info			(CajaSidebarTitle      *sidebar_title);
 static void		   update_title_font			(CajaSidebarTitle      *sidebar_title);
+#if GTK_CHECK_VERSION (3, 0, 0)
+static void                style_set                            (GtkWidget             *widget,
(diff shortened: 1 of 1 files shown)


lukefromdc:
Closing in favor of 
https://github.com/mate-desktop/caja/pull/559 based on this branch but also fixing all the rest of the GtkStyle instances I could find in Caja

PR #559, opened 2016-05-28
GTK3: port GtkStyle->GtkStyleContext

lukefromdc:
Port over all the remaining instances of GtkStyle I could find in Caja, keep caja-sidebar-title.c from fixes6 branch. Details below:

caja-sidebar-title.c Same file from https://github.com/lukefromdc/caja/tree/gtk3.21fixes6 GTK3: Ports update_title_font to GtkStyleContext and limit style_set function to calling update_title_font in GTK3 builds as latter function interferes with resizing fonts and does little else in GTK2.

caja-places-sidebar.c: Port caja_places_sidebar_style_set to GtkStyleContext in GTK3 builds

caja-pathbar.c: GtkStyleContext is already used to style the pathbar, so removed GtkStyle variables and unneeded caja_path_bar_style_set function

Note that this disables the most common call to caja_path_bar_check_icon_theme but icons update fine without it in GTK3. That being so, remove it and the call to it in caja_path_bar_screen_changed as well, so two functions not needed in GTK3 are removed. Theme and icon updating no change observed in testing

caja-information-panel.c: Port caja_information_panel_style_set to GtkStyleContext in GTK3 builds

caja-history-sidebar.c: port caja_history_sidebar_style_set GtkStyleContext in GTK3 builds

caja-location-bar.c: Port one variable in style_set_handler to GtkStyleContext in GTK3 builds

caja-zoom-control.c: label_style_set_callback to GtkStyleContext in GTK3 builds

Note that *previous_style is still used, unknown if this is having any effect or validity in GtkStyleContext. Not sure how to remove it but in this case it may simply be a variable name and thus valid anyway.

There is still an issue with icons in the places sidebar except information that only update after a theme change on restarting Caja, this is not new, predates any of these changes,can be observed with caja from Master over GTK 3.20, and is not affected by these commits. No other theme issues in tests with Gtk 3.16, 3.18, 3.20, or 3.21

PR #560, opened 2016-05-30
GTK3: remove 4 deprecated functions

lukefromdc:
Fix the build warnings from https://github.com/mate-desktop/caja/pull/559  by completely removing these four "style set" functions. 

No effect on theming, changing of themes, or font rendering that I could discern. Stops these build warnings.  Be advised one more file (at least) uses GtkStyle: eel-background.c and I have not been sucessful in porting that one to GtkStyleContext because that also requires porting GtkColor to GdkRGBA. Simply using the colors as-is in the first of two functions will build but only blue will be rendered. Thus, that file is unchanged from master for now.

### DIFF
--- a/src/caja-history-sidebar.c
+++ b/src/caja-history-sidebar.c
@@ -74,8 +74,13 @@ enum
 static void  caja_history_sidebar_iface_init        (CajaSidebarIface         *iface);
 static void  sidebar_provider_iface_init                (CajaSidebarProviderIface *iface);
 static GType caja_history_sidebar_provider_get_type (void);
+#if GTK_CHECK_VERSION (3, 0, 0)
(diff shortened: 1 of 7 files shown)


raveit65:
Ok, i just tested it and it seems fine.
As you're more expert than me i'm fine with removing the 'style-set' functions.
@monsta 
Can you please do a code review?

raveit65:
Btw. in caja git there are examples of porting GdkColor to RGBA_Colors.
http://git.mate-desktop.org/caja/log/?qt=grep&q=style

raveit65:
I merged the other one
https://github.com/mate-desktop/caja/pull/559
Thank you you again for your hard work on this.

Issue #124, opened 2013-05-14
Copy queue and pausing


handycz:
It would be handy if one could queue more copy operations so they would run serially. Also there should be an option to pause copying (whole queue or only one file)

alejandro-perez:
For the interface of queuing, I have a suggestions:

In the copy progress dialog, since you have a "row" per copy operation, and each copy operation has a single destination, you can drop files on top of these "rows", and files are queued to be copied to that very destination.

lukefromdc:
Fixed with the series of commits
https://github.com/mate-desktop/caja/commit/89058ed86d271f22cd9359c66e42ff9e8cfd4715
through
https://github.com/mate-desktop/caja/commit/9081b8d96936b625c3c83b88282ad7437f8ef462

PR #568, opened 2016-06-12
Implement copy queue and pausing

emanuele-f:
As many users (see #124), including me, have missed this feature from caja file manager for a long time, I've finally decided to try to implement a prototype of it. I've very little experience in glib/gtk framework so please watch out for errors!

copy

This pull request implements start/pause/queue logic for caja-file-operations. Since caja-file-operations uses GIO functions, a single file copy cannot be stopped in the middle. Instead, the actual stop is performed before handling next file.

sm

The buttons and internal events drive the state machine in figure. Progress widgets are visually sorted based on their current queue order.

- The start/pause button can be used to put the operation on sleep and to resume it later
  
- The queue button puts the operation in queue. When no more operations are in progress, the first operation of the queue is resumed
  
- The cancel button has been moved to the left side of the progress bar

I'm using some stock buttons, but I feel they are not appropriate.

Pausing and resuming execution is performed by waiting_c condition variable.

The operations can be started either as "running" or as "queued". Since the function caja_progress_info.c::new_op_started introduces a delay of 2 seconds before the windows showing up, move and copy operations now are not started until that point (this is a change of the previous behaviour, maybe a point to improove..).

It would be nice to implement user preferencies to define queue policy. For example, the policy QUEUE_ON_DEVICE_WRITING could indicate that a new operation should be queue if another operation exists which is writing the same device. Others could be QUEUE_ON_DEVICE_USED, QUEUE_ALWAYS, START_ALWAYS.

I hope you are interested in further developing this feature. I look forward for your feedback.


raveit65:
Nice feature, i just tested it.
But here with caja gtk3 version the pause button does not  work, i can't interrupt the copy action.
About GtkStock, in general we don't want to drop that deprecation, but in case of buttons without labels we should avoid that.
see this example https://github.com/mate-desktop/mate-notification-daemon/commit/fed38d6cf6f8141a8639723c6c5ccd5267fc5cc6

emanuele-f:
I've just tested gtk3 version on a debian stretch environment (mate 1.14.1 from official repositories, caja from my git repo) and the copy action worked as expected

emanuele-f:
Also tested on a fresh installation from latest github build and still no trouble. Please double check that you are not trying to pause a single-file transfer (because always next file is stopped). Can you give me more info? Do "pausing" or "pause" status even show up?

raveit65:
Sorry for late answering, we're are a small team and i will test it again if found some free time.
As far i remember copying files has now a delay from 1-2 secs before it starts, also with small text files.
Not sure if we can pull in your work for 1.16 because we was fighting against a gtk+-3.21 regression which caused non redraws on the desktop window, this is now particular fixed.
But now priority is on port to GtkAplication as debian wants to drop libunique for next release.
Sorry, your PR is welcome, hope i will find some time the next week.

emanuele-f:
I'm sad to hear the MATE team has to deal with such troubles, spending so much time about fixing things others break rather than working on new features. I know this work is necessary for MATE to keep itself relevant in software distributions while still maintain the look and usability we love. Thank you for your hard work.

That being said, I'm glad to hear my PR is welcome, and I wish to improve it.

I'd like to understand why the 2 seconds delay to popup the window was introduced in the first place: only to avoid flashing the screen for a small transfer, or to solve some kind of synchronization issues. Understanding that is crucial to remove that boring delay.

And I'd also really like to allow users to pause a file transfer any time: I see it hard to convince glib upstream developers about changing their code to support this, so maybe the easy way is to make a copy of the "copy" routine code here in MATE and patch it locally. Please tell me what you think about this.

raveit65:
> And I'd also really like to allow users to pause a file transfer any time: I see it hard to convince glib upstream developers about changing their code to support this, so maybe the easy way is to make a copy of the "copy" routine code here in MATE and patch it locally. Please tell me what you think about this.

that's OK

emanuele-f:
I'm sorry, I'm too busy with my thesis work right now. Hope to find some time soon.

lukefromdc:
I just tested this on Debian Unstable with GTK 3.22, seemed to work fine for pausing a directory full of video files (my normal use case for this). Cancelling a copy job works, that alone is a worthwhile addition.

There is an ugly bug, however on pausing copying a large single file. When paused, the file being copied is finished and the next file is held until the job is restarted. If a single large file is being copied however, it will pause near the end of the file (at least on flash media), then refuse to restart no matter what you do, and the partial copy is a corrupted file. A single test with the large file (a video that would not open unless fully copied) in a directory gave a finished copy of that file, and a hung paused job on the folder. The video could be played, indicating all of that non-web optimized MP4 had been copied but I still had to eject the test camera card to close the dialog.

In other words, that single file copy issue could generate "caja ate my files" bug reports unless fixed.

emanuele-f:
@lukefromdc, I really appreciate your testing and feedback. I'll work to fix this error and do more testing, thank you.

Regarding the "instant" pausing feature, I'm sorry but I think there is not a easy way to do this with glib. I've read some gio code, for copy and move operations it uses different backends based on OS configuration and file type. Only the [fallback implementation](https://github.com/djdeath/glib/blob/master/gio/gfile.c#L3110) provides a "plain read/write loop" which could be controlled.

I think forking and patching all this stuff is not really a good idea..

lukefromdc:
It's fine to finish a file, so long as that file gets all the way finished and the
dialog does not freeze after it. Nemo has pausing and queuing and it works,
so looking at the Nemo commits might give clues to fixing the dialog freeze.

I haven't done detail testing of single file pause results on Nemo, as that I've
only used, not tested.

Certainly forking and taking over any code for this is a bad idea, the MATE
team is already heavily burdened.

emanuele-f:
Nemo implements file queue but it does not allow to pause, resume or queue again, so they have done less modifications (still more elegantly...) to existing code than mine.
With commit 932beed I've modified dialog show logic, in order to both reduce the checks in the code and to eliminate the delay I introduced with previous commits. The problem is that queue logic is bound to the dialog code: I need a dialog to make things work.

Nevertheless, I'm trying to reproduce the bug on GTK2 with a USB stick but with no luck. Do the bug appeared with GTK3? I'm having troubles building a GTK3 environment, I cannot get rid of GTK2 symbols. Are there any virtualbox GTK3 images ready for testing?

lukefromdc:
This is with GTK3, my entire MATE environment is GTK3 and I cannot test
GTK2 over it

emanuele-f:
@lukefromdc I've done some test on GTK3, I notice nothing wrong so far.

When copying to usb stick, indeed the bar goes full quickly, countdown says 0 seconds left, the window stays open becouse the copy it's not finished yet. This same behaviour, however, can be also seen on upstream caja, and I think it's related once again to gio routines, plus OS buffering. I think this has nothing to do with this PR.

lukefromdc:
One thing that generates a very fast apparent copy is copy jobs going to cache,
then having to actually write to hardware. This is especially obvious on flash 
devices. What I was getting was no finish at all after a failed attempt to pause
a single-file copy, which does not work due to the limitations discussed.

lukefromdc:
This now seems to work properly. An attempt to pause a big single file simply does nothing and the file copies clean. Pausing and restarting muiltiple files works as expected. Would be nice to grey out the option to pause on one file, this might generate bug reports otherwise. If this can't be done I'm fine with merging this as-is and putting it in the documentation that pausing works only on multiple file jobs.

raveit65:
> Are there any virtualbox GTK3 images ready for testing?

fedora 25 or ubuntu-16.10 will use Mate gtk3 as default.
Beta releases are there, ie. fedora
https://spins.fedoraproject.org/prerelease

raveit65:
I just tested it again with copying 2 large files on an external usb drive.
After adding the second file, the second file is queued, which is good, but after starting it i can' pause it any more.
bildschirmfoto zu 2016-10-16_17-23-14
The dialog shows the second file is paused but it is still copying. 
Also pressing the queue buttons does not have any affect.
.......maybe i do not understand how it should work?

In my first test i noticed a delay before the copy action begun, but this seems to be fixed now.

emanuele-f:
@raveit65 yes the delay was fixed with 932beed. The behaviour you are reporting is expected, because I'm not able to pause a specific file while transfering.

@lukefromdc I'm doing some modifications to disable the controls, to make it clear when it is not possible to pause or queue a transfer.

Thank you for testing, stay tuned!

lukefromdc:
I hadn't noticed the delay but wasn't really looking for it. I normally test with copying into /tmp, which is on a tmpfs with a lot of RAM. Very fast copy so I have to work fast with multi-GB directories. Anyway, plently responsive to pause and queue with those big directories after the latest changes. Disabling the controls when a large single file is being transferred should make this ready to go,

emanuele-f:
Here is the list of changes with 96dcea5:
- Make progress bar stay zero when paused before file move
- Avoid file operations window popup when already open (e.g. when creating a new file)
- Disable controls where/when pausing cannot be applied
- Cancel button also disables pause and queue

Tested with GTK2 so far.

lukefromdc:
I just tested with GTK 3.22  https://github.com/mate-desktop/caja/commit/96dcea52556349decef63853e39f881ee550e33b

and found it mostly works with one issue: the FIRST file copy job from any newly opened directory cannot be paused as the controls are disabled even for a directory full of files. Second and subsequent copy jobs  from the same directory pause and queue just fine, going to a new directory brings the problem back. Disabling controls works as intended when a single large file is in question.

With this branch ending at the previous commit, pausing the first copy  of a directory full of files from a newly opened parent directory does work, so in this case the controls should be enabled.

Also there are no theme selectors to split out this dialog, which may be an issue with some themes now that it is more complex.

lukefromdc:
This patch makes the first multi-file copy job on a newly opened directory pausable at the expense of leaving the button sensitive when the last file is being copied. Pause button remains disabled on single-file job. Posted not for merge but as clue to where the problem may be. Somehow the code I deleted must have been getting executed on that first job from each new parent directory.

```
--- /home/luke/Desktop/caja-file-operations.c	2016-10-16 21:51:35.735584595 -0400
+++ /home/luke/Desktop/src/caja/caja-file-operations.c	2016-10-16 21:45:11.265705017 -0400
@@ -4989,9 +4989,6 @@ common = &job->common;
 		fallback = l->data;
 		src = fallback->file;
 
-		if ((!l->next) && (!is_dir(src)))
-			/* this is the last file, cannot pause anymore */
-			caja_progress_info_disable_pause (common->progress);
 
 		if (fallback->has_position) {
 			point = &fallback->position;
```

We need to find a fix that handles both cases, but this is what I am actually using at the moment

lukefromdc:
One minor quibble: it's really important that themes properly gray out images in disabled buttons with this. Setting an opacity value less than 1.0 is a simple way to do this if the background underneath is any kind of gray, just set this up in UbuntuStudio_Legacy

emanuele-f:
The bug should be fixed with 124df5f.

@lukefromdc can you handle the theme stuff after merge? I have no idea what to modify!

lukefromdc:
The theme stuff is simple, I'll take care of it. It's just a matter of ensuring that
images in disabled buttons are partially transparent so they can grey out.

lukefromdc:
OK, I just confirmed this works in GTK 3.22.

lukefromdc:
Rebased and merged ending with 
https://github.com/mate-desktop/caja/commit/9081b8d96936b625c3c83b88282ad7437f8ef462

Will now start checking themes

monsta:
I guess you guys can close https://github.com/mate-desktop/caja/issues/124 now.

lukefromdc:
https://github.com/mate-desktop/mate-themes/commit/e227d0ab9531813842716e8f4634493ef029ee39

applies the necessary theme fixes for mate-themes/master

emanuele-f:
Wonderful! Thank you!

raveit65:
Cool, but why using old and new syntax in themes? 
.......with a dot and w/o a dot.....css name and css style class.....
Or do we have something similar in firefox or libreoffice?

lukefromdc:
The dot is for a style class rather than a named widget or css node, and it
is used because Firefox needs it. Gtk 3.18 put style classes on almost 
everything, GTK 3.20 and 3.22 put them by default on only a few widgets
but they are widely used in programs that use GTK. The new (css node)
syntax is required to theme most widgets in GTK 3.20+ when the application
code does not expliclty add a theme class, which is most of the time.

raveit65:
@lukefromdc @monsta 
It look likes PR for overlapping desktop icons needs some work.
So, should we do a 1.17.0 release after all this changes beginning with GtkApplication port before we push 'making caja gtk3 only' to master.
For me all this changes needs a dev release for better testing.

raveit65:
> The dot is for a style class rather than a named widget or css node, and it is used because Firefox needs it. Gtk 3.18 put style classes on almost everything, GTK 3.20 and 3.22 put them by default on only a few widgets but they are widely used in programs that use GTK. The new (css node) syntax is required to theme most widgets in GTK 3.20+ when the application code does not expliclty add a theme class, which is most of the time.

I know the different between css name (node) and css style class and why we need it for ff and lo.
I was only surprised why you add ff workarounds here.

emanuele-f:
>It look likes PR for overlapping desktop icons needs some work.
So, should we do a 1.17.0 release after all this changes beginning with GtkApplication port before we push 'making caja gtk3 only' to master.
For me all this changes needs a dev release for better testing.

If possible, I'd like to see this feature in some caja release *before* going to GTK3 only. I know going to GTK3 is not really your free choice, but a need caused by too much compatibility work.
I really hate GTK3 look/accessibility so I don't know if I will update my mate environment further...

raveit65:
> I really hate GTK3 look/accessibility so I don't know if I will update my mate environment further...
 
We're not gnome-shell, i think most people won't see any different in look and feel of MATE.
And from my point of view as a theming artist i can only say that gtk3 is a big improvements.
gtk2 theming is something from the last millennium ;)

emanuele-f:
screenshot at 2016-10-18 14-23-22

How ugly is this GTK3 dialog (I use shiki colors theme)! I dislike the "blended" look, it feals like any application is alien, gorgeus, not functional! Should I consider myself a last millennium guy?

Ps. How much I love mate-calc <3, I rebuild it on any system I use. Galculator it's just terrible, buuu!
Pss. Sorry for the disgression

raveit65:
The small progressbar is a bug in gtk+-314, this is fixed with 3.16.
And there is space around the .linked buttons normal, looks like theme dev did remove that.
I ship also mate-calc for fedora.

lukefromdc:
The style class selectors were added to the theme changes because I wanted them to gray out icons in ALL disabled buttons and menu items, not just in this one dialog. Disabled items are much more obvious this way, especially when there is no text to gray out. The lack of text was what made it urgent for this change in Caja, which otherwise would have generated bug reports.

The style class did not work in Firefox, BTW. Neither the opacity used in my commits nor the mix fg color used in Menta had any effect in FF disabled menuitems. Still worth keeping around as it may be needed somwhere else and harms nothing.  FF is a nightmare to work with and Libreoffice such a mess I had to roll it back to GTK2 as my theme gave the white on white text in the calc formula bar and I could not find any selectors to fix it in LibreOffice.

raveit65:
> The style class did not work in Firefox, BTW. Neither the opacity used in my commits nor the mix fg color used in Menta had any effect in FF disabled menuitems. Still worth keeping around as it may be needed somwhere else and harms nothing.

Ok, good to know.

raveit65:
> (I use shiki colors theme)!

I think last update for this theme or the fork zukitwo colors was with gtk+-3.6 or 3.8  :)

monsta:
@emanuele-f: please have a look at recent reports: https://github.com/mate-desktop/caja/issues/844, https://github.com/mate-desktop/caja/issues/845, https://github.com/mate-desktop/caja/issues/846, https://github.com/mate-desktop/caja/issues/847.

PR #574, opened 2016-06-19
Avoid deprecated gdk_cairo_create

lukefromdc:
This is deprecated in GTK 3.21 as of the troublesome commit
https://git.gnome.org/browse/gtk+/commit/?id=ad78daaf8081d8ea72006832ee3dd5df2e1ed3f9

Replace it with just cairo_create, no adverse effects observed in tests with compiz or
Marco, with or without compositing.

This removes one more deprecated command in Caja, though it was not enough to fix the redraw issue in Caja with the newest GTK 3.21 master code. GTK development is crawling right not, not sure if they will keep those changes or not with so many issues. On the other hand, no issues I could find were created by replacing gdk_cairo_create, at least not in slightly older or in current GTK 3.21. Even if GTK rolls back their latest changes, using cairo_create() directly decouples this from future Gtk/Gdk changes to some extent.

Again, this is against the dev-style-context branch as that is what I am working with, though with some manual file editing (open in Pluma, replace every instance of "gdk_cairo_create" with "cairo_create" via search and replace, save each file and you are done) this could also be done with master, in which eel is quite different.

No idea what this would do in GTK2 but I am under the impression Caja 1.16 is not intended to support GTK2.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,16 @@
+Major changes in 1.15.0 are:
+
+* sync with Transiflex
+* remove old gnome files
+* GTK+-3 fm-icon-view: fix a Wincompatible-pointer-types build warning
+* GTK+-3 sidebars and caja-places: port style_set to style_updated
+* GTK+-3 caja-zoom-control: GtkStyle>GtkStyleContext
+* GTK+-3 caja-places-sidebar: use GtkStyleContext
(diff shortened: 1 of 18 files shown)


monsta:
Oops, merge conflicts...

lukefromdc:
Rebased this against todays' mate-desktop/master  as 
https://github.com/mate-desktop/caja/pull/579

PR #577, opened 2016-06-26
port to GtkStyleContext and GdkRGBA

raveit65:
Not just finished yet and help is needed!
Using a gradient background for the desktop is broken now.
Using a solid color works well. 
This PR needs to be used too
https://github.com/mate-desktop/mate-desktop/pull/230
and this branch from mate-control-center
https://github.com/mate-desktop/mate-control-center/pull/238

### DIFF
--- a/eel/eel-background.c
+++ b/eel/eel-background.c
@@ -80,7 +80,11 @@ struct EelBackgroundDetails
     MateBGCrossfade *fade;
     int bg_entire_width;
     int bg_entire_height;
+#if GTK_CHECK_VERSION (3, 0, 0)
+    GdkRGBA default_color;
+#else
     GdkColor default_color;
+#endif
 
     gboolean use_base;
 
@@ -166,7 +170,11 @@ eel_background_unrealize (EelBackground *self)
 
 static void
 make_color_inactive (EelBackground *self,
+#if GTK_CHECK_VERSION (3, 0, 0)
+		     GdkRGBA      *color)
+#else
 		     GdkColor      *color)
+#endif
 {
     double intensity, saturation;
     gushort t;
@@ -201,7 +209,11 @@ gchar *
 eel_bg_get_desktop_color (EelBackground *self)
 {
     MateBGColorType type;
+#if GTK_CHECK_VERSION (3, 0, 0)
+    GdkRGBA    primary, secondary;
+#else
     GdkColor   primary, secondary;
+#endif
     char      *start_color, *end_color, *color_spec;
     gboolean   use_gradient = TRUE;
     gboolean   is_horizontal = FALSE;
@@ -241,7 +253,11 @@ eel_bg_get_desktop_color (EelBackground *self)
 static void
 set_image_properties (EelBackground *self)
 {
+#if GTK_CHECK_VERSION (3, 0, 0)
+    GdkRGBA c;
+#else
     GdkColor c;
+#endif
 
     if (self->details->is_desktop && !self->details->color)
         self->details->color = eel_bg_get_desktop_color (self);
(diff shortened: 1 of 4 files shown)


raveit65:
thoughts from lukefromdc:

For the gradient background, it looks like eel_gradient_new in gdk_extensions.c may be the culprit, since it's the only real difference from the single color background in the function eel_bg_get_desktop_color in eel_background.c

Just took a look at eel_gradient_new, the color variables there are const char variables, while in eel_bg_get_desktop_color they are GdkRGBA 's. eel_bg_get_desktop_color loads new "char" variables from the RBGA's, wonder if the format put into these variables in not valid for eventually making the colors?

Now playing with the code below:

#if GTK_CHECK_VERSION (3, 0, 0)
    start_color = eel_gdk_rgb_to_color_spec (eel_gdk_rgba_to_rgb (&primary));

    if (use_gradient)
    {
        end_color  = eel_gdk_rgb_to_color_spec (eel_gdk_rgba_to_rgb (&secondary));
#else
    start_color = eel_gdk_rgb_to_color_spec (eel_gdk_color_to_rgb (&primary));

    if (use_gradient)
    {
        end_color  = eel_gdk_rgb_to_color_spec (eel_gdk_color_to_rgb (&secondary));
#endif
        color_spec = eel_gradient_new (start_color, end_color, is_horizontal);
        g_free (end_color);
    }
    else
    {
        color_spec = g_strdup (start_color);
    }
    g_free (start_color);

    return color_spec;``

The start_color and end_color variables both work, as simply setting color_spec = g_strdup to either color works properly. It's when `color_spec = eel_gradient_new ()` is called that a black background results, thus confirming that eel-gdk-extensions.c needs work on the gradient code.

The special case of a gradient where both colors are identical does work, it's in other cases where `eel_gradient_new()` breaks down.
`
 return g_strconcat (start_color, "-", end_color, is_horizontal ? ":h" : NULL, NULL);`
is not giving a useful result here, as replacing it with 
`return g_strdup (end_color);`
causes the expected result: end color only is used as the background.

https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strconcat

warns that this function often assembles a string other than in the order you want. My guess is that the char variables loaded from the GdkRBGA's when put into this function are returning a value that is formatted differently than GkdColor was giving and is not useful for setting up a gradient. 

using this test code:
`return g_strconcat (start_color, NULL, is_horizontal ? ":h" : NULL, NULL);`
Will force a color background to the start color whether solid or either gradient is selected. This shows that the start_color-end_color concatenated string is what is no longer valid here.

Thus, the returned value "color_spec" back in eel_background.c is coming back as a string that cannot be interpreted by another function due to a format change between GdkColor and GdkRBGA.

raveit65:
@monsta @XRevan86 @dnk
Can you please help?

raveit65:
merged
https://github.com/mate-desktop/caja/commit/6eb433420c347de2384dcfe3dbbd7b5fc97117da
https://github.com/mate-desktop/caja/commit/36050c2e84626e52c11790cdc706079f6861c4a5
https://github.com/mate-desktop/caja/commit/75659aa8d36813d4228949eeeafff317b71c5b93
https://github.com/mate-desktop/caja/commit/d829c648eb20daa6ad487768195e1379a2ddb135
https://github.com/mate-desktop/caja/commit/e983f64abd52242d96d9ed3825395f2e1f3dea5e
https://github.com/mate-desktop/caja/commit/0c08ec62529640df8e6c007af8ed3e4728d31e69
https://github.com/mate-desktop/caja/commit/3b8e0d41a7adca11014a3548f3ab19fd7e400c70

Issue #575, opened 2016-06-20
GTK 3.21.3: no desktop redraws or bg changes until restart


lukefromdc:
Beginning with the various  GdkDrawingContext changes to GTK during the 3.21.2 cycle, the desktop background in Caja has become "persistant" with a failure to redraw and "erase" when any icon is moved. In addition, all icon and text shadows are rendered at 100% opacity and the background cannot be changed. Restarting caja or adding/removing a secondary monitor will forcibly redraw the desktop, erase no longer valid images, and apply background changes.  Nemo also has this problem, nautilus with the new split-out "nautilus-desktop" binary does not and in fact running an instance of nautilus-desktop in a MATE session with normal desktop icons turned off is a workaround for this.

In searching for this, I have gotten as far as the function "draw_canvas_background" in caja-icon-container.c . If this function and the reference to it elsewhere in the same file are removed, no background can be rendered (as expected) but icons can now be moved around and shadows render normally.  

The desktop background failed redraw issue exists in Master(where I first saw it), in dev-style-properties, and even in my dev-deprecations-3.21 branch where the newly deprecated gdk_cairo_create() was replaced with cairo_create()

The buggy GTK changes have now made it out the door in GTK 3.21.3, implying they are now far more likely to persist all the way to final GTK 3.22. This broke a lot of things, like causing Firefox to render solid black until each widget is focussed, but it now looks like GTK devs expect all of us app developers to find and fix the major fallout from these gdk changes ourselves.

lukefromdc:
Just found the exact GTK commit causing this problem:
https://git.gnome.org/browse/gtk+/commit/?id=2c7b21718f18ab743cd79784ae28cad582f58503
"Simplify the widget rendering entry point"
lots of internal changes in gtkwidget internal cairo interaction

lukefromdc:
More experiments with Caja and GTK 3.21/3 tonight.

killing mate-settings-daemon or (in another DE) starting mate-settings-daemon, cinnamon-settings-daemon, or gnome-settings-daemon will force the background to redraw and "clean" the desktop. At this point icons and text render normally. Clicking on the desktop will darken shadows until after about 4 clicks they are opaque. This or any other interaction redraws all icons and text without erasing.  Behavior is as though on each redraw the entire desktop canvas was composited with the last copy of itself rather than the background. The failed redraw also blocks background changes, though that will work with icons on the desktop turned off.

raveit65:
Do we see gtk4 here?
[...]

PR #582, opened 2016-07-02
Avoid deprecated gdk_cairo_create (rebased again)

lukefromdc:
This is a rebase of
https://github.com/mate-desktop/caja/pull/574
against mate-desktop/master (release 1.15.1) current as of Sat Jul3  6 PM EDT after dev-style-context was merged

gdk_cairo_create is deprecated in GTK 3.21 as of the troublesome commit
https://git.gnome.org/browse/gtk+/commit/?id=ad78daaf8081d8ea72006832ee3dd5df2e1ed3f9

Replace it with cairo_create, no adverse effects observed in tests with compiz or
Marco, with or without compositing.

This removes one more deprecated command in Caja, though it was not enough to fix the redraw issue in Caja with the newest GTK 3.21 master code. GTK development is crawling right not, not sure if they will keep those changes or not with so many issues. On the other hand, no issues I could find were created by replacing gdk_cairo_create, at least not in slightly older or in current GTK 3.21. Even if GTK rolls back their latest changes, using cairo_create() directly decouples this from future Gtk/Gdk changes to some extent.

No idea what this would do in GTK2.  If this does not work in GTK2, each replacement of gdk_cairo_create with cairo_create could be put behind a GTK3 selector but from what I am hearing caja 1.16 is intended to be GTK3 only so I just did a simple find-replace in Pluma on each file.


XRevan86:
```c
cairo_t *cairo_create(cairo_surface_t *target);
cairo_t *gdk_cairo_create(GdkWindow *window);
```
You can't possibly replace one with another without consequences (GdkWindow is not a cairo surface).
And commit description doesn't even suggest that.

lukefromdc:
What do you suggest doing? I did exactly that and Caja with this has been running
without any problems I can find for two weeks.
I don't know of any other fix for this. If this creates problems or is just bad code,
we can close this one and leave it for someone else to find a solution for this.

I first tried this hoping it was the issue behind the failure to redraw the desktop
with GTK 3.21.3
https://github.com/mate-desktop/caja/issues/575
but no such luck. No new issues but didn't help either. I submitted the PR because
this is one less deprecated command but if it creates problems, build warnings,
or runtime warnings feel free to close it.

XRevan86:
I'm not sure (didn't test it) but I think this is how gdk_cairo_create() supposed to be replaced:
```c
cairo_region_t *region = gdk_window_get_clip_region(window);
GdkDrawingContext *context = gdk_window_begin_draw_frame(window, region);
cr = gdk_drawing_context_get_cairo_context(context);
<..>
gdk_window_end_draw_frame(window, context);
```

lukefromdc:
I won't be able to properly play with that until tomorrow night
but will see if I can get that working. Wonder if what I'm using
now works because of some corner case or undocumented 
feature which should never be used?

lukefromdc:
OK, I just found out what's going on: It turns out Caja only uses gdk_cairo_create
in GTK2 already! Every single instance was behind GTK2 only selectors, in some
cases very widely separated and away from the code in question. Also, some of 
the commenting wasn't very good, with this format used in worst cases:

#if GTK_CHECK_VERSION(3,0,0)
#else /* GTK_CHECK_VERSION(3,0,0) */
#endif /* GTK_CHECK_VERSION(3,0,0) */

I will now close this PR as it has no effect on the code built in GTK3 and probably
has no effect at all but to break GTK2 builds.

lukefromdc:
Closing this as it does not change code actually built with GTK3, presumably breaks GTK2 builds.

PR #583, opened 2016-07-04
deprecations fixes

raveit65:
Well, i was quite busy ;)

@monsta @flexiondotorg @lukefromdc 
As always, please test and review.

### DIFF
--- a/libcaja-private/caja-file-conflict-dialog.c
+++ b/libcaja-private/caja-file-conflict-dialog.c
@@ -584,7 +584,11 @@ diff_button_clicked_cb (GtkButton *w,
 static void
 caja_file_conflict_dialog_init (CajaFileConflictDialog *fcd)
 {
+#if GTK_CHECK_VERSION (3, 0, 0)
+    GtkWidget *hbox, *vbox, *vbox2;
+#else
     GtkWidget *hbox, *vbox, *vbox2, *alignment;
+#endif
     GtkWidget *widget, *dialog_area;
     CajaFileConflictDialogDetails *details;
     GtkDialog *dialog;
@@ -619,11 +623,19 @@ caja_file_conflict_dialog_init (CajaFileConflictDialog *fcd)
     details->titles_vbox = widget;
 
     /* Setup the hboxes to pack file infos into */
+#if GTK_CHECK_VERSION (3, 0, 0)
+    vbox2 = gtk_vbox_new (FALSE, 12);
+    gtk_widget_set_halign (vbox2, GTK_ALIGN_START);
+    gtk_widget_set_valign (vbox2, GTK_ALIGN_START);
+    gtk_widget_set_margin_start (vbox2, 12);
+    gtk_box_pack_start (GTK_BOX (vbox), vbox2, FALSE, FALSE, 0);
+#else
     alignment = gtk_alignment_new (0.0, 0.0, 0.0, 0.0);
     g_object_set (alignment, "left-padding", 12, NULL);
     vbox2 = gtk_vbox_new (FALSE, 12);
     gtk_container_add (GTK_CONTAINER (alignment), vbox2);
     gtk_box_pack_start (GTK_BOX (vbox), alignment, FALSE, FALSE, 0);
+#endif
 
     hbox = gtk_hbox_new (FALSE, 12);
     gtk_box_pack_start (GTK_BOX (vbox2), hbox, FALSE, FALSE, 0);
@@ -656,7 +668,11 @@ caja_file_conflict_dialog_init (CajaFileConflictDialog *fcd)
     g_signal_connect (widget, "clicked",
                       G_CALLBACK (reset_button_clicked_cb), dialog);
 
+#if GTK_CHECK_VERSION (3, 0, 0)
+    gtk_widget_show_all (vbox2);
+#else
     gtk_widget_show_all (alignment);
+#endif
 
     /* Setup the diff button for text files */
     details->diff_button = gtk_button_new_with_label (_("Differences..."));
--- a/src/file-manager/fm-properties-window.c
+++ b/src/file-manager/fm-properties-window.c
@@ -1980,7 +1980,9 @@ attach_combo_box (GtkTable *table,
                   gboolean two_columns)
 {
 	GtkWidget *combo_box;
+#if !GTK_CHECK_VERSION (3, 0, 0)
 	GtkWidget *aligner;
+#endif
 
 	if (!two_columns) {
 		combo_box = gtk_combo_box_text_new ();
@@ -1998,6 +2000,9 @@ attach_combo_box (GtkTable *table,
 					       "text", 0);
 
 	}
+#if GTK_CHECK_VERSION (3, 0, 0)
+	gtk_widget_set_halign (combo_box, GTK_ALIGN_START);
+#endif
 	gtk_widget_show (combo_box);
 
   	gtk_combo_box_set_row_separator_func (GTK_COMBO_BOX (combo_box),
@@ -2005,6 +2010,10 @@ attach_combo_box (GtkTable *table,
 					      NULL,
 					      NULL);
 
+#if GTK_CHECK_VERSION (3, 0, 0)
+	gtk_grid_attach_next_to (grid, combo_box, sibling,
+				 GTK_POS_RIGHT, 1, 1);
+#else
 	/* Put combo box in alignment to make it left-justified
 	 * but minimally sized.
 	 */
@@ -2012,10 +2021,6 @@ attach_combo_box (GtkTable *table,
 	gtk_widget_show (aligner);
 
 	gtk_container_add (GTK_CONTAINER (aligner), combo_box);
-#if GTK_CHECK_VERSION (3, 0, 0)
-	gtk_grid_attach_next_to (grid, aligner, sibling,
-				 GTK_POS_RIGHT, 1, 1);
-#else
 	gtk_table_attach (table, aligner,
 			  VALUE_COLUMN, VALUE_COLUMN + 1,
 			  row, row + 1,
@@ -3669,8 +3674,8 @@ create_basic_page (FMPropertiesWindow *window)
 	GtkGrid *grid;
 #else
 	GtkTable *table;
-#endif
 	GtkWidget *icon_aligner;
+#endif
 	GtkWidget *icon_pixmap_widget;
 	GtkWidget *volume_usage;
 	GtkWidget *hbox, *vbox;
@@ -3685,6 +3690,13 @@ create_basic_page (FMPropertiesWindow *window)
 
 	icon_pixmap_widget = create_image_widget (
 		window, should_show_custom_icon_buttons (window));
+#if GTK_CHECK_VERSION (3, 0, 0)
+	gtk_widget_set_halign (icon_pixmap_widget, GTK_ALIGN_END);
+	gtk_widget_set_valign (icon_pixmap_widget, GTK_ALIGN_START);
+	gtk_widget_show (icon_pixmap_widget);
+
+	gtk_box_pack_start (GTK_BOX (hbox), icon_pixmap_widget, FALSE, FALSE, 0);
+#else
 	gtk_widget_show (icon_pixmap_widget);
 
 	icon_aligner = gtk_alignment_new (1, 0, 0, 0);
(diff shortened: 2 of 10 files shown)


src/file-manager/fm-properties-window.c
```diff
@@ -3685,13 +3690,20 @@ create_basic_page (FMPropertiesWindow *window)
...
 		window, should_show_custom_icon_buttons (window));
+#if GTK_CHECK_VERSION (3, 0, 0)
+	gtk_widget_set_valign (icon_pixmap_widget, GTK_ALIGN_START);
+	gtk_widget_show (icon_pixmap_widget);
+
+	gtk_box_pack_start (GTK_BOX (hbox), icon_pixmap_widget, FALSE, FALSE, 0);
+#else
 	gtk_widget_show (icon_pixmap_widget);
 
 	icon_aligner = gtk_alignment_new (1, 0, 0, 0);
```
monsta: The first argument here means xalign is at the end... guess it should be done with halign in GTK+3 code.

libcaja-private/caja-file-conflict-dialog.c
```diff
@@ -619,11 +623,18 @@ caja_file_conflict_dialog_init (CajaFileConflictDialog *fcd)
     details->titles_vbox = widget;
 
     /* Setup the hboxes to pack file infos into */
+#if GTK_CHECK_VERSION (3, 0, 0)
+    vbox2 = gtk_vbox_new (FALSE, 12);
+    gtk_widget_set_halign (vbox2, GTK_ALIGN_START);
+    gtk_widget_set_margin_start (vbox2, 12);
+    gtk_box_pack_start (GTK_BOX (vbox), vbox2, FALSE, FALSE, 0);
+#else
     alignment = gtk_alignment_new (0.0, 0.0, 0.0, 0.0);
```
monsta: Second argument means yalign is at start - should be probably set with valign in GTK+3?

src/file-manager/fm-properties-window.c
raveit65: indeed, from logic point it should be
gtk_widget_set_halign (icon_pixmap_widget, GTK_ALIGN_END);
gtk_widget_set_valign (icon_pixmap_widget, GTK_ALIGN_START);
but upstream commit use only valign.
Need to think about and test if it makes a different.

libcaja-private/caja-file-conflict-dialog.c
raveit65: Same here, i took it from upstream. I will test if adding
gtk_widget_set_valign (vbox2, GTK_ALIGN_START);
make  a different.

src/file-manager/fm-properties-window.c
raveit65: same as other commits

lukefromdc:
Just built and installed, most things fine except for text in eel-editable-label being offset to the top as far as it can go. This text appears smaller as a result, though that may be an optical illusion. Got this with both GTK3.20 and 3.21 and with every theme tested including Adwaita (so not a GTK theme bug).

eel-editable-label

raveit65:
@monsta 
file-conflict-dialog and fm-properties-window commits a updated, but honestly i see no different.

raveit65:
> Just built and installed, most things fine except for text in eel-editable-label being offset to the top as far as it can go. This text appears smaller as a result, though that may be an optical illusion. Got this with both GTK3.20 and 3.21 and with every theme tested including Adwaita (so not a GTK theme bug).Example below, same problem on desktop too:

yeah, i noticed this too.
Replacements for Alignments aplies to left and top side.
Ideally, we should have the same space to all sides.
.....thinking about.

raveit65:
'GTK+-3 eel-editable-label: drop usage of GtkMisc'
 is updated, i guess this is what we want.
editable-label-1
editable-label-2
For me this looks better than before this PR.

lukefromdc:
I don't see a new version on github yet, only the same one listed as
8 days old that I pulled a few hours ago

raveit65:
hmm, this is different from first attempt
```
        gint         position);
 static gint     editable_get_position         (GtkEditable *editable);
 
+#if GTK_CHECK_VERSION (3, 0, 0)
+G_DEFINE_TYPE_WITH_CODE (EelEditableLabel, eel_editable_label, GTK_TYPE_WIDGET,
+#else
 G_DEFINE_TYPE_WITH_CODE (EelEditableLabel, eel_editable_label, GTK_TYPE_MISC,
+#endif
                          G_IMPLEMENT_INTERFACE (GTK_TYPE_EDITABLE, eel_editable_label_editable_init));
 
 static void
@@ -1329,7 +1333,9 @@ get_layout_location (EelEditableLabel  *label,
                      gint      *xp,
                      gint      *yp)
 {
+#if !GTK_CHECK_VERSION (3, 0, 0)
     GtkMisc *misc;
+#endif
     GtkWidget *widget;
     gfloat xalign, yalign;
     GtkRequisition req;
@@ -1339,9 +1345,18 @@ get_layout_location (EelEditableLabel  *label,
 #endif
     GtkAllocation allocation;
 
+#if !GTK_CHECK_VERSION (3, 0, 0)
     misc = GTK_MISC (label);
+#endif
     widget = GTK_WIDGET (label);
+#if GTK_CHECK_VERSION (3, 0, 0)
+    margin_start = gtk_widget_get_margin_start (widget);
+    margin_end = gtk_widget_get_margin_end (widget);
+    margin_top = gtk_widget_get_margin_top (widget);
+    margin_bottom = gtk_widget_get_margin_bottom (widget);
+#else
     gtk_misc_get_alignment (misc, &xalign, &yalign);
+#endif
 
     if (gtk_widget_get_direction (widget) != GTK_TEXT_DIR_LTR)
         xalign = 1.0 - xalign;
```

raveit65:
btw. i rebase the commit, so no differ in date.

lukefromdc:
Much better now, I just built this and tested against the last good version of GTK 3.21 and it works fine and centers the renaming labels. This seems ready to go on this end

eel/eel-editable-label.c
```diff
@@ -1339,9 +1345,18 @@ get_layout_location (EelEditableLabel  *label,
...
+    margin_start = gtk_widget_get_margin_start (widget);
+    margin_end = gtk_widget_get_margin_end (widget);
+    margin_top = gtk_widget_get_margin_top (widget);
+    margin_bottom = gtk_widget_get_margin_bottom (widget);
+#else
     gtk_misc_get_alignment (misc, &xalign, &yalign);
+#endif
 
     if (gtk_widget_get_direction (widget) != GTK_TEXT_DIR_LTR)
         xalign = 1.0 - xalign;
```
monsta: Well, xalign is used here and both xalign and yalign are used near the end of the function... but in GTK+3 build, xalign and yalign aren't initialized :confused:

eel/eel-editable-label.c
```diff
@@ -1339,9 +1345,18 @@ get_layout_location (EelEditableLabel  *label,
 #endif
     GtkAllocation allocation;
 
+#if !GTK_CHECK_VERSION (3, 0, 0)
     misc = GTK_MISC (label);
+#endif
     widget = GTK_WIDGET (label);
+#if GTK_CHECK_VERSION (3, 0, 0)
```
monsta: All ```margin_*``` variables are assigned here, but in the block a few lines below they're assigned once again, so the values assigned here aren't used at all...

monsta:
@lukefromdc: did you notice any changes in scrolling behavior in the icon view? I just wonder if anything would change after https://github.com/mate-desktop/caja/pull/583/commits/662ba8c3442b934a31d1c872632b5a323f4b09c4 and https://github.com/mate-desktop/caja/pull/583/commits/cea36abb48aec0cfcb6a12b311d3c8ff0a4468da because of this comment in the removed code:

```
/* The standard gtk_adjustment_set_value ignores page size, which
 * disagrees with the logic used by scroll bars, for example.
 */
```

I don't know if this statement is true or not these days.

eel/eel-editable-label.c
raveit65: Yes, removing the 2 lines here was the trick to get this working.

eel/eel-editable-label.c
raveit65: I will test to remove the first block complete as it does nothing if the variables aren't assigned there.

raveit65:
@monsta 
result before this PR
editable-label-before
with using xalign and yalign
editable-label-with alignment
and w/o alignment
editable-label-2
So, i think it's better not to use xalign and yalign here because it's move the selected text to the right and top.
It is enough only to use xpad and ypad which is set some lines below.
This shows also that padding wasn't working before this commit.
I will also remove the xalign and yalign for RTL direction and push the results in an additional commit.

raveit65:
@monsta 
Ok, this works.
If you agree i will squash the last commit.

monsta:
As I said before, xalign and yalign are used near the end of that function:
```C
    x = floor (xpad
               + ((allocation.width - req.width) * xalign)
               + 0.5);

    y = floor (ypad
               + ((allocation.height - req.height) * yalign)
               + 0.5);

    if (xp)
        *xp = x;

    if (yp)
        *yp = y;
```

In GTK+3 build xalign and yalign are not initialized before that block. As a result, after the call of this function the output variables (xp and yp) will contain random values.

raveit65:
Well, but the result is fine or not?
I don't see  random values in practise.
Feel free to commit on this branch here to remove the xalign and yalign from this lines.

monsta:
I'll test it with GTK+3 later to see how it looks.

The problem with uninitialized variables is that you might see how everything works fine in your system, but on some other one there might be some weird bugs (which are usually hard to track).

Here we have these two vars initialized in GTK+2 build via this function call:
```C
gtk_misc_get_alignment (misc, &xalign, &yalign);
```
In GTK+3, there's no any initialization for them.

raveit65:
```
The “xalign” property

  “xalign”                   gfloat

Horizontal position of child in available space. A value of 0.0 will flush the child left (or right, in RTL locales); a value of 1.0 will flush the child right (or left, in RTL locales).

GtkAlignment:xalign has been deprecated since version 3.14 and should not be used in newly-written code.

Use gtk_widget_set_halign() on the child instead

Flags: Read / Write

Allowed values: [0,1]

Default value: 0.5
```
from https://developer.gnome.org/gtk3/unstable/GtkAlignment.html#GtkAlignment--xalign
I don't see the we can get random values.
With latest commit:
editable-label-3

raveit65:
Maybe you like this ?

raveit65:
Now it looks like
editable-label-4
Not so perfect than previous but i can live with it.

raveit65:
rebased and using now
gtk_widget_get_halign (widget);
gtk_widget_get_valign (widget);

lukefromdc:
Seemed to scroll fine, and if I selected an item to rename and then 
scrolled no obvious problems.

lukefromdc:
Seems to be working fine over here other than the usual issues with GTK 3.21.3

eel/eel-editable-label.c
```diff
@@ -1349,9 +1353,21 @@ get_layout_location (EelEditableLabel  *label,
     misc = GTK_MISC (label);
 #endif
     widget = GTK_WIDGET (label);
-#if !GTK_CHECK_VERSION (3, 0, 0)
+#if GTK_CHECK_VERSION (3, 0, 0)
+    margin_start = gtk_widget_get_margin_start (widget);
+    margin_end = gtk_widget_get_margin_end (widget);
+    margin_top = gtk_widget_get_margin_top (widget);
+    margin_bottom = gtk_widget_get_margin_bottom (widget);
+    gtk_widget_get_halign (widget);
```
monsta: These two functions return a value but it's not assigned to any variable here...

src/caja-navigation-window.c
```diff
@@ -76,6 +76,10 @@
  */
 #include "caja-desktop-window.h"
 
+#if GTK_CHECK_VERSION (3, 0, 0)
```
monsta: This is already done several lines below in this file...

eel/eel-editable-label.c
raveit65: Please fix it, i have no idea to do this.

monsta: Yes, I'll check it later. I suggest moving all changes to ```eel/eel-editable-label.c``` to another branch and merge PR w/o them (as other changes seem ok).

raveit65: done

raveit65:
Now, top padding is missing.
I will create another branch for editable-lable, where we can work on,
and merge the other commits in master.

PR #586, opened 2016-07-08
deprecations and warnings

raveit65:
I know GMutex/GThread API is already fixed in caja, but this way it fixes a Wimplicit-function-declaration build warning in caja-search-engine-simple
@monsta @lukefromdc @flexiondotorg 
Please test and review.

### DIFF
--- a/src/caja-spatial-window.c
+++ b/src/caja-spatial-window.c
@@ -77,6 +77,7 @@
 
 #if GTK_CHECK_VERSION (3, 0, 0)
 #define gtk_hbox_new(X,Y) gtk_box_new(GTK_ORIENTATION_HORIZONTAL,Y)
+#define gtk_vbox_new(X,Y) gtk_box_new(GTK_ORIENTATION_VERTICAL,Y)
 #endif
 
 struct _CajaSpatialWindowDetails
--- a/libcaja-private/caja-file-changes-queue.c
+++ b/libcaja-private/caja-file-changes-queue.c
@@ -25,14 +25,6 @@
 
 #include "caja-directory-notify.h"
 
-#ifdef G_THREADS_ENABLED
-#define MUTEX_LOCK(a)	if ((a) != NULL) g_mutex_lock (a)
-#define MUTEX_UNLOCK(a)	if ((a) != NULL) g_mutex_unlock (a)
-#else
-#define MUTEX_LOCK(a)
-#define MUTEX_UNLOCK(a)
-#endif
-
 typedef enum
 {
     CHANGE_FILE_INITIAL,
@@ -57,9 +49,7 @@ typedef struct
 {
     GList *head;
     GList *tail;
-#ifdef G_THREADS_ENABLED
     GMutex mutex;
-#endif
 } CajaFileChangesQueue;
 
 static CajaFileChangesQueue *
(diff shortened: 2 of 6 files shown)


lukefromdc:
Seems to work fine here. No change in Gtk3.21.3 desktop situation but none expected. Everything else fine.

src/caja-spatial-window.c
```diff
@@ -64,6 +64,7 @@
 
 #if !GTK_CHECK_VERSION(3,0,0)
 #define gtk_widget_get_preferred_size(x,y,z) gtk_widget_size_request(x,y)
+#define gtk_vbox_new(X,Y) gtk_box_new(GTK_ORIENTATION_VERTICAL,Y)
```
monsta: This should be placed in a block several lines lower, along with ```gtk_hbox_new``` define. For now, it breaks GTK+2 build.

raveit65: Are sure?, this is what we use all over the code and not the new attempt from Sorkin :)
Still wondering that something which is used since 2 years in our code is bad now :)

raveit65: this looks different to me :smile: 
https://github.com/mate-desktop/mate-screensaver/commit/dd837483d4435bfac418ef9c006cff642990a83f

raveit65:
updated, anything else?

monsta:
Thanks, other changes are fine. Works alright in GTK+2 build here.

lukefromdc:
One of those last changes totally broke desktop rendering in at least GTK 3.18 and 3.20. I got the default window background (grey in my theme) drawn on top of both the desktop and the icons except for a border around the edge about the size of one icon.

lukefromdc:
Commit https://github.com/mate-desktop/caja/commit/d866d7a3a5b778cbd413eaa67ad69c80e9277c6b is the cause of the desktop being covered over with the default window color. The Mutex work is fine

raveit65:
hmm, here the desktop render fine with gtk+-3.20,
not sure if i understand what you mean.
Do you mean the desktop window or the navigation window?
Also you mentioned it was all fine some days ago.
But i will rebuild caja from master, maybe something was wrong with latest rebase before merge.

lukefromdc:
The desktop window, not the navigation window, is the culprit here, and I tested both GTK 3.20 and 3.21. Used newly downloaded tarballs of both master and of dev-locationbar, got same results. Then reverted https://github.com/mate-desktop/caja/commit/d866d7a3a5b778cbd413eaa67ad69c80e9277c6b and the desktop worked again,

The build itself was over GTK 3.21.3, wonder if that made a difference.

lukefromdc:
On still further testing, manually adding back 
`#define gtk_vbox_new(X,Y) gtk_box_new(GTK_ORIENTATION_VERTICAL,Y)`
to caja-spatial-window.c from the rolled back version brought back the rendered over desktop. Wonder if that box what is actually picking up the BG color and previously was not defined?  Does not seem to be theme-sensitive, got the bg color in every theme I have

raveit65:
wtf, now i see it here too.
So this is the culprit ?
https://github.com/mate-desktop/caja/pull/586/commits/d866d7a3a5b778cbd413eaa67ad69c80e9277c6b
Let me check.

raveit65:
Ok, confirmed.
I will revert this commit and test if using a GtkCheck instead of a define is better,
otherwise we have to live with that deprecation.
Still wondering why such a standard define which we use all over the code does not work here. Maybe because of missing port GtkTable to GtkGrid?
I tried this several times, but all what i see after using GtkGrid is exactly that what we see here :).
At first time the define was at the wrong place in this PR, for that reason i didn't saw the issue.
Thanks for figuring this out.

PR #589, opened 2016-07-11
locationbar: improvements and deprecations

raveit65:
Note, first commit is needed to fix deprecated GTK_TYPE_HBOX usage without a fallout which was done in a previous attempt.
And this affect only the bar with the location entry and not the button bar, which called pathbar.
Hope this works with gtk2.
@monsta @lukefromdc 
please test and review again.


monsta:
Very nice, it gets rid of deprecation and doesn't cause the expand issue. Works fine in GTK+2 for me.

lukefromdc:
Location bar worked fine with this, but the underlying master upon which this was based has a serious problem and covers up the desktop with the window default color:
https://github.com/mate-desktop/caja/issues/591

raveit65:
@lukefromdc 
So locationbar changes are fine after reverting the other culprit commit?

lukefromdc:
I haven't had any location bar trouble at all and am using that right now.

PR #592, opened 2016-07-12
pathbar: improvements and deprecations

raveit65:
Beside from deprecation fixes and adding some style classes to the buttons, this is a very little design change.
I want to show the right slider(arrow) button every time.
This makes it possible to use a design what you know from nemo or nautilus with some themes, but without using the link button style class, using link buttons on pathbar is still broken since 3.16.
With using TOK or BlackMate themes it looks like this.
bildschirmfoto zu 2016-07-12 10-54-03
Menta or Submarine themes will obtain this pathbar design
bildschirmfoto zu 2016-07-12 10-54-30
I hope you agree with making the slider button always visible.

@monsta @lukefromdc @flexiondotorg 
please test

### DIFF
--- a/src/caja-pathbar.c
+++ b/src/caja-pathbar.c
@@ -90,7 +90,9 @@ struct _ButtonData
 
     GtkWidget *image;
     GtkWidget *label;
+#if !GTK_CHECK_VERSION (3, 0, 0)
     GtkWidget *alignment;
+#endif
     guint ignore_changes : 1;
     guint file_is_hidden : 1;
     guint fake_root : 1;
@@ -1589,7 +1591,7 @@ set_label_size_request (ButtonData *button_data)
 
     pango_layout_get_pixel_size (layout, &bold_width, &bold_height);
 
-    gtk_widget_set_size_request (button_data->alignment,
+    gtk_widget_set_size_request (button_data->label,
         			 MAX (width, bold_width),
         			 MAX (height, bold_height));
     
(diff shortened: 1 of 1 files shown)


flexiondotorg:
I very much like this new design :+1:

clefebvre:
+1. I don't like it personally, but it has to be done to work around an issue in recent versions of GTK. It was done for Nemo 3.0.

lukefromdc:
Seems OK with my theme, a disabled forward button is greyed out, I don't
see any problem here.  If a theme author wants to hide a disabled button,
it can be done by turning it and everything in it transparent.

raveit65:
rebased for changes in master

src/caja-pathbar.c
```diff
@@ -90,6 +90,7 @@ struct _ButtonData
 
     GtkWidget *image;
     GtkWidget *label;
+    GtkWidget *alignment;
```
monsta: minor thing: can be marked as GTK+3 only :smile:

monsta: sorry, that was before I noticed it's already done in the third commit

src/caja-pathbar.c
```diff
@@ -719,7 +719,7 @@ caja_path_bar_size_allocate (GtkWidget     *widget,
     GtkRequisition child_requisition;
     GtkAllocation widget_allocation;
 
-    need_sliders = FALSE;
+    need_sliders = TRUE;
```
monsta: not sure why, but upstream commit has one more change below this...

monsta:
I don't like how it looks, but otherwise it builds and works ok with GTK+2 here.

monsta:
What exactly is broken in GTK+3 build without this PR? Pathbar seems to work fine in my VM with 3.20, I've tried TraditionalGreen and Menta.

src/caja-pathbar.c
raveit65: this commit is not 100% like upstream, i nneded to adjust it a bit for get it working.

raveit65:
Broken is a great word :)
The point is that i needed to do a ugly hack with using margin and margins at other places to revert the first margin at a wrong place, at several places for styling Menta and Submarine themes. see mate-applications.css under pathbar.
Adding style different classes for the buttons make it easy to do this.
And without making the right arrow button visible such a styling looks a bit weird because the right middle button use no border radius.
At least this is the only visible change for other themes, but it looks same to other windows, eg. filechooser.
Here they make the right button also always visible.
I can also make the button unvisible in boring standard themes like TOK, if you want.
Like Luke mention it here.

raveit65:
For better understanding, themes like BlackMate or TOK or Greenlaguna will have normal buttons after this change.
The only change for those themes is that the right arrow button is always visible for any button state.
Tok themes will lock like in my screenshot.

PR #593, opened 2016-07-14
GTK 3.21: stop properties window going full height

lukefromdc:
GTK 3.21: stop properties window going full height 

Stop the properties window from ballooning to full screen height when built and run with GTK 3.21/22. In GTK 3.21 the "emblems" tab's scrolled window will expand to the entire height of the screen attempting to accomodate all icons unless this is explicitly limited using the new command gtk_scrolled_window_set_max_content_height().

Unfortunately that command is GTK 3.21 and later only, so this forces GTK 3.21 builds to be run only with Gtk 3.21 and later.  Thus this is put behind a GTK 3.21 selector and packages built from it must depend on GTK 3.21 and later. Fortunately no distro is going to ship any file manager built with newer GTK 3 than they ship. A GTK 3.21 build of this will actually open in 3.20 but hit the undefined symbol and crash if you select "properties" on any item, exactly as I would expect it to so.

Caja built with GTK 3.21 won't run with 3.18 already so separating 3.20 and 3.22 builds is not as big a hassle as it might seem.


raveit65:
merged https://github.com/mate-desktop/caja/commit/fb2849550b6916b6d0399a163a756ec3b911d704
Thank you

lukefromdc:
GTK finally fixed this upstream with 
https://git.gnome.org/browse/gtk+/commit/?id=0984d1622d022bf67207f985f7842b6299818e20
but AFTER the release of GTK 3.21.5. When GTK 3.21.6 (or more likely 3.21.91) comes out and others can test, this can now be reverted with no change in behavior (just tested) for one less #ifdef to deal with.
Apparently too many GNOME apps were broken...

raveit65:
Why they can't test more before they push something?

lukefromdc:
My guess is GNOME app devs only use the point releases or even that some of 
the GNOME devs don't try the new versions until they are released. I have entirely
ignored another "dialogs opening top left" issue because I expect that like last time
that will eventually get fixed. If not I will have to fix it myself.

PR #602, opened 2016-07-29
GTK 3.21: Fix desktop redraw, do not change other GTK version builds

lukefromdc:
This should be a 3-way version: No changes here that should affect GTK 2 builds, nor that would affect GTK 3.20 and earlier GTK3 builds. Fix the desktop in GTK 3.21 and later but do not disturb other builds. No impact on distros until they actually use GTK 3.22.

Apply the Nautilus/Nemo based changes for a transparent desktop that works in GTK 3.21 and later only when actually building against GTK 3.21. Tested in GTK 3.21.4 and in GTK 3.20.3 and worked both ways, desktop backgrounds rendered fine. In GTK 3.21 a fallback color is used if Caja is started without a compositor available, normally the desktop is transparent and mate-settings-daemon draws the background with GTK 3.21. In GTK 3.20 or older Caja draws the background as before.

Note that two files are not needed in GTK 3.21 builds but still build their output objects: caja-directory-background.c and caja-directory-background.h , I was unable to create a conditional makefile.am to exclude them. They caused no harm in GTK 3.21 but are dead code could be excluded with some makefile work. Also note that the change to caja.css is unconditional but the added selectors seemed to be ignored without problems in GTK 3.20 test builds.

Passes distcheck in both GTK 3.20 and 3.21, can't test GTK2


raveit65:
Thanks Luke for this work.
I just tested the PR and it does what it promised.
I noticed no regression for other gtk+ versions, so it is save to use it for gtk+-3.21.
And i checked the desktop with gtk3-inspector in compositor case, here the new style class isn't there.
@monsta 
Can you please do a code review?
F25 is branched from rawhide and i'm standing at the front line again.
Without the PR the desktop is unusable in f25, so i really need this for fedora.
Well, i know it isn't perfect and i will try to disable non compositor usage for f25, but switching back to gtk2 for f25 isn't really a solution for me as it will decrease my motivation dramatically for Mate in general, and this will show that we are not able to develop Mate.
Thank you

lukefromdc:
I haven't been able to duplicate session hangs (mentioned in your email) on startup after changing preferences such as icons on the desktop-sessions start fine no matter what I've changed.

Also, I found broken background fading also occurs with master m-s-d, gtk 3.21.4, and no icons on the desktop so the background shows  regardless of caja version. This appears to be broken in mate-desktop, probably in mate_bg_set_surface_as_root_with_crossfade which is found in mate-bg.c

I'll do some digging to see if this works in GNOME (if they still have it) or Cinnamon and it it does what they are doing differently. I looked at that function, looks like another nasty GTK 3.21/cairo issue.

lukefromdc:
I just managed to complete a GTK2 build of my branch combining this branch with the GtkApplication work, no build errors at all came from this code and only one (quickly fixed) with the other. Expected a morass of typos but didn't get it.

src/file-manager/fm-desktop-icon-view.c
```diff
@@ -38,6 +38,9 @@
 #include <gdk/gdkx.h>
 #include <glib/gi18n.h>
 #include <libcaja-private/caja-desktop-icon-file.h>
+#if !GTK_CHECK_VERSION(3, 21, 0)
```
monsta: Looks like this isn't needed, that header is already included a few lines below :smile:

monsta:
Ok, seems good in GTK+2 build, no code is affected.

raveit65:
@lukefromdc 
Should i remove the second include during merge, or is excluding that unneeded?
Btw. for what is this good?
```
.caja-desktop-window,
.caja-desktop:not(:selected):not(:active):not(.rubberband){
	background-color: transparent;
}
```
As it affect all gtk+ versions.

lukefromdc:
You caught a flat-out mistake, one that did not stop any builds but just added dead code to GTK 3.21 builds. I had meant to exclude that line entirely from GTK 3.21 builds where it is never used.


Thus what should be removed is the second, unconditional appearance of 
#include <libcaja-private/caja-directory-background.h>
Just ran sucessful test builds in both GTK 3.20 and 3.21.4 with that change

raveit65:
merged , thank you
I will open a new report based on what i wrote in m-s-d PR.
So we have a reminder that we have some regressions and we can work on it.

PR #607, opened 2016-08-08
Fix background rendering in noncomposited case

lukefromdc:
Apply Alberts Muktupāvels' https://bugzilla.gnome.org/attachment.cgi?id=332444&action=diff to Caja, it just plain works. Tested by switcing between Marco (no compositing) and compiz, both with and without compositing. Background change works fine too.


raveit65:
Ok, i will merge this, remaining issues are minor.

raveit65:
@lukefromdc 
I guess i can remove the new stuff from caja.css?

lukefromdc:
Yes, the portion dealing with the fallback case is no longer relevent and 
should be removed. Somehow I forgot about that.

raveit65:
Rebased and merged https://github.com/mate-desktop/caja/commit/8c926115c99100a581abfd20dced5dfbe4922792

lukefromdc:
Too much of caja.css was removed, breaking the composited case and turning the desktop white. I added back the portion for the desktop and desktop window transparency while leaving out the no longer required portion for the former fallback color:

https://github.com/mate-desktop/caja/pull/609

PR #610, opened 2016-08-10
Fix most crashes on no-composite/no icons bg change 

lukefromdc:
Check for icons on the desktop before running background_changed, do not run it if icons are not being used.  User can now turn icons on or off, then change BG without crash.

There is still one condition that will crash caja but should be less common:  With compositing OFF, user turns icons off, then turns them back on, then changes the background. This will still crash caja, a race condition between the background drawing and reading the gsettings key may be to blame.  

True>false>change BG OK, false>true>change  BG OK, false>true>false>BG change OK, true>false>true>bg change=still crashes.


lukefromdc:
Closing in favor of https://github.com/mate-desktop/caja/pull/611 which fixes the rest of the crashes

PR #611, opened 2016-08-10
Fix background change crash 

lukefromdc:
Apply https://gist.github.com/raveit65/51b40a20f8836562369fc155a828edc4 to add a "finalize" function to the desktop window and stop background-change crashes after turning icons off-or off and back on.

Also match tab styling to that in rest of file, the one difference here from the patchfile.


raveit65:
merged https://github.com/mate-desktop/caja/commit/f5164f75918a8e65ccde9565c4fd9f366064172e

Issue #481, opened 2015-11-09
xfce: When caja is run, it requires 6quits to end the program


raveit65:
origin upstream report:
https://bugzilla.redhat.com/show_bug.cgi?id=1278079
```
 John Duchek 2015-11-04 11:35:17 EST

Description of problem:

If I run caja, it display a directory. I click on the upper right x to close it, and it reopens. If I reclick the x, it closes and reopens.  This occurs for a total of 6 times until the program closes for good.
Version-Release number of selected component (if applicable):
1.10.4

How reproducible:
Every time it runs

Steps to Reproduce:
1. Click on caja
2.click x to close
3.repeat

Actual results:
reopens 6 times

Expected results:
Close the first time I tell it to.


Additional info:
I am running F23 xfce window manager
```
```
 Kevin Fenzi 2015-11-08 11:29:35 EST

I can confirm the behavior, but not sure why it's doing it. ;( 

I do see:

(caja:30907): Eel-WARNING **: "unique eel_ref_str" hash table still has 4 elements at quit time (keys above)
(caja:30907): Eel-WARNING **: "caja-directory.c: directories" hash table still has 2 elements at quit time

Perhaps that means something to the mate folks?
```
```
 Wolfgang Ulbrich 2015-11-08 14:12:54 EST

(In reply to Kevin Fenzi from comment #4)
> I can confirm the behavior, but not sure why it's doing it. ;( 
> 
> I do see:
> 
> (caja:30907): Eel-WARNING **: "unique eel_ref_str" hash table still has 4
> elements at quit time (keys above)
> (caja:30907): Eel-WARNING **: "caja-directory.c: directories" hash table
> still has 2 elements at quit time
> 
> Perhaps that means something to the mate folks?

What you see here is the result of an improvement by this commit.
https://github.com/mate-desktop/caja/commit/4b0d231fdf8dfd6d299ec07cc933000a610f0a74
This avoid that caja hijack the desktop in another enviroment or if you start caja as root.
Normaly caja handles the desktop in mate-session, ie. icons and contextmenus.
This warnings i see in mate-session too if i start caja as root.
So, it's a bit unusual to print such warnings (debuginfos) in a terminal but it's nothing bad and never cause this issue in xfce seesion, imo.
And those are only warnings.
```
```
 Kevin Fenzi 2015-11-08 14:43:43 EST

The logic there seems a bit wrong to me. 

It says:

if you are running as root and not on mate, exit when the window closes. 
[...]

Issue #40, opened 2012-11-08
Bring Caja up to speed


tamer-hassan:
Per the [Roadmap](http://wiki.mate-desktop.org/roadmap):
Done: Start GTK3 support

Per to the mate-file-manager/TODO file:
Done: all the destroy and move things into finalize or dispose
Should be fixed: "Figure out why we get NULL font problems in Pango when the progress dialog comes up."
(changes to progress dialog and ports to cairo-drawing for GTK2 merged in master).

TBD: Migrate libunique to GApplication

I've followed all nautilus changes up until the GtkApplication changes, starting with:
2010-10-30  application: split SmClient code in its own module.
http://git.gnome.org/browse/nautilus/commit/?id=dc0b129436d5c9eb712a80852768387fc2b8cd7c
... http://git.gnome.org/browse/nautilus/log/?ofs=2300

In the future: A bunch of memleak fixes. See Nautilus: [#682603](https://bugzilla.gnome.org/show_bug.cgi?id=682603), [#683769](https://bugzilla.gnome.org/show_bug.cgi?id=683769), and the [list from the forum thread](http://forums.mate-desktop.org/viewtopic.php?f=17&t=821) which Stefano noted in [comment below](https://github.com/mate-desktop/mate-file-manager/issues/40#issuecomment-10446468).

PR #618, opened 2016-08-19
 GTK3: port libunique ->GtkApplication

lukefromdc:
Port GTK3 builds only to GtkApplication, keep GTK2 builds unchanged 

Based on cherrypicked nautilus commits from
https://github.com/GNOME/nautilus/commit/a8481ee4bd8d34e792d63598fa5efb47736f9de4
through
https://github.com/GNOME/nautilus/commit/a8481ee4bd8d34e792d63598fa5efb47736f9de4
and nautilus bigfix
https://github.com/GNOME/nautilus/commit/6ac24b5381c7a26f85b8c72662c9ac4a54269e26

### DIFF
--- a/src/caja-application.c
+++ b/src/caja-application.c
@@ -155,6 +155,7 @@ G_DEFINE_TYPE (CajaApplication, caja_application, GTK_TYPE_APPLICATION);
 struct _CajaApplicationPriv {
 	GVolumeMonitor *volume_monitor;
     gboolean no_desktop;
+    gboolean force_desktop;
     gchar *geometry;
 };
 
@@ -448,7 +449,7 @@ static void
 caja_application_open (GApplication *app,
                GFile **files,
                gint n_files,
-               const gchar *hint)
+               const gchar *options)
 {
     CajaApplication *self = CAJA_APPLICATION (app);
     gboolean browser_window = FALSE;
@@ -458,15 +459,15 @@ caja_application_open (GApplication *app,
     g_debug ("Open called on the GApplication instance; %d files", n_files);
 
     /*Check if local command line passed --browser or --geometry */
-    if (strcmp(hint,"") != 0 ){
+    if (strcmp(options,"") != 0 ){
         if (g_str_match_string ("browser",
-                    hint,
+                    options,
                     FALSE) == TRUE){
             browser_window = TRUE;
-            geometry = strchr(hint, splitter);
+            geometry = strchr(options, splitter);
         }
         else {
-        geometry = hint;
+        geometry = options;
         }
         /*Reset this or 3ed and later invocations will use same
 	     *geometry even if the user has resized open window
@@ -2836,8 +2837,11 @@ caja_application_local_command_line (GApplication *application,
     const gchar *autostart_id;
     gboolean no_default_window = FALSE;
     gchar **remaining = NULL;
-    const gchar *hint = "";
     CajaApplication *self = CAJA_APPLICATION (application);
+
+    /*First set these FALSE */
+    self->priv->force_desktop = FALSE;
+    self->priv->no_desktop = FALSE;
 
     const GOptionEntry options[] = {
 #ifndef CAJA_OMIT_SELF_CHECK
@@ -2852,6 +2856,8 @@ caja_application_local_command_line (GApplication *application,
           N_("Only create windows for explicitly specified URIs."), NULL },
         { "no-desktop", '\0', 0, G_OPTION_ARG_NONE, &self->priv->no_desktop,
           N_("Do not manage the desktop (ignore the preference set in the preferences dialog)."), NULL },
+        { "force-desktop", '\0', 0, G_OPTION_ARG_NONE, &self->priv->force_desktop,
+          N_("manage the desktop regardless of set preferences or environment(on new startup only)"), NULL },
         { "browser", '\0', 0, G_OPTION_ARG_NONE, &browser_window, 
           N_("Open a browser window."), NULL },
         { "quit", 'q', 0, G_OPTION_ARG_NONE, &kill_shell, 
@@ -2972,27 +2978,33 @@ caja_application_local_command_line (GApplication *application,
         files[1] = NULL;
     }
 
-    /*Set up geometry and --browser options for "Open" */
+    /*Set up geometry and --browser options  */
+    /*Invoke "Open" to create new windows */
 
     if (browser_window == TRUE && self->priv->geometry == NULL){
-            hint = "browser";
+
+        if (len > 0)  {
+        g_application_open (application, files, len, "browser");
+        }
     }
 
     else if (browser_window == FALSE && self->priv->geometry != NULL){
-            hint = g_strdup(self->priv->geometry);
+        if (len > 0)  {
+            g_application_open (application, files, len, self->priv->geometry);
+        }
     }
 
     else if (browser_window == TRUE && self->priv->geometry != NULL){
-             hint = g_strconcat("browser","=", self->priv->geometry, NULL);
+        if (len > 0)  {
+            g_application_open (application, files, len, (g_strconcat("browser","=",
+                                self->priv->geometry, NULL)));
+        }
     }
 
     else {
-        hint = ("");
-    }
-
-    /* Invoke "Open" to create new windows */
-    if (len > 0)  {
-        g_application_open (application, files, len, hint);
+        if (len > 0)  {
+            g_application_open (application, files, len, "");
+        }
     }
 
     for (idx = 0; idx < len; idx++) {
@@ -3049,11 +3061,11 @@ init_desktop (CajaApplication *self)
 
     if (running_as_root () || !running_in_mate ())
 	{
-        /* do not manage desktop when running as root or on other desktops */
+        /* do not manage desktop when running as root or on other desktops unless forced */
         self->priv->no_desktop = TRUE;
     }
 
-    if (!self->priv->no_desktop) {
+    if (!self->priv->no_desktop || self->priv->force_desktop) {
         caja_application_open_desktop (self);
     }
 
--- a/src/caja-application.c
+++ b/src/caja-application.c
@@ -2833,7 +2833,6 @@ caja_application_local_command_line (GApplication *application,
     gboolean version = FALSE;
     gboolean browser_window = FALSE;
     gboolean kill_shell = FALSE;
-    gboolean autostart_mode = FALSE;
     const gchar *autostart_id;
     gboolean no_default_window = FALSE;
     gchar **remaining = NULL;
@@ -2885,7 +2884,7 @@ caja_application_local_command_line (GApplication *application,
 	 */
 	autostart_id = g_getenv ("DESKTOP_AUTOSTART_ID");
 	if (autostart_id != NULL && *autostart_id != '\0') {
-		autostart_mode = TRUE;
+		no_default_window = TRUE;
         }
 
 
--- a/src/caja-application.c
+++ b/src/caja-application.c
@@ -21,7 +21,8 @@
  *  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
  *
  *  Authors: Elliot Lee <sopwith@redhat.com>,
- *           Darin Adler <darin@bentspoon.com>
+ *           Darin Adler <darin@bentspoon.com>,
+ *          Cosimo Cecchi <cosimoc@gnome.org>
  *
  */
 
@@ -38,6 +39,9 @@
 #include "caja-information-panel.h"
 #include "caja-history-sidebar.h"
 #include "caja-places-sidebar.h"
+#if GTK_CHECK_VERSION (3, 0, 0)
+#include "caja-self-check-functions.h"
+#endif
 #include "caja-notes-viewer.h"
 #include "caja-emblem-sidebar.h"
 #include "caja-image-properties-page.h"
@@ -45,7 +49,9 @@
 #include <string.h>
 #include <unistd.h>
 #include "caja-desktop-window.h"
+#if !GTK_CHECK_VERSION (3, 0, 0)
 #include "caja-main.h"
+#endif
 #include "caja-spatial-window.h"
 #include "caja-navigation-window.h"
 #include "caja-window-slot.h"
@@ -67,6 +73,9 @@
 #include <libcaja-private/caja-debug-log.h>
 #include <libcaja-private/caja-file-utilities.h>
 #include <libcaja-private/caja-global-preferences.h>
+#if GTK_CHECK_VERSION (3, 0, 0)
+#include <libcaja-private/caja-lib-self-check-functions.h>
+#endif
 #include <libcaja-private/caja-extensions.h>
 #include <libcaja-private/caja-module.h>
 #include <libcaja-private/caja-desktop-link-monitor.h>
@@ -77,6 +86,11 @@
 #define MATE_DESKTOP_USE_UNSTABLE_API
 #include <libmate-desktop/mate-bg.h>
 
+#if GTK_CHECK_VERSION (3, 0, 0)
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+#else
 enum {
 	COMMAND_0, /* unused: 0 is an invalid command */
 
@@ -84,7 +98,7 @@ enum {
 	COMMAND_STOP_DESKTOP,
 	COMMAND_OPEN_BROWSER,
 };
-
+#endif
 /* Keep window from shrinking down ridiculously small; numbers are somewhat arbitrary */
 #define APPLICATION_WINDOW_MIN_WIDTH	300
 #define APPLICATION_WINDOW_MIN_HEIGHT	100
@@ -94,9 +108,10 @@ enum {
 /* Keeps track of all the desktop windows. */
 static GList *caja_application_desktop_windows;
 
+#if !GTK_CHECK_VERSION (3, 0, 0)
 /* Keeps track of all the caja windows. */
 static GList *caja_application_window_list;
-
+#endif
 /* Keeps track of all the object windows */
 static GList *caja_application_spatial_window_list;
 
@@ -107,7 +122,9 @@ static gboolean save_of_accel_map_requested = FALSE;
 static CajaFreedesktopDBus *fdb_manager = NULL;
 
 static void     desktop_changed_callback          (gpointer                  user_data);
+#if !GTK_CHECK_VERSION (3, 0, 0)
 static void     desktop_location_changed_callback (gpointer                  user_data);
+#endif
 static void     mount_removed_callback            (GVolumeMonitor            *monitor,
         GMount                    *mount,
         CajaApplication       *application);
@@ -125,9 +142,23 @@ static void     volume_removed_callback            (GVolumeMonitor           *mo
         CajaApplication      *application);
 static void     drive_listen_for_eject_button      (GDrive *drive,
         CajaApplication *application);
+#if GTK_CHECK_VERSION (3, 0, 0)
+static void     caja_application_load_session    (CajaApplication *application);
+static char *   caja_application_get_session_data (CajaApplication *self);
+void caja_application_quit (CajaApplication *self);
+#else
 static void     caja_application_load_session     (CajaApplication *application);
 static char *   caja_application_get_session_data (void);
+#endif
+#if GTK_CHECK_VERSION (3, 0, 0)
+G_DEFINE_TYPE (CajaApplication, caja_application, GTK_TYPE_APPLICATION);
+struct _CajaApplicationPriv {
+	GVolumeMonitor *volume_monitor;
+    gboolean no_desktop;
+    gchar *geometry;
+};
 
+#else
 G_DEFINE_TYPE (CajaApplication, caja_application, G_TYPE_OBJECT);
 
 static gboolean
@@ -210,20 +241,23 @@ caja_application_get_window_list (void)
 {
     return caja_application_window_list;
 }
+#endif
 
 GList *
+
 caja_application_get_spatial_window_list (void)
 {
     return caja_application_spatial_window_list;
 }
 
+#if !GTK_CHECK_VERSION (3, 0, 0)
 unsigned int
 caja_application_get_n_windows (void)
 {
     return g_list_length (caja_application_window_list) +
            g_list_length (caja_application_desktop_windows);
 }
-
+#endif
 static void
 startup_volume_mount_cb (GObject *source_object,
                          GAsyncResult *res,
@@ -242,7 +276,11 @@ automount_all_volumes (CajaApplication *application)
     if (g_settings_get_boolean (caja_media_preferences, CAJA_PREFERENCES_MEDIA_AUTOMOUNT))
     {
         /* automount all mountable volumes at start-up */
+#if GTK_CHECK_VERSION (3, 0, 0)
+        volumes = g_volume_monitor_get_volumes (application->priv->volume_monitor);
+#else
         volumes = g_volume_monitor_get_volumes (application->volume_monitor);
+#endif
         for (l = volumes; l != NULL; l = l->next)
         {
             volume = l->data;
@@ -274,9 +312,15 @@ smclient_save_state_cb (EggSMClient   *client,
                         CajaApplication *application)
 {
     char *data;
+#if GTK_CHECK_VERSION (3, 0, 0)
+    data = caja_application_get_session_data (application);
 
+    if (data != NULL)
+#else
     data = caja_application_get_session_data ();
+
     if (data)
+#endif  
     {
         g_key_file_set_string (state_file,
                                "Caja",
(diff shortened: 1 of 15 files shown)


lukefromdc:
Just rebased this for changes in master. Those changes are fine here and with this branch, no issues from that. Also last night updated the handing of GtkApplication's "singleton" to match what Nemo (forked from Nautilus 3.4) does, as that is now well-tested code. That commit isn't identical due to differences in Caja's window system, which accomodates both spatial and navigation windows

lukefromdc:
I just found Nemo bug https://github.com/linuxmint/nemo/issues/569,which is an inability to set --geometry except on first invocation caused by how the Nautilus 3.4/Nemo command line handling works. It also effects this branch since the code responsible is the same.

The issue is obscure with no actiivty since 2014 so must not affect many users, but I will try to get the more complex, newer Nautlus code that fixes it to work. Everything else seems fine now, but a lot of rework if --geometry is to work on a second invocation. Culprit is in how GtkApplication passes information from the local to the main instance.  

A local command line cannot set a global variable in the main instance and can only pass four arguments to it via "open."   Setting "HANDLES_COMMAND_LINE" gets around that, but then requires handling "local options" to get relative paths to work when opening folders. Thus nautilus now splits up their command line handling over this, the code is quite complex. 

Since I now have the Nautilus 3.4/Nemo startup system, if I can fix this for Caja Nemo will also be able to use it.

lukefromdc:
Rebased for today's change in master, fixed using --geometry and --browser together, and squashed all

raveit65:
@lukefromdc 
I'm a bit out, where are the latest changes for testing?
This PR or from your repo?

lukefromdc:
The PR has been kept up to date and contains the latest changes and rebase

lukefromdc:
https://github.com/lukefromdc/caja/tree/dev-GtkApplication3 should
be exactly the same as the PR as it stands now, as all updates and
rebases recently have been to that branch on which the PR is based.

lukefromdc:
Just found that GtkApplication has been in glib since 2.28, it may not be
GTK version specific at all, only GLIB version specific. That means we 
could strip out all the GTK3 build selectors for it and replace them with
!has_libunique selectors configured in the build system assuming it actually
WORKS with GTK2 which I do not know. That would permit a distro that
dumps libunique to build with GTK2 if they want. 

https://github.com/mate-desktop/caja/issues/40

raveit65:
I get now a segfault (only with GtkApplication commit) from this branch.
[rave@mother ~]$ LANG=C
[rave@mother ~]$ caja
Segmentation fault (core dumped)
[rave@mother ~]$ 
Tested with gtk+-3.21.x and 3.20.x

lukefromdc:
Where was it? I wil get rare segfaults on opening a window from ANY
branch of Caja, but have not had this problem specific to my branch 
in a while. 

If I know what is different in your setup I should be able to track it down.
I'll try a GTK 3.20 build and see if that can provoke it.

raveit65:
I simply tested this commit https://github.com/mate-desktop/caja/pull/618/commits/67e8b06d8f96c3629a8481fbd3a03308cd504f3d with gtk+-3.20/21 or f24/f25.
Result is that caja doesn't start with session and from terminal i got posted output.
All was working 2 weeks ago before i was in hospital.
I search if i found the working commit from previous build.
My setup is normal gtk3 build from this repo for fedora.
https://copr.fedorainfracloud.org/coprs/raveit65/Mate-GTK3/
Sure, i'm testing all PRs with my box but that caja failed to start in both installations shows me that it isn't a setup problem.

lukefromdc:
Last commit applied is 
https://github.com/lukefromdc/caja/commit/d1630c3aad15745e67431adef9e52cc527f2673b

which modified string handling for the --geometry option. Possibly some difference in Fedora
might cause asegfault on the earlier code? I wasn't even getting warnings from it here though.

lukefromdc:
GTK 3.20 build with the newest commit worked fine here, so at least at
that point the GTK version is now ruled out

raveit65:
What is the branch in your repo with single libunique commits before you squashed all in one.
Let me test this one.

lukefromdc:
This branch is a bit old and does not support the --geometry option in
a second invocation but otherwise generally worked:

https://github.com/lukefromdc/caja/tree/dev-GtkApplication2

I was doing the squash commits because usually this has been preferred,
but in this case the whole history probably should have been kept.

lukefromdc:
Just found this warning after forcing a rebuild for just caja-main.c:

caja-main.c: In function ‘main’:
caja-main.c:384:19: warning: implicit declaration of function ‘caja_application_new’ [-Wimplicit-function-declaration]
     application = caja_application_new();
                   ^~~~~~~~~~~~~~~~~~~~
caja-main.c:384:17: warning: assignment makes pointer from integer without a cast [-Wint-conversion]
     application = caja_application_new();


I'll fix those and add a new commit to the PR. looks like Nautilus put something at the top of nautilus-main to
cover this.

raveit65:
This patch list works
# drop backgrounds
Patch33:        caja_0033-do-not-show-property-browser-in-menu.patch
# https://github.com/mate-desktop/caja/pull/626
Patch68:        caja_0068-GTK3-stop-warning-spew-from-sidebar-pathbar.patch
# https://github.com/lukefromdc/caja/tree/dev-GtkApplication2
Patch69:        caja_0069-GTK3-port-libunique-GtkApplication-w-correct-command.patch
Patch70:        caja_0070-GTK3-restore-browser-option.patch
Patch71:        caja_0071-only-register-app-when-running-in-MATE-not-as-root.patch
Patch72:        caja_0072-Allow-registering-caja-in-other-sessions.patch
Patch73:        caja_0073-GTK3-fix-typo-imported-from-Nautilus.patch
Patch74:        caja_0074-GTK3-application-make-sure-no-default-window-works-w.patch
Patch75:        caja_0075-GTK3-re-enable-session-saving.patch
Patch76:        caja_0076-GTK3-fix-warnings-follow-browser-preference-fully.patch
Patch77:        caja_0077-GTK3-drop-deprec-g_thread_init-imported-from-Naut.patch
Patch78:        caja_0078-GTK3-fix-desktop-after-sudo-spatial-window-bug.patch

Note, patch33 disables only the menu entry for the property browser.

lukefromdc:
Ok, that tells me to look only in code after that.  I am looking at the 
"don't handle the singleton ourselves" change as possible source of
the build warnings, I will fix that if I can, else revert it.

lukefromdc:
Just pushed two new commits, see if they stop the segfault in Fedora. The first one stops
the build warnings in caja-main.c relating to an incompatable pointer and an implicilty
declared function. The other fixes more incompatable pointer warnings that were hard to
find because there are so many build warnings in master's versions of those file.

https://github.com/lukefromdc/caja/commit/8cde5ebb6cf43ffe014d15e7b7268e1e8b47cb6e
GTK3: fix typo in caja-application.h 

https://github.com/lukefromdc/caja/commit/cda83db791122df7320b10ade5a064f11cf64ca9
GTK3: fix some build warnings

lukefromdc:
https://github.com/lukefromdc/caja/commit/5ef357fa48f3e50849dc890b165db5cbf4eea876

is a separate branch split from the PR that reverts the two "don't handle the singleton ourselves"
commits. I created it in case those were the cause of the segfaults. The last commit to 
dev-GtkApplication3 hopefully will fix those but if not try this branch,

raveit65:
Thanks, the PR is working now with f24/gtk+-3.20 and f25/gtk+-3.21.
```
port to GtkApplicaton, https://github.com/mate-desktop/caja/pull/618
Patch61:        caja_0061-GTK3-port-libunique-GtkApplication.patch
Patch62:        caja_0062-GTK3-add-force-desktop-option.patch
Patch63:        caja_0063-GTK3-fix-typo-in-caja-application.h.patch
Patch64:        caja_0064-GTK3-fix-some-build-warnings.patch
warnings, https://github.com/mate-desktop/caja/pull/626
Patch68:        caja_0068-GTK3-stop-warning-spew-from-sidebar-pathbar.patch
```
I can test it now more on my daily production box.
I suggest not to squash it again, better we do that when all is done.
Did @flexiondotorg gave you acls for caja and mate-panel?

raveit65:
I just tested the --force-desktop option in f25 cinnamon.
Works well if i remove caja from cinnamon blacklist and nemo from autostart.

> The other fixes more incompatable pointer warnings that were hard to find because there are so many build warnings in master's versions of those file.

That's true, for that reason i started to fix deprecations to make the build log readable but there are a lot of left.
........deprecations wasn't a high priority by former devs.

raveit65:
i forgot , i had to use caja --force-desktop -n, otherwise a caja window is open.

lukefromdc:
No reply yet on access, I agree about no further squashing until this is
committed. Squashing commits gets in the way of bisecting if an error
comes up. 

I have been abe to get away with it locally because I have a directory
full of source tarballs of almost everything I have tested, 46 of them 
at the moment just for Caja. In a real worst-case situation I would have
been able to regenerate the branch history from these on a new branch.

I don't know why those errors didn't show themselves with segfaults here,
possibly some difference in memory management, in compiler behavior,
something like that. I've seen this before, when Kdenlive devs committed 
changes that must have worked for them as they kept going but segfaulted
instantly on my box when I tried to render out video.

lukefromdc:
I just pushed one more commit that removes the need for -n in the caja desktop file used at startup and thus removes the need to change that file at all. Instead autostart ID is used to set --no-desktop on autostart as it is with libunique. One less file to change.

lukefromdc:
Just cleaned up some more of those runtime terminal warnings:
https://github.com/mate-desktop/caja/pull/629

raveit65:
'startup using autostart id rather than caja -n' works well, also i don't need to add the -n option in another desktop with --force-desktop anymore.

lukefromdc:
Are we still holding this for CAJA 1.18 given Debian's schedule, merging it in 1.16, or doing something else? If we hold it for 1.18 and Debian drops Libunique I could switch the build system to a build selector to use libunique or Gtkapplication by a configuration option or some such thing, and that could qualify as a point release because it would fix a build failure on Debian while not affecting any distro that keeps libunique. Then we could switch over for good for 1.18.

raveit65:
I would hold it for 1.18 and add a configure option for it. We target to release 1.16 in 2 weeks or so. Than we can test/improve it more. And other guys should test it too.
Btw. You should have access now to caja and m-p directly. Maybe start with a branch there?
I'm also always online at Hipchat mate-dev if my box is powerered on.

raveit65:
For fedora i would like to tart with such a massive change with rawhide (f26) which will be released next year in june.

lukefromdc:
OK, I'll branch mate/desktop/master to a new dev-gtkapplication branch,
then I can get to work on the build system to allow a configuration option.
My guess is that will be needed for a bugfix point release for Debian unless
they change their plans regarding libunique and we can then consider using
GtkApplication for all builds with 3.18. Apparently it comes with glib and not
with GTK so it would work even with GTK2, in fact the first tests by GNOME
were with it but glib was changing too fast for it to get out with 2.32.

GTK2 builds will need to be tested if we plan to support GTK2 at all in 3.18.
I think comparison builds of master with libunique are all I have left on my
box that uses it

lukefromdc:
I just pushed this branch to https://github.com/mate-desktop/caja/tree/dev-GtkApplication3, a new PR can be opened from there when the MATE 1.17 cycle starts. One issue I'm working on now: opening w window from the places menu  or panel trash applet gets a busy cursor and a second "opening" button in mate-panel for maybe 15 seconds or so. Looks like an interaction between how i'm using GtkApplication and what gtk_show_uri does. Letting Caja start a new instance (setting name to NULL in caja_application_new()) will stop that behavior but Nemo and Nautius don't have this problem if set as the preferred file manager.   I will look for that difference so this can be fixed. Other than that behavior has been a close match for Master.

raveit65:
> One issue I'm working on now: opening w window from the places menu or panel trash applet gets a busy cursor and a second "opening" button in mate-panel for maybe 15 seconds or so. Looks like an interaction between how i'm using GtkApplication and what gtk_show_uri does. Letting Caja start a new instance (setting name to NULL in caja_application_new()) will stop that behavior but Nemo and Nautius don't have this problem if set as the preferred file manager.

In my opinion we don't need the busy cursor for  the places menu or panel trash applet, because caja starts direct without a delay or not on normal systems. So the user have a visual information.
And if caja does not start a busy cursor is quite useless.
This behaviour can be disabled in desktop files with 'StartupNotify=false'.
I will change that for fedora in any case.
Does Nemo or Nautilus have 'StartupNotify=true' in desktop files?

raveit65:
I see nemo.desktop use  'StartupNotify=false'.

lukefromdc:
Setting 'StartupNotify=false' in caja-folder-handler.desktop fixed the spinning cursor and extra button in the panel. Is this something that can be done unconditionally? If not, making that a GTK3 only change requires two versions of the file and a makefile change.  Also, I just confirmed that GtkApplication is GTK3-only, meaning no distro could build with GTK2 and dump libunique at the same time.

lukefromdc:
https://github.com/mate-desktop/caja/commit/a733a73e083fc7477d58f752ad9f27663f114029
fixes the busy spinning cursor, the change is in caja-folder-handler.desktop. Should I change the other .desktop files to use StartupNotify=false as well or leave them be? Default is to leave them alone.

raveit65:
As it effects gtk2 we should ask @monsta about his opinion.
+1 from my side for use StartupNotify=false for all desktop files.
A busy cursor looks always like something is wrong.
A least this is independent from GtkApplication port and can be done directly in master.

lukefromdc:
OK, I will hold off on the others until we hear from https://github.com/monsta, if he approves I will change them in master and rebase dev-GtkApplication. If he does not I will restrict this to  GTK3 using two versions of each file and a makefile change, in master or just in this branch as you and him prefer.

monsta:
Let's keep a single .desktop file, at least for this PR. It will be simpler and we can change it if GTK+2 build will have issues.

That said, I see that commit https://github.com/mate-desktop/caja/pull/618/commits/b4a32ea9f7e6c020fd80bc4c89ce6313b7414c62 already affects another part of .desktop file and isn't GTK+3 specific.

monsta:
>  Also, I just confirmed that GtkApplication is GTK3-only, meaning no distro could build with GTK2 and dump libunique at the same time.

That's true. Only GApplication from glib can be used independently of GTK+ (it's a base class of GtkApplication). But it doesn't have some useful stuff in its API, so it's probably suitable for simpler cases than this one (see https://github.com/mate-desktop/mate-netbook/pull/33 for an example).

lukefromdc:
https://github.com/mate-desktop/caja/pull/618/commits/b4a32ea9f7e6c020fd80bc4c89ce6313b7414c62
was a revert of a prior change to startup .desktop file formerly needed to keep a navigation window from opening at startup. When I found out how to use the autostart ID to do this I was able to simplify things a bit

lukefromdc:
I do not know how to set up the build system to replace the GTK3 check with a --with-libunique option for this so distros can choose. My first attempt went nowhere. Can anyone else give me some pointers on how to go about this? Changing the source files isn't the problem, it's getting configure.ac to recognize a --with-libunique option that has me stumped. On top of all else I need a way to make that option only available with GTK3 is selected, and ./configure --help will need to explain what it is.

monsta:
Won't it be more complicated than checking for libunique in GTK+2 build only? Like it was done in https://github.com/mate-desktop/mate-user-share/pull/40 and https://github.com/mate-desktop/mate-power-manager/pull/199.

raveit65:
> Won't it be more complicated than checking for libunique in GTK+2 build only?

As only debian maintainer will drop libunique w/o any reason, i do not see why i shouldn't use libunique3 for fedora/rhel.
Again, rhel7 will reach his eol in June 30, 2024, i'm really in doubt that libunique upstream will end before this date.
So i prefer  a solution for both toolkit versions, btw. i general if caja will be gtk3 only.

lukefromdc:
If there is to be the option of building with libunique and GTK3, than it  needs
to be checked for in GTK3 unless explicitly disabled or not detected. I don't know
much about the make system and getting new options into configure.ac and the
makefiles. that is what is getting in my way here.

Goal here is to let distros build with GTK2 and libunique, GTK3 and libunique,
or GTK3 and GtkApplication.  Debian can't bind any other distro nor should they
be able to-but they could drop MATE. Probably in that case Ubuntu MATE would
pick it back up-including libunique if necessary-but GTK version messes would make
the Ubuntu versions useless in Debian until Ubuntu caught up with GTK 3.22 in
April 2017. 

Also a build option is the only way a point release for Debian if they do proceed
with dumping libunique could be released, since making it a nondefault build 
option does not break it for any other distro and is a fix for a build failure in a
sense.  It is possible they underestimated the difficulty of removing libunique
dependency from every app. As it stands, for Debian Stable they must either
use a point release of Caja containing this, cherrypick a branch, keep libunique,
or dump Caja and presumably the rest of MATE. My suspicion is someone went
looking for a list of old little-used packages to get rid of.

BTW, In Debian, libunique has had no updates since 2013 and the upstream 
NEWS file last changed in 2011. In other words it is getting less attention than
GTK2 itself.  It is still in Debian Unstable and in testing at the moment, but 
Debian does not plan to release until January at the earliest.

Some help from Debian devs on this would be nice...

lukefromdc:
https://github.com/mate-desktop/caja/commit/7fb718d51bb85ec2f80aad4bbfa0e69db22aa2d2 permits buidling for GTK3 with libunique by default or with --disable-libunique to force use of GtkApplication. One FIXME: when libunique is not found in a default configuration for GTK3. the error message should suggest use of --disable-libunique, haven't figured out how to change that yet but the rest is here

sc0w:
about https://github.com/mate-desktop/caja/commit/7fb718d51bb85ec2f80aad4bbfa0e69db22aa2d2

Why we need gtk3 with the deprecated libunique ?

lukefromdc:
Because some distros want to keep it (Fedora et all)-and because this big a
change could not be included in a default build of a point release. Debian
Stretch comes out well before MATE 1.18 it looks like, so merging this into
a point release would allow Debian to build Caja with no effect on other 
distros until MATE 1.18-or at all, if they prefer.

monsta:
This seemingly small issue became so complicated that I'm already thinking about making Caja GTK+3 only and make 1.18 release early next year so that it would make it into next Debian Stable. :confused:

lukefromdc:
That would incredibly simplify  a total of 13 files in src, making them a
lot easier to read.

lukefromdc:
I suspect that an early 1.18 release might be the only practical solution to 
this Debian mess. Their timing was already delayed to accomodate a kernel
release, otherwise this would be even worse.

lukefromdc:
Debian STILL has not picked up MATE 1.16 and is still distributing the broken 1.14.2 with GTK 3.22.
https://packages.debian.org/stretch/caja
I wonder if that's because of this libunique issue or something else?

lukefromdc:
https://github.com/mate-desktop/mate-media/pull/81 for mate-media seems to have been forgotten, that's another libunique issue and that leaves mate-control-center.

lukefromdc:
https://github.com/mate-desktop/mate-control-center/pull/253 along with the abovementioned 
https://github.com/mate-desktop/mate-media/pull/81 (by sc0w)
plus this PR all used together on my system allow the full session without libunique installed. Haven't checked mate-screensaver or mate-netbook as I don't have them installed, I'll have to check every remaining configure.ac in MATE for any other instances of libunique

monsta:
>  I wonder if that's because of this libunique issue or something else?

I think that issue shouldn't hold the uploads - the lib isn't removed from repos yet.

lukefromdc:
I didn't get the warning in Master-but didn't get it when closing by closing
the window on my system wither after putting that one line behind GTK2
only selectors.

lukefromdc:
I did not find any sign of libunique in the configure.ac files of mate-indicator-applet, mate-netbook, caja-dropbox python-caja, or mate-screensaver, so along with the majority of MATE 1.16 that I have installed I think we now have PR's for all remaining libunique-using packages.

raveit65:
@monsta 
Can we review and merge this?
As far i can see all this changes are GTK+-3 only.
This PR blocks any other and i don't want to force Luke to rebase this one.
We have a very important PR abut fixing accessible object which is more urgent than debian drop of libunique.
So no need to wait longer for merging this one.

monsta:
I'll probably start moving libmate-desktop to GTK+3 in order for future moving Caja and other packages as well. That should make the changes proposed here a bit smaller, and we will get rid of libunique completely as a side effect.

raveit65:
> I'll probably start moving libmate-desktop to GTK+3 in order for future moving Caja and other packages as well. That should make the changes proposed here a bit smaller, and we will get rid of libunique completely as a side effect.

I know your plans, but does that make sense?
Making first caja gtk3 only will break every open PR, and we can can wait another 2 month for merging this PR and developing stops.
I suggest to merge this before any other action.

monsta:
Agree. Let's merge it then. We're on the road to 1.18 anyway, so we'll have some time to find & fix any possible regressions.

lukefromdc:
Squashed and merged, commit messages cleaned up for changelogs:
https://github.com/mate-desktop/caja/commit/9e5ea15d104720cfee752c193b77f8b03558c6b9

No effect unless --disable-libunique selected at configuration. ./configure --help warns that is for GTK3 only. One commit only, can be reverted if anything ugly comes up.

https://github.com/mate-desktop/caja/pull/652 to fix accessability comes up clean, as still being mergable without manual intervention. There are no common files, so it should also actually work in GtkApplication builds.

raveit65:
@lukefromdc @monsta 
We have a regression with caja from master :-/
https://github.com/mate-desktop/caja/issues/481 is back.
caja needs 5 times again to close in xfce session, but only with starting caja as normal user.
Starting caja as root caja works fine.
I recall there is a check for session-managers here in PR, but i don't find it.
That should be disabled for xfce.
Not sure what happens in other sessions like cinnamon or gnome.
Also, one time caja did hijack xfce desktop in my tests.
But this i couldn't reproduce any more.
1.16 tarball doen't have this issue.

I really suggest not to merge more stuff until this is fixed and we are sure we don't have more regressions.

raveit65:
> Also, one time caja did hijack xfce desktop in my tests.

I can reproduce this now.

raveit65:
> Also, one time caja did hijack xfce desktop in my tests.

> I can reproduce this with using https://github.com/mate-desktop/caja/pull/653 on top of master.

653 is unreleated to this issue. I can reproduce it also with master.
1 in a terminal become root with su
2 start caja as root
3 close caja
4 exit root user in terminal
5 start caja in terminal as normal user
After closing caja several times, caja stays open and control xfce desktop
Here i get also the stupid well known dconf/systemd-pam warning.
```
(caja:1962): dconf-CRITICAL **: unable to create file '/run/user/1000/dconf/user': Permission denied.  dconf will not work properly.
```

raveit65:
opps, starting caja as root (not sudo) in xfce session creates /run/user/1000/dconf/user with such permissions:
```
[root@f24-gtk3 dconf]# ls -l
insgesamt 4
-rw------- 1 root root 2  9. Okt 12:50 user
```
This is bad.

raveit65:
Ok, this  happens if the file 'user' doesn't exits, like in xfce session.
If i delete the file in Mate session and start caja as root, then this file will created with root permissions.

monsta:
Use ```su -l``` or ```sudo``` instead. Using just ```su``` causes the same old ```XDG_RUNTIME_DIR``` collision (see the part of discussion starting at https://bugs.debian.org/732209#247).

monsta:
So looks like the logic from https://github.com/mate-desktop/caja/pull/489 was lost in the process.

monsta:
Side note: I also missed some duplicated code. Reading of and connecting to ```MATE_BG_KEY_SHOW_DESKTOP``` key was already present in the code before this PR.

raveit65:
> Use su -l or sudo instead. Using just su causes the same old XDG_RUNTIME_DIR collision (see the part of discussion starting at https://bugs.debian.org/732209#247).

It's not the point what i use :) 
I use consolehelper + a pam config for starting apps as root.
The question is what users out side of Mate desktop do, eg. in xfce which do not use dconf normal.
But i'm not sure if this is related to the PR.

The real problem is that caja restarts now in xfce.
As i said,  some checks for other SMs are added
```
[rave@f25 ~]$ sudo caja

(caja:1426): Gtk-WARNING **: Failed to get the GNOME session proxy: The name org.gnome.SessionManager is not owned

(caja:1426): Gtk-WARNING **: Failed to get the Xfce session proxy: The name org.xfce.SessionManager is not owned

(caja:1426): Gtk-WARNING **: Failed to get an inhibit portal proxy: The name org.freedesktop.portal.Desktop is not owned
```
This should be removed for xfce, imo.

lukefromdc:
Closing the PR as it has been squashed and merged

lukefromdc:
These three warnings also occur in both Nemo and Nautilus:

```
(caja:1426): Gtk-WARNING **: Failed to get the GNOME session proxy: The name org.gnome.SessionManager is not owned

(caja:1426): Gtk-WARNING **: Failed to get the Xfce session proxy: The name org.xfce.SessionManager is not owned

(caja:1426): Gtk-WARNING **: Failed to get an inhibit portal proxy: The name org.freedesktop.portal.Desktop is not owned
```

If I can track them down and fix them I should probably submit PR's for Nemo and Nautilus too.

lukefromdc:
I tried to use the "no restart" mode from https://github.com/mate-desktop/caja/pull/489 but  I never got it to work in GtkApplication. As a result I had to use a more general fix for the restart loop that resulted from combining the no_desktop and exit_with_last_window options, having Caja hold open if autostarted. Otherwise any session of MATE started as root got the restart loop. With the fix I used, in GtkApplication builds user sessions are now protected from the restart loop as well if a user turns off icons on the desktop and sets exit_with_last_window.

If I can get the no restart mode to work right, I could set it to always be used with exit_with_last_window, and then instead of being forcibly held open to stop the restart loop, caja in a root session would start and immediately exit, permitting the user to restart it with --force-desktop if desired. In a user session the same behavior would be available with no_desktop and exit_with_last_window.

To see the restart loop in action, build caja 1.16.0, run it and open dconf-editor. Turn off icons on the desktop, then set exit_with_last_window. Now restart caja with caja --quit or killall caja. Caja now restarts over and over while the panel fills up with more and more "starting caja" notification buttons. This cannot be stopped except by either putting icons back on the desktop or setting caja not to close with the last window, because https://github.com/mate-desktop/caja/pull/489 didn't cover that case.   The startup_notify-false change I made to the .desktop files will keep the panel from filling up with notification buttons in a default (without --disable-libunique) build of Caja 1.17 but in the background it will still be restarting endlessly with these settings, unless a change is made in the code for libunique builds as well to deal with that case.

lukefromdc:
I set up configure.ac to enable libnunique by default unless --disable-libunique
is passed, and we need to add a message if libunique is not found suggesting
building with --disable-libunique. Not sure how to do that.

To test the xfce case I will have to install xfce, not sure if I have enough bandwidth to pull the packages at home right now.

lukefromdc:
https://github.com/mate-desktop/caja/pull/654 should fix the XFCE case but I don't have XFCE installed for a direct test.A root session test worked fine. The "no restart" logic did not work because the sm client mode can only be set before the main loop starts, and in GtkApplication most of that sort of logic moves out of the main loop. Thus the need to turn off session management in root/not in mate sessions. It defaults to restart and by the time the program can change that it's too late to reset it with GtkApplication it seems.

lukefromdc:
BTW, the  /run/user/1000/dconf/user problem is one I've had from time to time for years, mostly on running Caja or Nemo as su root. Not recently though. Can anyone point to a commit that stopped this by or before the 1.16 release?

raveit65:
I don't think that this was never not a problem with caja since the port to dconf/gsettings with 1.5.
With 1.5/1.6 we had huge probs with spanning caja windows during session start, and wrong permissions for /run/user/1000/dconf/user was one cause of it.
But that caja itself can write a file with wrong permissions as root is new for me.
I think caja could do this all the years.
Imo, this is more a systemd/pam problem, but sadly the main systemd developer ignore this since years, see https://bugzilla.redhat.com/show_bug.cgi?id=753882
.......a very long report.
As you see from time to time i get reports with this problem and Mate applications and i always re-asign it to this report.
.....but it is like sending reports to a black hole ;)

monsta:
> Can anyone point to a commit that stopped this by or before the 1.16 release?

You can't fix it with some commit... :smile:
It's libpam-systemd and su not getting along with each other. See my comment above (and link to discussion) at https://github.com/mate-desktop/caja/pull/618#issuecomment-252486720.

src/caja-application.c
```diff
@@ -745,7 +995,65 @@ do_initialize_screensaver (CajaApplication *application)
...
+    }
+    g_free (metafile_dir);
+
+    caja_dir = g_build_filename (g_get_home_dir (),
+                     ".caja", NULL);
+    xdg_dir = caja_get_user_directory ();
+    if (g_file_test (caja_dir, G_FILE_TEST_IS_DIR)) {
+        /* test if we already attempted to migrate first */
+        updated = g_build_filename (caja_dir, "DEPRECATED-DIRECTORY", NULL);
+        message = _("Caja 3.0 deprecated this directory and tried migrating "
```
monsta: Hmm... "Caja 3.0"?

lukefromdc:
commit https://github.com/mate-desktop/caja/commit/481d6eeee76b95faec6053280739c6bed1dd23b3 fixes that "Caja 3.0" message created when all instances of Nautilus in GtkApplication commits applies to Caja were changed to Caja references. Thus, "Nautilus 3.0" became "Caja 3.0" and I didn't catch it.. Changed it to read "Caja newer than 1.16 deprecated this directory and tried migrating"

monsta:
Maybe we don't even need this code? I'm not sure we ever have ```~/.caja``` directory supported in some release.

Nautilus needed it for migration from 2.xx stuff:
https://git.gnome.org/browse/nautilus/commit/?id=60b27c11e0dd3416eee13bdea45b280ec2562a36
https://git.gnome.org/browse/nautilus/commit/?id=521e3c8481fcf2af398d2ed62692d415448386cd

Some time ago they removed all migration stuff:
https://git.gnome.org/browse/nautilus/commit/?id=87d24047f674ca941e2691602da2501945a82330

lukefromdc:
OK, I can confirm that no ~/.caja directory exists on my system, so probably all the migration code is dead code and can go.

lukefromdc:
New PR https://github.com/mate-desktop/caja/pull/678  removes the unused migration code from both libunique and GtkApplication builds, not sure how it ever got to the libunique code in the first place unless it went into Nautlus prior to their own switch to GTK3, perhaps just after the 2.32 release/

monsta:
Seems so. Look at the dates of first two commits I've posted :smile:

raveit65:
@lukefromdc 
Do you agree with dropping libunigue support for gtk3 ?
@monsta and me are fine with it.
After using GtkApplication since 2 month it's Ok for me.

monsta:
@lukefromdc: also this unresolved question about keeping or dropping libunique had been holding me from making a migration branch...

lukefromdc:
I favor dropping libunique at this point: it's a lot more code to maintain, requires testing builds both ways after other changes, and we've already got one GtkApplication-only feature, namely the --force-desktop option so useful when running in other sessions. 

If someone has an actual need for libunique that would be the only reason to keep it, understanding that keeping such a large dual codebase makes more work going forward.

monsta:
That's good to hear. Can you remove it now in GTK+3 build? Or should I make GTK+3 migration branch and remove it there?

lukefromdc:
I will take it out today

lukefromdc:
Just did some experimentation, best way is probably to do this in dev-gtk3-migration as all of the libunique code is required to support GTK2

monsta:
Ok, I'll try to do it tomorrow then. I was asking because the code is quite complicated...

lukefromdc:
It is, I just sucessfully  built  a version with all libunique selectors converted back to gtk3 selectors, I will start pushing that code as the prototype of GTK3-migration and probably can finish it myself. GTK3 build is a success, I will ignore GTK2 for now on that but push it as the first commit to dev-gtk3-migration, then use those selectors to chop out both libunique and gtk2-specific code in the next commit

monsta:
Ok good, I'll wait for its completion then.

BTW, we discussed migration branches with @raveit65 on IRC, he suggested to start merging them into master ones. Previously I was going to create PRs for them, but now I really doubt whether someone who uses GTK+2 build daily would test them.

lukefromdc:
dev-gtk3-migration is ready. The configure.ac work might not be the neatest approach but it works. If someone wants to do something more elegant with it they are more than welcome to

monsta:
Ok, I will try to find some time to test it (was busy with various stuff yesterday)...

lukefromdc:
Here are some anwsers to questions raised in commit comments that can be difficult to find because of the length of that page. They appeared under indivdual changed files, of which there are 105 listed.

1: A GTK_VERSION_CHECK inside a negation of it was found in the removal of the GTK3 selectors. This was almost certainly a result of using Pluma to replace all libunique selectors with GTK version selectors in the first commit. That gave very screwy results but somehow still built. That should all be gone with the next commit, the only remaining GTK_VERSION_CHECK instances should be those for specific GTK 3 versions.  Those for 3.14 probably can go too, since previous discussion has indicated that only 3.14 and newer are to be supported and I've got configure.ac set up to require 3.14 or later

2: init_css was removed from caja-application.c because that function is in the removed libunique block, in the GtkApplication block the equivalalent function is called init_icons_and_styles as there is more in it

3: The caja-main.h file is not used in GtkApplication builds because the original Nautilus commits for GtkApplication removed nautilus-main.h and most of the porting was an application of these commits to Caja with only the changes needed
to accomodate changes in Caja since it was forked off.

raveit65:
i will move that to another report :-)

Issue #492, opened 2015-12-30
Switching/closing tab doesn't change the icon on taskbar


zaps166:
Let's have two tabs:
1. /home/user (folder icon)
2. /home/user/Desktop (desktop icon) - current

Then switch to tab 1 and notice that the icon is not changed (still has desktop icon).
##

monsta:
@zaps166: fixed in https://github.com/mate-desktop/caja/commit/887c8e8608f51f59bf520ba691e9eb9077d8e92d, but there's a related issue which I described here: https://github.com/mate-desktop/caja/pull/619#issuecomment-247020227.

PR #619, opened 2016-08-19
Fix for #492

bl0ckeduser:
Let me know if this works for you


monsta:
Seems to work, but I also noticed a related thing... When I right-click on some folder with its own icon (Desktop, Music, etc.) and select "Open in new tab", then the taskbar icon immediately changes to that folder's icon. Yes, even before I switch to that other tab. This happens with or without this PR, I just wanted to point it out.

raveit65:
https://github.com/mate-desktop/caja/commit/887c8e8608f51f59bf520ba691e9eb9077d8e92d
Thanks

Issue #561, opened 2016-06-03
GTK 3.21 Caja runtime warnings


lukefromdc:
These are the current runtime warnings with GTK 3.21 and  Caja, Exact values for underallocations (both required and requested) change with the theme, this is with my UbuntuStudio_Legacy theme. Not a theme issue, Adwaita does this too with larger numbers:

(caja:14357): Gtk-WARNING **: Theme parsing error: <data>:1:51: The style property GtkStatusbar:shadow-type is deprecated and shouldn't be used anymore. It will be removed in a future version

(caja:14357): Gtk-WARNING **: Allocating size to GtkButton 0x21e8500 without calling gtk_widget_get_preferred_width/height(). How does the code know the size to allocate?

(caja:14357): Gtk-WARNING **: gtk_widget_size_allocate(): attempt to underallocate CajaPathBar's child GtkButton 0x21e8500. Allocation is 19x22, but minimum required size is 25x21.

(caja:14357): Gtk-WARNING **: gtk_widget_size_allocate(): attempt to underallocate GtkButton's child GtkArrow 0x2225120. Allocation is 9x16, but minimum required size is 15x15.

(caja:14357): Gtk-WARNING **: gtk_widget_size_allocate(): attempt to underallocate CajaPathBar's child GtkButton 0x21e8500. Allocation is 19x22, but minimum required size is 25x21.

(caja:14357): Gtk-WARNING **: gtk_widget_size_allocate(): attempt to underallocate CajaPathBar's child GtkButton 0x21e8500. Allocation is 19x22, but minimum required size is 25x21.

(caja:14357): Gtk-WARNING **: Allocating size to GtkTable 0x1eec700 without calling gtk_widget_get_preferred_width/height(). How does the code know the size to allocate?

(caja:14357): Gtk-WARNING **: gtk_widget_size_allocate(): attempt to underallocate CajaPathBar's child GtkButton 0x21e8500. Allocation is 19x22, but minimum required size is 25x21.
##

monsta:
Ok... how's it with latest 3.22.x?

lukefromdc:
On running "sudo caja" to force a new instance as root, I get only these three warnngs, which both  Nemo and Nautilus also get:

```
(caja:20474): Gtk-WARNING **: Failed to get the GNOME session proxy: The name org.gnome.SessionManager is not owned

(caja:20474): Gtk-WARNING **: Failed to get the Xfce session proxy: The name org.xfce.SessionManager is not owned

[...]

PR #626, opened 2016-09-01
GTK3: stop warning spew from sidebar, pathbar

lukefromdc:
The button underallocation warnings came from pathbar sliders picking up padding from themes, then being forced to render to a smaller size in the code.

The GTK_ICON_LOOKUP_GENERIC_FALLBACK warnings came from the sidebar icons, this flag is now a source of errors. Unfortunately NULL does not work, so set GTK_ICON_LOOKUP_FORCE_SIZE, which does not create errors and ensures the icon size we actually want in the sidebar is used anyway.

Stops all but two of the warnings listed in https://github.com/mate-desktop/caja/issues/561

### DIFF
--- a/data/caja.css
+++ b/data/caja.css
@@ -62,3 +62,8 @@
    border-width: 0px;
    border-style: none;
 }
+
+/* Padding in slider buttons causes GTK errors in GTK 3.20 or later */
+.caja-navigation-window .slider-button {
+	padding: 0px;
+}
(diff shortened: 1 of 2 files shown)


raveit65:
Thank you.

lukefromdc:
I needed all that spam out of my way to make debugging dev-GtkApplication
easier.

PR #629, opened 2016-09-06
GTK3: fix pathbar pango and "preferred width/height"warnings

lukefromdc:
Fix gtk_widget_create_pango_layout: assertion 'GTK_IS_WIDGET (widget)' failed warning and those descending from it. This does NOT fix the  gtk_widget_set_size_request: assertion 'GTK_IS_WIDGET (widget)' failed, which occurs because recent GTK3 versions warn that button_data is not a GtkWidget but still succeed in setting the size. Since GTK 3.22 will be the LAST version of GTK3 this won't break until GTK4 if ever. Taking this out caused the pathbar buttons to jump on bold text so it had to be retained.

Also fix those "allocating size of GtkButton <xxxx> without calling gtk_widget get preferred width/height()" warnings from caja-pathbar's slider buttons

### DIFF
--- a/src/caja-pathbar.c
+++ b/src/caja-pathbar.c
@@ -560,6 +560,7 @@ caja_path_bar_get_preferred_width (GtkWidget *widget,
     gint child_height;
     gint height;
     gint child_min, child_nat;
+    gint slider_width;
 
     path_bar = CAJA_PATH_BAR (widget);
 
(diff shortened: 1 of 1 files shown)


raveit65:
does what it promised, thank you.

lukefromdc:
https://github.com/mate-desktop/caja/pull/631
Cleans up that last warning from the pathbar.

PR #631, opened 2016-09-06
GTK3: Fix last pathbar warning

lukefromdc:
Replace size request code with padding code borrowed from Nemo. For some reason gtk_widget_set_margin_start(or end) does not throw GTK warnings when called on GTK_WIDGET(button_date->label) while  gtk_widget_set_size_request does throw warnings. Neither generates a build warning.


lukefromdc:
Also added a fix for a statusbar deprecation warning in caja-window.c. I did a test merge of dev-GtkApplication3 into a local dummy branch based on this and had no problems, so this should not interfere with any later merge of dev-gtkApplication3.

PR #633, opened 2016-09-09
Add swap icons when moving them too close

x1596357:
When you move icons near existent icons at a very close distance, caja allow that behavior for now. However this kind of behavior can make desktop icons very messy because one can stack on another. This PR prevents icon moving when they're too close by switching their positions.

### DIFF
--- a/libcaja-private/caja-icon-container.c
+++ b/libcaja-private/caja-icon-container.c
@@ -50,6 +50,7 @@
 #include <gdk/gdkx.h>
 #include <glib/gi18n.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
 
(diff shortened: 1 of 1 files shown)


lukefromdc:
I tried to test this by dropping the single altered file into a build directory for my dev-GtkApplication3 branch but got a build error:
```
caja-icon-container.c: In function ‘is_icon_nearer’:
caja-icon-container.c:2657:19: error: expected identifier before ‘=’ token
     icon_position.= caja_icon_canvas_item_get_icon_rectangle(((CajaIcon *)icon)->item);
```
Pretty sure this is a problem with the PR, as this file does not change from master in my branch(nothing in libcaja-private does) and I downloaded the entire changed file and dropped it in. Also the error itself looks like a typo got into the file when it was committed.

PR #635, opened 2016-09-11
Add swap desktop icons when they move too close

zzhiyi:
When you move icons on desktop near existent icons at a very close distance, caja allow that behavior for now. However this kind of behavior can make desktop icons very messy because one can stack on another. This PR prevents icon moving when they're too close by switching their positions.
@lukefromdc I built it again. The typo was my bad. Made a stupid mistake. Sorry. And I want to create a new PR as I don't want to add another commit that says "fix typo" :)

### DIFF
--- a/libcaja-private/caja-icon-container.c
+++ b/libcaja-private/caja-icon-container.c
@@ -50,6 +50,7 @@
 #include <gdk/gdkx.h>
 #include <glib/gi18n.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
 
 #if !GTK_CHECK_VERSION(3, 0, 0)
@@ -247,6 +248,16 @@ static int compare_icons_vertical (CajaIconContainer *container,
                                    CajaIcon *icon_b);
 
 static void store_layout_timestamps_now (CajaIconContainer *container);
+
+typedef gboolean (*IsBetterIconFunction)(CajaIconContainer *container,
+                                         CajaIcon *start_icon,
+                                         CajaIcon *best_so_far,
+                                         CajaIcon *candidate, void *data);
+
+static CajaIcon *   find_best_icon (CajaIconContainer *container,
+                                    CajaIcon *start_icon,
+                                    IsBetterIconFunction function,
+                                    void *data);
 
 static gpointer accessible_parent_class;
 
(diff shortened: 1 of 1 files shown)


lukefromdc:
Got this error on attempting to build against caja master as of 10-14-2016 10PM EDT:

`caja-icon-container.c:6376: undefined reference to `eel_accessibility_set_atk_object_return'`

probably needs rework due to commit https://github.com/mate-desktop/caja/commit/99a9c6e7c6a67484c029518bdb31160469627701 which was needed to get Orca to work

lukefromdc:
I just got this to work, turned out my prior file-based test was invalid. I rebased and merged to a branch on my own repo https://github.com/lukefromdc/caja/tree/dev-swapicons, it built fine and is running fine right now.

lukefromdc:
Got it to work, created a rebased branch at https://github.com/mate-desktop/caja/tree/dev-swapicons2 but found that at least in GTK 3.22 the icons don't switch positions, they just act the same as before. You can get around some rebasing nastiness by merging my branch back to a new branch in your own repo and working on it there if you want.

x1596357:
I think I will look into it when I have more free time. You're free to use merge the rebased branch and close this MR.

raveit65:
> You're free to merge the rebased branch and close this MR.

We can't merge something which do nothing :)

lukefromdc:
Should I close this?

lukefromdc:
I just found this does in fact work-but only if the icons are dropped almost exactly on top of oneanother. They can still overlap, they need to be within about 20px or so of a perfect match. This is enough to prevent an exact stack of the icons but that's about it unless the minimum distance is opened up to perhaps 48px, the typical icon size on the desktop, maybe a bit more

lukefromdc:
OK, the issues here seem to be problems with how "keep aligned" works. When "keep aligned" is selected, icons are held to gridded positions horizontally but not vertically. Thus, if and only if keep aligned is used on the desktop, this will swap them if dropped into the same column from separate columns-if they are at almost exactly the same height.  If the heights differ by more than about 10px, they will still end up overlapping, presumably same as before.

This is because if you move an icon vertically onto another icon, staying in the same column, the lack of a vertical version of keep aligned means unless they are dropped within that 10px or so of a perfect center to center match, the swap function is not triggered. Increasing that center to center distance could fix this

lukefromdc:
Just got this to work in the vertical plane too with 
https://github.com/mate-desktop/caja/commit/d3810cedb02b4660c2d1c771ae682868b6df32d4 to 
https://github.com/mate-desktop/caja/tree/dev-swapicons2

opening a new PR for this because it is works, is rebased on today's master and is ready to go

lukefromdc:
https://github.com/mate-desktop/caja/pull/663
is this PR with the hardcoded fix for vertical icon swapping

x1596357:
Re: EDIT: also requires "keep aligned" to be enabled, I assume that is intentional?

Yes. It's intentional. I want that users could do anything they want such as overlapping icons when keep aligned is disabled.

lukefromdc:
Closing as there is no simple way to save the results of this so it does not get reverted on the next start of caja. Even a simple use of "keep aligned" requires selecting the entire desktop, moving all the icons together, and then moving them back to write the filenames and positions to the file where they are saved. The functions modified in these PR's have no access to the necessary information, so no easy way to fix this.

PR #647, opened 2016-10-05
GTK3: fix warnings cause by GtkGrid

lukefromdc:
Apply https://github.com/itzexor/nemo/commit/2c60a3edaef10e9ef331740d55d3c45dd366c5db
window: remove custom get_preferred_width/height implementation
We already take care of the default size when we create the window, and these only break assumptions of the default GtkWindow's handlers.

in turn from
https://github.com/GNOME/nautilus/commit/da06dfc438e578405d616c48e8e3809d37c908bb

Nautilus commit is from Feb 17, 2014 prior to release of GTK 3.14. Could not build against GTK 3.14 over GTK 3.22 system due to mate-desktop errors but should work as Nautilus used it with 3.14. Test builds with GTK 3.16 and 3.22 worked fine, no issues with saved or command line geometry

### DIFF
--- a/src/caja-window.c
+++ b/src/caja-window.c
@@ -1050,51 +1050,6 @@ caja_window_slot_close (CajaWindowSlot *slot)
     caja_window_pane_slot_close (slot->pane, slot);
 }
 
-#if GTK_CHECK_VERSION(3,0,0)
-static void
-caja_window_get_preferred_width (GtkWidget *widget,
-    			     gint *minimal_width,
-    			     gint *natural_width)
-{
-    GdkScreen *screen;
-    gint max_w, min_w, min_h, default_w, default_h;
-    CajaWindow *window = CAJA_WINDOW (widget);
-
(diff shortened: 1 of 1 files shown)


lukefromdc:
Sucessful build with GTK2, though I could not run that over a GTK3 system due to GTK3 libraries. Enough to show no errors in the GTK2/3 build selectors, so no change in GTK2 and sucessful tests with multiple GTK3 versions. Merging this, if others find problems it can be reverted.

raveit65:
Thanks, the warning is gone.
....sorry for late testing.
I see no regressions with a quick test.

lukefromdc:
OK, merging this. BTY, I did a quick test on a throwaway local branch: this
and dev-GtkApplication 3 can merge in either order, so no worries about breaking
the dev-GtkApplication 3 PR. In fact, a merge of these two is what I am running 
locally right now.

Also checked the crucial accessabiity PR, saw no common files between them.

PR #653, opened 2016-10-08
Fix hashtable warnings on close

lukefromdc:
As in Nautilus, use g_hash_table_new rather than eel_g_hash_table_new_free_at_exit

### DIFF
--- a/eel/eel-string.c
+++ b/eel/eel-string.c
@@ -792,8 +792,7 @@ eel_ref_str_get_unique (const char *string)
     if (unique_ref_strs == NULL)
     {
         unique_ref_strs =
-            eel_g_hash_table_new_free_at_exit (g_str_hash, g_str_equal,
-                                               "unique eel_ref_str");
+            g_hash_table_new(g_str_hash, g_str_equal);
     }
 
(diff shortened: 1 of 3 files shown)


lukefromdc:
Caja now closes clean at exit, the only remaining warnings on a commmand line startup are this block  which both Nautilus and Nemo also get:
```
(caja:16495): Gtk-WARNING **: Failed to get the GNOME session proxy: The name org.gnome.SessionManager is not owned

(caja:16495): Gtk-WARNING **: Failed to get the Xfce session proxy: The name org.xfce.SessionManager is not owned

(caja:16495): Gtk-WARNING **: Failed to get an inhibit portal proxy: The name org.freedesktop.portal.Desktop is not owned
```

raveit65:
merged https://github.com/mate-desktop/caja/commit/6e142d59d7408e53ab1e579dba62517b6f59e307

monsta:
Should we apply https://git.gnome.org/browse/nautilus/commit/?id=0d46b0620b3b120cde310cb2319c309d8f3e3b82 to the remaining files and then drop that function?

lukefromdc:
That's probably a good idea. It's getting old and buggy

PR #654, opened 2016-10-09
GtkApplication: Disable session management when root/not in MATE

lukefromdc:
Limit session management to non-root sessions within MATE to stop restart loops in root sessions and inablity to kill off Caja in non-mate sessions such as XFCE. Also remove the autostarted hold-in previously used to stop restart loops.

egg_sm_client_set_mode only works when called before the main loop starts, thus it cannot be used in GtkApplication startup functions or otherwise later in the program.

Note that this commit restores the Caja 3.16 and earlier restart loop in normal mate sessions if a user disables icons on the desktop, then sets exit_with_last_window. The combination is the default for root/not in mate sessions though, so without the ability to set "no restart" later than the start of the main loop, session management outside of MATE and in root sessions has to be disabled.

Fixes https://github.com/mate-desktop/caja/issues/481


lukefromdc:
Fixes https://github.com/mate-desktop/caja/issues/481

raveit65:
Thanks, this fixes the restart issue in non Mate sessions.
I'm fine with it.

PR #658, opened 2016-10-14
Fix Gtk2 build issues with libunique and configure script error

emanuele-f:
Please use this as a reference. Building gtk3 without libunique does not work (edit: *with this patch*, for me). Possibly fixes #655.

PR #660, opened 2016-10-14
Fix missed selector change for separating GTK3 and libunique options

lukefromdc:
One selector was not updated in caja-location-bar.c to follow use or nonuse of libunique in GTK3 builds. Fixing this revealed need for a 3-way selector in one place in caja-application.c

Needs GTK2 test build, GTK3 builds both with and without libunique worked

### DIFF
--- a/src/caja-application.c
+++ b/src/caja-application.c
@@ -704,6 +704,8 @@ check_required_directories (CajaApplication *application)
 #if ENABLE_LIBUNIQUE == (0)
         gtk_application_add_window (GTK_APPLICATION (application),
                                     GTK_WINDOW (dialog));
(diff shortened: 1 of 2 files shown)


lukefromdc:
Note that this may or may not create a conflict with https://github.com/mate-desktop/caja/pull/658 if my PR https://github.com/emanuele-f/caja/pull/1 is merged into it to let non-libunique builds work, then it is merged into master to let GTK2 builds function correctly.  I will take care of that if this is merged first but we should wait to hear back about https://github.com/mate-desktop/caja/pull/658 first

raveit65:
I guess we can delete this branch?

lukefromdc:
Yes we can

PR #663, opened 2016-10-19
Add swap desktop icons when they move too close-rebased and works

lukefromdc:
based on https://github.com/mate-desktop/caja/pull/635 with 
https://github.com/mate-desktop/caja/commit/d3810cedb02b4660c2d1c771ae682868b6df32d4
to get swapping icons to work in the vertical plane.

Rebased on today's master, works well with GTK 3.22

### DIFF
--- a/libcaja-private/caja-icon-container.c
+++ b/libcaja-private/caja-icon-container.c
@@ -53,6 +53,7 @@
 #include <gdk/gdkx.h>
 #include <glib/gi18n.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
 
 #if !GTK_CHECK_VERSION(3, 0, 0)
@@ -265,6 +266,16 @@ static int compare_icons_vertical (CajaIconContainer *container,
                                    CajaIcon *icon_b);
 
 static void store_layout_timestamps_now (CajaIconContainer *container);
+
+typedef gboolean (*IsBetterIconFunction)(CajaIconContainer *container,
+                                         CajaIcon *start_icon,
+                                         CajaIcon *best_so_far,
+                                         CajaIcon *candidate, void *data);
+
+static CajaIcon *   find_best_icon (CajaIconContainer *container,
+                                    CajaIcon *start_icon,
+                                    IsBetterIconFunction function,
+                                    void *data);
 
 static gpointer accessible_parent_class;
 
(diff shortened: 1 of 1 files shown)


raveit65:
Ok, give me half an hour for testing :)

lukefromdc:
I just sucessfully tested this with GTK 3.16 though GTK 3.22, mixing up libunique and GtkApplication builds. This worked the same in every one of them

lukefromdc:
One issue I did find is this: when a pair of icons are swapped, killing Caja causes them to come back up on top of oneanother, apparently the swapped positions are not being stored persistantly

lukefromdc:
One issue I did find is this, and it is an issue with all icon position changes set by keep aligned. If icons are swapped, or if keep aligned is set and moves icons, killing Caja or logging out, then back in leaves the icons in the dropped or previous positions. Unsetting, then resetting keep aligned fixes this but this seriously degrades the functionality of all keep aligned functionality. I filed this new issue about that:

https://github.com/mate-desktop/caja/issues/665

This is not caused by the swap icons function, rather the swap icons function makes this issue more important than before. We can either merge this now and work on that issue, or hold this pending resolution of the position storage issue.  I will first see if Nautilus and Nemo have this problem and if they do not I will try to find what they do differently

lukefromdc:
Just found Nautilus and Nemo don't remember the changed positions from keep aligned either-and in Nautilus's case not at all even on a 2nd restart. On all three file managers only manual position changes seem to be saved

lukefromdc:
Found this old Nautilus bug report:
https://bugs.launchpad.net/ubuntu/+source/nautilus/+bug/12608

raveit65:
I would prefer to find a solution for this issue before we merge this.
Also I think the vertical 60px distance isn't enough, after the swap the vertical distance isn't the same like it looks with a fresh  aligned.
Any ways, i will create the 1.17.0 release yet.

lukefromdc:
OK, I will work on this on my branch, I'm looking at a bunch of Nautius stuff now. We can merge this  when and if I or someone else can fix the keep aligned issue. That would be a bugfix as well as a new feature.

lukefromdc:
gvfs-info -a metadata::caja-icon-position <any desktop icon moved by keep aligned> does not update when an icon is moved by keep aligned or this PR's changes to it. It updates when and only when an icon is moved by hand. How is this stored, and how do I get it to update when an icon is auto-moved?

lukefromdc:
In digging through the code I see why Nautilus never fixed this: the file is needed to set the metadata string, but the auto move functions only use the icon itself, not the file and thus do not have access to the file. 

It looks like icon_position_changed_callback in fm-icon-view.c is called by the icon_position_changed signal, and I tried sending that signal explicitly with no effect. No surprise, without the file there is no way to store the metadata. Thus to fix this, any time icons are snapped it would be necessary to iterate through every file on the desktop, get their positions, and store the positions.

lukefromdc:
Should I close this? None of the other Nautilus based file managers have fixed this, and there seems to be no easy way to do so. Can leave it open if we think someone can find a way to save the swapped positions as in my previous comment, or close both this and the original https://github.com/mate-desktop/caja/pull/635 if we think that is not going to get done

raveit65:
I think , we can close both. The result wasn't sufficient if i remember correctly.

PR #667, opened 2016-10-23
GtkApplication: Notify when ejected volumes safe to unplug

lukefromdc:
Notify users when an external drive is still writing data on attempt to eject, and again when drive is safe to unmount. Notifications behave as they do in Nemo when ejecting/unmounting a flash drive. Tested with GTK 3.22 and a known slow flash drive.

This is in GtkApplication builds only, code is from Nemo. FIXME: build system does not limit libnotify dependency to GtkApplication builds and is rather improvised. Adding libnotify to CORE_MODULES via configure.ac did not work, and using Nemo's changes put -lnotify into CORE rather than ALL libs and that caused build failures on failure to link.

A better way to add libnotify to the linked libraries in "CORE" and to limit the dependency to GtkApplication builds during configuration would be nice, but this works as-is and mate-notification-daemon already depends on libnotify so no new dependencies are added to MATE as a whole

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -11,6 +11,7 @@ m4_define(xml_minver,                  2.4.7)
 m4_define(exif_minver,                 0.6.14)
 m4_define(exempi_minver,               1.99.5)
 m4_define(gail_minver,                 0.16)
+m4_define(notify_minver,               0.7.0)
 
 
 dnl 1. If the library code has changed at all since last release, then increment revision.
--- a/libcaja-private/Makefile.am
+++ b/libcaja-private/Makefile.am
@@ -39,6 +39,7 @@ libcaja_private_la_LIBADD = \
 	$(top_builddir)/eel/libeel-2.la \
 	$(top_builddir)/libcaja-extension/libcaja-extension.la \
 	$(CORE_LIBS) \
+    -lnotify
 	$(NULL)
 
 libcaja_private_la_SOURCES = \
--- a/libcaja-private/caja-file-operations.c
+++ b/libcaja-private/caja-file-operations.c
@@ -55,7 +55,9 @@
 #include <gtk/gtk.h>
 #include <gio/gio.h>
 #include <glib.h>
-
+#if ENABLE_LIBUNIQUE == (FALSE)
+#include <libnotify/notify.h>
+#endif
 #include "caja-file-changes-queue.h"
 #include "caja-file-private.h"
 #include "caja-desktop-icon-file.h"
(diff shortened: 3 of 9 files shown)


lukefromdc:
I've more than once lost data on flash drives due to not having this in Caja.

lukefromdc:
Note that just like in Nemo, the notifications are only triggered by unmounting from the sidebar eject button, not from the right-click menu on a desktop icon for the drive.

raveit65:
Is there a way to make that configurable (not default)?
For me the linux desktop sends to much notifications and i can can pretty live without it.
Personal i would disable 80% of current notifications if this would be possible.
But i agree that is a nice feature for some ppl.

lukefromdc:
Probably could be done, would mean setting up a gsettings preference and reading it. That would permit it to be turned off in gsettings. Adding it to the file management preferences dialog would be a bit more complex but probably doable. 

In both gnome-shell and cinnamon, whether to show notifications is globally configurable for the entire DE. We could do that in mate-notification-daemon by simply including a preference to enable or disable it.

lukefromdc:
Just added notifications for ejecting from the desktop too

lukefromdc:
I've now refactored this to use the libcaja-private function needed for notifications from desktop eject operations for sidebar eject operations too. In the process I was able to enable this for libunique builds as well as GtkApplication builds. If someone would test this with GTK2 we would know if can be merged prior to the GTK3-only transition or could me with a bit more work.

I want to add a dconf option to turn this on and off as requested next, will default to "on."

lukefromdc:
Gsettings preference added

lukefromdc:
The notifications work well, warning if data is still being written, then telling the user the drive is safe to eject. The gsettings option works, it comes up in dconf-editor as org>mate>caja>preferences>show-notifications and (at least here) appears just underneath "show-image-thumbnails". Test is with GTK 3.22. Can someone test this with GTK2 before I merge this?

lukefromdc:
I set up a GTK2 partition, no issues there and the notifications worked fine. Worked from a Cinnamon session there (only built enough with GTK2 to build caja) and this connected fine with cinnamon's notification daemon as well. No issues in tests with GTK 3.16, 3.18, 3.20, or 3.22 so merging this

lukefromdc:
Merged -ff-only
https://github.com/mate-desktop/caja/commit/ec529e70c20c4164fe34179fc1956676bd10e665
through
https://github.com/mate-desktop/caja/commit/570d6afed8ff62e430f46b3d87870702240ce3e8

lukefromdc:
Probably could be done, would mean setting up a gsettings preference
and reading it. That would permit it to be turned off in gsettings. Adding
it to the file management preferences dialog would be a bit more complex
but probably doable.

Issue #535, opened 2016-04-21
Copying directories with same name


sc0w:
with this structure (all directories)

```
testcaja
|
|-----aa
|      |-----bb
|
|-----bb
```

I select the subdir bb of aa, right button and copy

I go to the dir testcaja, over a white space, right button and paste

![bugcaja](https://cloud.githubusercontent.com/assets/7734191/14719241/78f7dbb4-07fb-11e6-9897-06940995bf93.png)

Original file? <--- It is not a file

Size: (null) <--- null?

by the way, the copy is sucesfull
##

sc0w:
now I still reproduce the "null" issue, but I can't with new user account, so I think this issue can be closed

PR #673, opened 2016-11-03
Improved merge folder dialog

sc0w:
based on nautilus commit:
https://git.gnome.org/browse/nautilus/commit/?id=b370817

Fixes partially #535 

**currently:**

modify_caja_folder_before

**with my PR:**

caja_folder_later_classic

we only need to solve: **Null**

By the moment I don't know how to solve it

### DIFF
--- a/libcaja-private/caja-file-conflict-dialog.c
+++ b/libcaja-private/caja-file-conflict-dialog.c
@@ -306,8 +306,14 @@ file_list_ready_cb (GList *files,
     }
 
     str = g_string_new (NULL);
-    g_string_append_printf (str, "<b>%s</b>\n", _("Original file"));
-    g_string_append_printf (str, "%s %s\n", _("Size:"), size);
+    if (dest_is_dir) {
+        g_string_append_printf (str, "<b>%s</b>\n", _("Original folder"));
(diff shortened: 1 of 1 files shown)


lukefromdc:
I can confirm that this builds and runs on GTK 3.22, and that the dialog is cleaned up as shown. There is one difference here though: I don't get the "Items(null)" issue for folder contents, rather I get "Items (0 items)" both with and without this commit. The rest works as described.

lukefromdc:
Merged via https://github.com/mate-desktop/caja/commit/1ca6c5f9c7ced0f906de5dda119cdd37c76d8cfa

monsta:
@sc0w: I have the same results as @lukefromdc - I can't reproduce the "null" problem. The dialog shows "0 items" instead. I used the same steps to reproduce as you posted in https://github.com/mate-desktop/caja/issues/535. Do you still get the issue?

Tested with:
- Caja 1.16.5, GTK+2 build
- Caja 1.16.5, GTK+3 build, GTK+ 3.20.9
- Caja from git master, GTK+ 3.22.11

sc0w:
@monsta ok, thanks for your tests

I see it with caja 1.16.2 gtk 3.22.11

I will try with latest versions

PR #674, opened 2016-11-04
Fix mousewheel scrolling in compact view

lukefromdc:
Apply Nautilus commit https://github.com/GNOME/nautilus/commit/0832acdb4371fc7de957303e220e899c8fdcdf5c

This causes vertical scroll wheel movement from the mouse to scroll the COMPACT view horizontally, the direction the scrollbars run in compact view. Previously there was no mousewheel scroll in compact view at all. NOT sure if we want to do this, Nemo does and has had some complaints by users wanting to turn it off. Should be possible to add a dconf key for that if people think that's necessary-or we could just leave this as-is.

### DIFF
--- a/src/file-manager/fm-icon-view.c
+++ b/src/file-manager/fm-icon-view.c
@@ -2507,16 +2507,15 @@ fm_icon_view_scroll_event (GtkWidget *widget,
 
             scroll_event_copy = (GdkEventScroll *) event_copy;
 
-           /* transform vertical integer smooth scroll events into horizontal events */
-           if (scroll_event_copy->direction == GDK_SCROLL_SMOOTH &&
-               scroll_event_copy->delta_x == 0) {
-                   if (scroll_event_copy->delta_y == 1.0) {
-                       scroll_event_copy->direction = GDK_SCROLL_DOWN;
-                           } else if (scroll_event_copy->delta_y == -1.0) {
-                               scroll_event_copy->direction = GDK_SCROLL_UP;
-                               }
-                           }
-            if (scroll_event_copy->direction == GDK_SCROLL_UP)
+            /* transform vertical integer smooth scroll events into horizontal events */
+            if (scroll_event_copy->direction == GDK_SCROLL_SMOOTH && scroll_event_copy->delta_x == 0) {
+                if (scroll_event_copy->delta_y == 1.0) {
+                    scroll_event_copy->direction = GDK_SCROLL_DOWN;
+                } else if (scroll_event_copy->delta_y == -1.0) {
+                    scroll_event_copy->direction = GDK_SCROLL_UP;
+                }
+            }
+            if ((scroll_event_copy->direction == GDK_SCROLL_UP) || (scroll_event_copy->delta_x == -1.0))
             {
                 scroll_event_copy->direction = GDK_SCROLL_LEFT;
             }
--- a/src/file-manager/fm-icon-view.c
+++ b/src/file-manager/fm-icon-view.c
@@ -2495,9 +2495,9 @@ fm_icon_view_scroll_event (GtkWidget *widget,
     icon_view = FM_ICON_VIEW (widget);
 
     if (icon_view->details->compact &&
-            (scroll_event->direction == GDK_SCROLL_UP ||
-             scroll_event->direction == GDK_SCROLL_DOWN))
-    {
+        (scroll_event->direction == GDK_SCROLL_UP ||
+        scroll_event->direction == GDK_SCROLL_DOWN ||
+        scroll_event->direction == GDK_SCROLL_SMOOTH)) {
         ret = fm_directory_view_handle_scroll_event (FM_DIRECTORY_VIEW (icon_view), scroll_event);
         if (!ret)
         {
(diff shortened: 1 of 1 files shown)


sc0w:
your PR works for me as expected

I think it is usefully for the people that use the compact view

I agree with this PR

sc0w:
wait!

your PR have one problem:

my mouse have horizontal and vertical scroll wheel

with the vertical scroll it works as expected and your PR works

**the problem is with the horizontal scroll wheel**:

- without your PR, it works as expected, and I can move to left and right

- with your PR, **always** go to the **right** if I move that wheel to the left or right

see the picture, my mouse is something like this:

horizontalmouse

sc0w:
I solved the problem:

I modify the line 2519

`if (scroll_event_copy->direction == GDK_SCROLL_UP)`

by this:

`if ((scroll_event_copy->direction == GDK_SCROLL_UP) || (scroll_event_copy->delta_x == -1.0))`

and fix some alineations

**my patch:**

```patch
diff --git a/src/file-manager/fm-icon-view.c b/src/file-manager/fm-icon-view.c
index fe6148e..d598051 100644
--- a/src/file-manager/fm-icon-view.c
+++ b/src/file-manager/fm-icon-view.c
@@ -2507,16 +2507,15 @@ fm_icon_view_scroll_event (GtkWidget *widget,
 
             scroll_event_copy = (GdkEventScroll *) event_copy;
 
-           /* transform vertical integer smooth scroll events into horizontal events */
-           if (scroll_event_copy->direction == GDK_SCROLL_SMOOTH &&
-               scroll_event_copy->delta_x == 0) {
-                   if (scroll_event_copy->delta_y == 1.0) {
-                       scroll_event_copy->direction = GDK_SCROLL_DOWN;
-                           } else if (scroll_event_copy->delta_y == -1.0) {
-                               scroll_event_copy->direction = GDK_SCROLL_UP;
-                               }
-                           }
-            if (scroll_event_copy->direction == GDK_SCROLL_UP)
+            /* transform vertical integer smooth scroll events into horizontal events */
+            if (scroll_event_copy->direction == GDK_SCROLL_SMOOTH && scroll_event_copy->delta_x == 0) {
+                if (scroll_event_copy->delta_y == 1.0) {
+                    scroll_event_copy->direction = GDK_SCROLL_DOWN;
+                } else if (scroll_event_copy->delta_y == -1.0) {
+                    scroll_event_copy->direction = GDK_SCROLL_UP;
+                }
+            }
+            if ((scroll_event_copy->direction == GDK_SCROLL_UP) || (scroll_event_copy->delta_x == -1.0))
             {
                 scroll_event_copy->direction = GDK_SCROLL_LEFT;
             }

lukefromdc:
I just applied this patch, it built and ran fine with no adverse effects, though I can't directly test the horizontal scrollwheel function with my $5 mouse. Should I merge this or wait a day or so for anyone else to weigh in?

lukefromdc:
I had to put this behind GTK3-only selectors to fix the build with GTK2. In GTK2 mousewheel scrolling works fine with the old code anyway, which means this should be merged as it makes the GTK2 and GTK3 behavior consistant.

sc0w:
I have tested with your last changes and works for me (gtk 3.22.2)

> Should I merge this or wait a day or so for anyone else to weigh in?

@raveit65 @monsta 

are you agree with this PR ?

kn00tcn:
agree to merge, though i'm not set up to test

@sc0w nice catch on the horizontal wheel, i forgot about those mice

lukefromdc:
I can confirm sucessful builds with GTK 3.22, GTK 3.18, and with GTK2 on another partition

monsta:
> Previously there was no mousewheel scroll in compact view at all. NOT sure if we want to do this, Nemo does and has had some complaints by users wanting to turn it off.

I just tried it in GTK+2 build and it works there out of the box, so people are probably familiar with it for some time - maybe even since gnome 2 days?
And since compact view never shows any vertical scrollbars by design, I wonder why someone would need to turn this off :smile:

monsta:
This PR works for me with 3.22.2, using a simple mouse with vertical scroll only. I'll be able to test horizontal wheel movement at work on Monday (I have a mouse with this feature at work).

raveit65:
My goal is to work with linux w/o using a mouse :smile: and i prefer to use iconview,
.... but i'm fine with this PR if it works.

monsta:
So I tested it with my mouse at work (Logitech RX300). Horizontal wheel movement works in GTK+2 build of Caja... but it does nothing with 3.22.2, with this PR or without it. :confused:

Vertical scrolling works fine.

monsta:
Ok, so I also checked ```mate-system-monitor``` which is GTK+3 only now. It's possible to test horizontal scrolling in the process table, if you resize the columns enough.

- host system, 3.14.5: horizontal scrolling works fine
- VM, 3.18.9: doesn't work
- VM, 3.20.9: doesn't work
- VM, 3.22.2: doesn't work

So I suspect it's VirtualBox that somehow prevents horizontal scrolling events from getting into the guest system. Can anyone confirm that horizontal scrolling doesn't work in a VM?

sc0w:
horizontal scrolling works for me on a virtual machine, but I use virt-manager, and gtk2 on it

It seems to be a problem with virtualbox, you can see:

https://forums.virtualbox.org/viewtopic.php?f=1&t=1794&start=0

https://www.virtualbox.org/ticket/700

sc0w:
sorry, I remembered that works but no,  the horizontal scroll doesn't work on my VM, I have tested now (gtk2 without this PR)

conclusion, it needs to test on a host system

monsta:
Ok, so it's indeed VirtualBox issue. Let's merge it then.

lukefromdc:
Merged --ff-only with 
https://github.com/mate-desktop/caja/commit/f34590dd7a71251515b90cd9812878b7d26d30b6
through
https://github.com/mate-desktop/caja/commit/3e93f938000cdb24a6137daecd77fe010ae93f36

PR #678, opened 2016-11-08
remove convert metadata

lukefromdc:
This appears to be dead code in Caja anyway

Apply https://git.gnome.org/browse/nautilus/commit/?id=87d24047f674ca941e2691602da2501945a82330


lukefromdc:
I can confirm no problems here with build including newest commit to remove the unused function

Issue #688, opened 2016-12-01
Invisible cursor in dark themes, or wrong text/bg color in rename textbox


aykevl:
When renaming a file while using any of the dark themes (BlackMATE, Blackbird, High Contrast Inverse), the cursor is practically invisible.
This  occurs in Icons view and Compact view, List view works properly. In the non-working views, the textbox has black text on a white background with a light cursor. In List view, the textbox has the style of the rest of the theme (dark background, white text), making the cursor visible. I prefer the text box as in the List view, as it is more consistent with the rest of the theme. Maybe the issue really is that the textbox has mostly wrong colors, not following the rest of the theme (except for the cursor, apparently).

Screenshot of a rename (note the almost invisible cursor between `u` and `n`):
![screenshot at 2016-12-01 22-17-52](https://cloud.githubusercontent.com/assets/729697/20813055/7387a35c-b814-11e6-90cf-244cb69e7e13.png)

System: Debian stretch (testing), Caja 1.16.1.

raveit65:
Imo, this can be set with carret-color in themes.

lukefromdc:
The right place to set this is in caja-css, so ALL dark themes are fixed. There's only one problem and it is a nasty one: caret-color is only recognized by GTK 3.20 or later so if it's in caja.css in GTK 3.18 or earlier the file throws an error and won't load. Thus we would need to different versions of caja.css and either a buildtime selector or a runtime selector to pick the right version. Doing it at build time requires that configure.ac detect/declare the GTK versionagain and select the right file, not sure how to do that. I could easily set this up for runtime selection with two files, but installing two when only one will be loaded is a bit more bloat.

If we don't want to fool with that sort of thing at all I can easily fix this in BlackMATE but that leaves the other dark themes broken here. Not sure what to do

raveit65:
Using different caja.css sounds very complicated for me and isn't the carret-color not different for light themes?
I prefer to fix it in BlackMate only and 3 party themes can use it as reference.

raveit65:
I forgot to say, for me this is a theme issue as gtk+ provide a setting for it.

lukefromdc:
https://github.com/mate-desktop/mate-themes/commit/1147914097234d4d6fda9aff0353d7e698fd7667

fixes this for BlackMATE in master

lukefromdc:
The gtk 3.22 and 3.20 branches of BlackMATE are out of sync, I can't fix that right now. That needs to be resolved so this can be applied to them

[...]

PR #691, opened 2016-12-07
Backport rename label fix to 1.16

lukefromdc:
Fixes https://github.com/mate-desktop/caja/issues/688
Backport of https://github.com/mate-desktop/caja/commit/aa7af8673a3d894fc99a00d1a30ea57f44e3e1c5 to version 1.16

*Split caja.css. Much of it can load with GTK_STYLE_PROVIDER_PRIORITY_THEME to allow themes that support Caja to set their own style. CSS required for the desktop background to show and to keep the white text suitable for most backgrounds needs GTK_STYLE_PROVIDER_PRIORITY_APPLICATION so it is loaded separately from new file caja-desktop.css

*Move all css items that must not be overridden by themes (most of them are for the desktop) to caja-desktop.css and keep those that are only needed when themes do not set them at all in caja.css

### DIFF
--- a/data/Makefile.am
+++ b/data/Makefile.am
@@ -34,6 +34,7 @@ cajadata_DATA = \
 	caja-extras.placeholder  \
 	caja-suggested.placeholder \
 	caja.css \
+    caja-desktop.css \
 	$(NULL)
 
 # app data file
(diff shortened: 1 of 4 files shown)


raveit65:
Tested with gtk+-3.20 on top of 1.16.1 tarball, works fine here.

Issue #530, opened 2016-03-31
Accessible text of icons is concatenated without spaces


Joanmarie:
Steps to reproduce:
1. Enable display of file size in icon view
2. Switch to icon view
3. In a terminal, launch the attached pyatspi accessible-event listener
4. Arrow around in Caja to change the selected icon
5. Examine the text reported by the accessible-event listener

Expected results: There would be whitespace (e.g. a newline char) separating each line of text displayed beneath the icon.

Actual results: There is no whitespace separating each line of text displayed beneath the icon.

Sample output:
installer_linux.sh111.0 MB
index.html1.1 MB
mt7601u-master.zip68.9 kB

Impact: Orca speaks the two lines as if they are a single line because all it knows is what it gets when it asks for the displayed text via AT-SPI2.

Most of the time this is not a big deal because speech synthesizers speaking things like "zip68.9" will have to say "zip 68.9". However, if the file extension ends with a digit, such as "mp4", you wind up with accessible text like "foo.mp4109.8 MB", and that text gets spoken as "foo.mp 4109.8 MB" (where "4109" will, depending on the user's settings, be spoken as "four thousand one hundred nine").

Listener: Delete the 'txt' extension before launching it. Github didn't like the ".py".
[text.py.txt](https://github.com/mate-desktop/caja/files/197858/text.py.txt)
##

coffeeking:
hi. This happens in mate 1.12 as well as mate 1.10. Thanks joanie for the pyatspi listener

lukefromdc:
On my files, this gives a whitespace only if there is a whitespace in the filename, with Caja 1.17.1.  The only text under them is the filename, which is a single string, not a concatenation of strings.

Adding a whitespace to a filename where none actually exists makes the filename incorrect, meaning if it was read back exactly the same way the file would not be found.

lukefromdc:
It's worth keeping in mind, with a filename under an icon there is one string, and the only thing breaking it into multiple lines is the limit to the length of one line. This changes with options such as "use compact layout"  which is not the same thing as compact view.

cwendling:
> On my files, this gives a whitespace only if there is a whitespace in the filename, with Caja 1.17.1. >The only text under them is the filename, which is a single string, not a concatenation of strings.
>
> Adding a whitespace to a filename where none actually exists makes the filename incorrect, meaning if it was read back exactly the same way the file would not be found.

[...]

PR #692, opened 2016-12-14
Add a separator when concatenating strings in icon text accessible

cwendling:
This makes the string exposed to ATK more sensible, instead of having unrelated words coupled together.

### DIFF
--- a/libcaja-private/caja-icon-canvas-item.c
+++ b/libcaja-private/caja-icon-canvas-item.c
@@ -3564,6 +3564,8 @@ caja_icon_canvas_item_accessible_factory_create_accessible (GObject *for_object)
     }
     if (item->details->additional_text)
     {
+        if (item_text->len > 0)
+            g_string_append_c (item_text, ' ');
         g_string_append (item_text, item->details->additional_text);
     }
     item->details->text_util = gail_text_util_new ();


lukefromdc:
I can confirm that this builds and runs (with GTK 3.22) when accessability is not turned on, but that it all. On my rather heavily hacked Debian setup speech dispatcher and Orca have never worked when installed for some reason. If anyone else here can get a test with Orca and confirm that part working this should be mergable.

lukefromdc:
https://github.com/raveit65 and https://github.com/monsta , can either of you test this with a functioning Orca install to verify audio works with it?  If not should I go ahead and merge this? It looks simple enough but I can't get Orca working and thus can't test the accessability function of this.

raveit65:
I am quite busy with reconstruction my flat in real life, hope that i can finished that before christmas.
Than i can test more PRs and start codding again.
Accessible support is sensitive stuff, so maybe one of the guys who use it can test it?
@coffeeking
@TheMuso
...please...

lukefromdc:
OK, I won't merge this until it's had a proper test by someone who uses it.

cwendling:
@lukefromdc @raveit65 it's @Joanmarie that pointed the issue to me, and you can see the problem e.g. using Accerciser.  When additional info is displayed under the name, like the element size, you'll see they are concatenated without any kind of separator.

See before the patch (`caja.css2,4 kB`): 
caja-before

And after (`caja.css 2,4 kB`):
caja-after

lukefromdc:
I have zero experience with the python console and this is the first I've heard of Accerciser. I've never developed on the accessability code, simply because it does not run on my setup which is loosely based on Debian Unstable but has diverged a long way.  I will leave this to those more familiar with it, as https://github.com/raveit65 has advised.

I just don't want to start merging accessability fixes that I cannot test, then when 1.18 comes out accessaibility does not work at all for anyone because I missed something I could not test. Someone else would then have to clean up the mess, and distros would have already release the bad code.

coffeeking:
hi

That's a great idea. I might be able to help a little, although I can't 
code either. Joanmeri diggs says that all that should be necessary is to 
add a space between the accessible description and the number of items. 
Example, test.ogg 3.2 mb instead of test.ogg3.2 mb. Also, this is an 
entirely unrelated comment but I figured I'd better bring it up while 
I'm thinking about it. Why does caja report 0 items even though there 
are usually not 0 items in the view? Example, if you open a folder full 
of mp3 or ogg files caja says icon view layored pane 0 items. This could 
be a similar bug in the accessibility code.

lukefromdc:
https://github.com/coffeeking, if you can test the existing PR and your test shows that it works as intended with accessability, I can then merge it.

coffeeking:
hi

I'll do my best. How do I test? Pull caja from git and build??

lukefromdc:
https://github.com/coffeeking , just pull
https://github.com/cwendling/caja/tree/at/icon-text
build it, and test run with Orca to see if this works as intended

coffeeking:
hi

Got it, I'll do that right now. Let me boot into linux on this dell, and 
I'll test that right now. I'll keep this email and report back with my 
results. What am I looking for so I know if it works? I have to 
configure caja to show size information in the icon captions?

Thanks

Kendell Clark

cwendling:
It's a good think not wanting to merge things you can't test, but I think the change here is fairly innocent :)  One could wonder if a space is the best possible separator, but well.

If anything, it's questionable to put the number of items there -- or actually, implementing the text interface altogether? -- especially as the accessible description and name should already be enough (and actually, at some zoom levels, redundant).  But well, I prefer not to remove it without knowing why it was added in the first place :) It comes from https://git.gnome.org/browse/nautilus/commit/?id=684be6bd53a944e48b04392414c8cb320f916db7, but the motivation is unclear.
Also, the stuffing together in question here comes from https://git.gnome.org/browse/nautilus/commit/?id=42a6c32020780b11279449f4ebb48a2187cb7dbe, again without clear reason for it.

coffeeking:
hi

One final question. Is there a reason why the number of items is 
repeated twice? Or rather, the size of a file and the number of items in 
a folder. Example, test.ogg 3.2 mb ogg audio. 1 of 1 3.2 mb. Test 13 
items folder, 1 of 1 13 items. Is this test branch built against gtk2 or 
gtk3, or does it matter, it will work on either? I use the gtk3 version 
of mate in my linux distribution called sonar, but if I build caja 
against gtk2 on a gtk3 based mate desktop, will it still function? I'm a 
little new at this so excuse the questions. I'm just about to build and 
test, and then I'll report back. I fell asleep before I could yesterday :).

lukefromdc:
https://github.com/coffeeking, I have no idea why the repetition you asked about is used

coffeeking:
hi

Your caja branch works perfectly. It even seems to perform slightly 
faster than the released version of caja.  Orca no longer reads file 
sizes as 30 mb when they are really 3.0 mb. Fantastic job!

lukefromdc:
OK, i will merge this

lukefromdc:
Merged with
 https://github.com/mate-desktop/caja/commit/17895425eb9fc73e5a4d71f8321954d83f182137

cwendling:
@coffeeking 
> One final question. Is there a reason why the number of items is repeated twice?

I didn't investigate deeply, but this is likely as I mentioned because there is a duplication of information: the *name* and *description* should be enough, but icons also implement *text*, and report the item name plus some of its description through that.  If Orca reads *text* + *description*, there will be some redundancy.

As said, I can't figure out why the icons implemented *text* in the first place though.

@lukefromdc thanks!

coffeeking:
hi

No worries, it's not a huge issue, more of a curiosity. Sounds like 
your'e saying it's my orca settings, which is more than possible. Our 
linux distro has orca set to verbose mode by default, which means it 
reads out a lot more information than the default brief mode. As for the 
icons implementing text, I'm wondering if the nautilus guys did it 
before it was forked into caja? Either way I'm glad this bug got fixed 
and I'm extremely impressed as always with you guys' dedication. Not 
many projects, no matter what OS they're supporting, care about 
accessibility enough unless they're getting paid. It's nice to have that 
change.

cwendling:
> Sounds like your'e saying it's my orca settings, which is more than possible. 

No, I'm saying that Caja exposes redundant information to Orca.  That Orca reads both may depend on the setting, but it's not wrong, the problem is reporting the same information twice in complementary fields.

> As for the icons implementing text, I'm wondering if the nautilus guys did it before it was forked into caja?

Yes, long before.  As pointed in my previous comment, I traced those back to 2002 and 2003, but without explanations on why they were introduced.

monsta:
Guys, does it fix https://github.com/mate-desktop/caja/issues/530?

alexarnaud:
Le 06/01/2017 à 19:30, monsta a écrit :
> Guys, does it fix #530 <https://github.com/mate-desktop/caja/issues/530>?

Yes, it resolves the issue you have mentioned.

Best regards.
-- 
Alex ARNAUD

monsta:
Hmm, looks like @lukefromdc found an issue... https://github.com/mate-desktop/caja/issues/530#issuecomment-270973350

cwendling:
@monsta as mentioned on #530, I believe there's a misunderstanding on what this is supposed to do.  I hope I have clarified it there, though :)

Issue #91, opened 2013-03-20
Index files "index.theme" behaviour


MDykstra:
In **"~/.theme/some-theme/"** the **"index.theme"** is shown by its **"Name"** field, as specified inside the **"index.theme"**. Right-clicking on the file does not allow to edit the **"index.theme"** file by selecting pluma. Copying a **"index.theme"** file to another theme folder, thereby replacing an existing **"index.theme"** file, renames the **"index.theme"** file to something else.
##

NiceandGently:
related fixes for nautilus are here:
https://git.gnome.org/browse/nautilus/commit/?h=gnome-3-4&id=b9c51fd4eb84a9f52ee4be1bf183fc516984130b
https://git.gnome.org/browse/nautilus/commit/?id=c1551df
https://bugzilla.gnome.org/show_bug.cgi?id=319981
https://bugzilla.gnome.org/show_bug.cgi?id=672809
But code is too different from nautilus to apply them,
but maybe this helps to find a fix for caja.

erikdubois:
strange that this issue is not yet resolved. As an icon theme designer I have the need to change this file.
Linux Mint 18 Cinnamon and 18.1 open such a file with any editor you like. In mate we get an error.
![index](https://cloud.githubusercontent.com/assets/10594806/21294665/648c0a6a-c542-11e6-90b0-c5e80006e66e.png)
I resolve it now by opening e.g. sublime text and drag/drop index.theme on sublime text to open it.

raveit65:
I tried again to apply the commits from nautilus. They apply now but i get an error with open the file.
![bildschirmfoto zu 2016-12-18_20-12-53](https://cloud.githubusercontent.com/assets/961604/21295860/719fbad2-c55e-11e6-8dc7-a3c4ea7cdb69.png)
Resulting patch is here https://gist.github.com/raveit65/19cfc5fe706ac6a2ea5d663a94fb02af

lukefromdc:
It seems to me that mimetype "theme(application-x-theme) should be treated as a text file by Caja. Obviously Caja it is trying to open these with something else with this patch, no idea what that application is as no text gets sent to terminal if started from there.  Just removing the code that hides the right-click menu for .desktop files might help-and also would make it easier to edit launchers without having to drag and drop them into Pluma

lukefromdc:
https://github.com/mate-desktop/caja/pull/694
[...]

PR #694, opened 2016-12-19
Show "open with" menu for all index.theme files, open them in Pluma

lukefromdc:
Fix for https://github.com/mate-desktop/caja/issues/91
also applies when user wants to edit .desktop files
This does not change left-click behavior, desktop files launch as before and index.theme files will still generate an error on left-click, but the right-click menu can now be used to open either filetype in Pluma

### DIFF
--- a/libcaja-private/caja-file.c
+++ b/libcaja-private/caja-file.c
@@ -6964,9 +6964,12 @@ caja_file_get_symbolic_link_target_uri (CajaFile *file)
 gboolean
 caja_file_is_caja_link (CajaFile *file)
 {
-	/* NOTE: I removed the historical link here, because i don't think we
-	   even detect that mimetype anymore */
-	return caja_file_is_mime_type (file, "application/x-desktop");
+    if (file->details->mime_type == NULL) {
+        return FALSE;
+        }
+    return g_content_type_equals (eel_ref_str_peek (file->details->mime_type),
+                                                   "application/x-desktop");
+	
 }
 
 /**
--- a/libcaja-private/caja-mime-actions.c
+++ b/libcaja-private/caja-mime-actions.c
@@ -860,9 +860,7 @@ get_default_executable_text_file_action (void)
 gboolean
 caja_mime_file_opens_in_view (CajaFile *file)
 {
-    return (caja_file_is_directory (file) ||
-            CAJA_IS_DESKTOP_ICON_FILE (file) ||
-            caja_file_is_caja_link (file));
+    return (caja_file_is_directory (file));
 }
 
 static ActivationAction
@@ -871,7 +869,7 @@ get_activation_action (CajaFile *file)
     ActivationAction action;
     char *activation_uri;
 
-    if (caja_file_is_launcher (file))
+    if (caja_file_is_caja_link (file))
     {
         return ACTIVATION_ACTION_LAUNCH_DESKTOP_FILE;
     }
--- a/src/file-manager/fm-directory-view.c
+++ b/src/file-manager/fm-directory-view.c
@@ -4711,9 +4711,7 @@ reset_open_with_menu (FMDirectoryView *view, GList *selection)
 
 		file = CAJA_FILE (node->data);
 
-		other_applications_visible &=
-			(!caja_mime_file_opens_in_view (file) ||
-			 caja_file_is_directory (file));
+		other_applications_visible &= (!caja_file_is_directory (file));
 	}
 
 	default_app = NULL;
--- a/src/file-manager/fm-properties-window.c
+++ b/src/file-manager/fm-properties-window.c
@@ -4978,7 +4978,6 @@ is_a_special_file (CajaFile *file)
 {
 	if (file == NULL ||
 	    CAJA_IS_DESKTOP_ICON_FILE (file) ||
-	    caja_file_is_caja_link (file) ||
 	    is_merged_trash_directory (file) ||
 	    is_computer_directory (file)) {
 		return TRUE;
@@ -4992,7 +4991,7 @@ should_show_open_with (FMPropertiesWindow *window)
 	CajaFile *file;
 
 	/* Don't show open with tab for desktop special icons (trash, etc)
-	 * or desktop files. We don't get the open-with menu for these anyway.
+	 * We don't get the open-with menu for these anyway.
 	 *
 	 * Also don't show it for folders. Changing the default app for folders
 	 * leads to all sort of hard to understand errors.


raveit65:
I applied the change for fm-directory-view.c but i don't see the option open with in right-click menu.
```
diff --git a/src/file-manager/fm-directory-view.c b/src/file-manager/fm-directory-view.c
index 02a22da..3370941 100644
--- a/src/file-manager/fm-directory-view.c
+++ b/src/file-manager/fm-directory-view.c
@@ -4712,8 +4712,7 @@ reset_open_with_menu (FMDirectoryView *view, GList *selection)
 		file = CAJA_FILE (node->data);
 
 		other_applications_visible &=
-			(!caja_mime_file_opens_in_view (file) ||
-			 caja_file_is_directory (file));
+			(!caja_mime_file_opens_in_view (file));
 	}
 
 	default_app = NULL;

```
Is this the same what you use local?

lukefromdc:
menu

Your patch looks the same, it's certainly what I was trying to so. The attached image shows the "open with" menu that does not appear without this change. No idea what you did differentlyt, suggest pulling the PR and building from it. Note that this is not a patch to your patch, rather it is applied directly to master

lukefromdc:
Try the build again with the second commit. It turns out that the change to  get_default_executable_text_file_action in libcaja-private/caja-mime-actions.c is also required to get the right-click menu to show.  When I first got this working, I had tried 
https://gist.github.com/raveit65/19cfc5fe706ac6a2ea5d663a94fb02af , got it running but it didn't change anything, soI got the right-click menu working, then reverted the other changes. Turned out "make" did not rebuild libcaja-private when I did so, leaving it applied although it was no longer in the source. Thus the second commit re-applies this and it now works.

lukefromdc:
Just redid the entire PR, it is essentially just your patch now since that makes the right-click "open with" menu work and is the least difference from what GNOME did.  I think another GNOME commit altogether must control the default action on left clicking on an index.theme file, which in Nemo will open in Pluma or Gedit but still gives an error message in Caja.

In other words, this PR/the Raveit65 G'ist patch makes the "open with" right-click menu appear for desktop and index.theme files and that's all it does.  That's enough to save users the drag and drop hassle with them, and we can keep looking for the issue with the default.

Nemo and Nautilus btw also seem to distinguish among more subvarieties of text files in assigning icons, which is also a mimetype function.  There might yet be a commit to Nautilus that if sucessfully applied  bring this functionality to Caja and solves the default action on index.theme file issue.

lukefromdc:
PR rebased after
https://github.com/mate-desktop/caja/commit/17895425eb9fc73e5a4d71f8321954d83f182137

raveit65:
the open-with entry is there now but....
1. selecting pluma from list is a oneshot and using pluma isn't saved.
2. i need to add pluma with the command line entry in dialog window for saving pluma for application/x-theme files, which generates a second entry for pluma in this list.
3. For some reasons gedit and libreofficewriter are the default for application/x-theme mimetype, and i see them in menu and can't remove them.
4. left-click on the file produced this ugly error message now.
bildschirmfoto zu 2016-12-18_20-12-53


I tried to add the mimetype to pluma desktop file and expected to see pluma as default app like gedit and libreofficewriter in right-click menu, but no success.
And i am still wondering how the system recognizes that gedit and libreofficewriter are the default apps for this file.
Of course playing with ~.config/mimeapps.list gives me no success.
So, i am not sure about this fix because of point 1 and 4.
Well the mimetype stuff isn't related to caja but the error popup is really ugly.

Btw. i am still wondering why using the 2 origin commits from nautilus/nemo doesn't work with caja.

lukefromdc:
Somewhere in the properties dialog must be the code that shows or does not show the "open with" tab. When I get a bit more time I could dig for that, it would allow users to reset the applications. As for the error message, you get one either way, with or without this change. Without it is its "could not display(the selected item)" and "the location is not a folder" indicating that the index.theme files are being mistaken for directories. 

The "open with" tab in properties does appear for index.theme files in Nemo, BTW. Fixing the default would then be possible, and might stop the error message. Next needed would be a way to reset that default by default on a new Caja installs where the user has not set a custom "open with" default setting for index.theme files

lukefromdc:
https://github.com/mate-desktop/caja/commit/f776cb7163e4d033c7bdb13d56284266bca5cc0f
shows the "open with" tab in the properties window for these files. Unfortunately the selections made are saved but the selected default is not actually applied.

Thus selecting "open" from the right-click menu or left-clicking on the file still generates the same error message.  You can add applications or remove inappropriate ones but that's it until I found out where this gets applied and fix it there.

lukefromdc:
The default application for an index.theme file comes up as "pluma" yet neither this nor any user-set default is applied. My guess is something is defining index.theme files as desktop launchers and overriding the default.

raveit65:
With this properties tab i can remove other applications like gedit or libreoffice as default application. This gives me this entry in ~~.config/mimeapps.list
```
[Removed Associations]
application/x-theme=org.gnome.gedit.desktop;libreoffice-writer.desktop;
```
So the mimetype is clear, but adding application/x-theme=pluma.desktop under [Added Associations] or [Default Applications] does not have any effect.
So it looks like that something with pluma is wrong.
Btw. I found out that all expressions in our desktop files with ie. 'X-MATE-FullName=Pluma Text Editor' do not have any effect.
Changing this to 'X-GNOME-FullName=Pluma Text Editor' gives me a workking result in properties tab.

lukefromdc:
Resetting the default application to something other than Pluma does not change the error message. Therefore, I think index.theme files are being treated as desktop launchers and the set default is being overridden

lukefromdc:
I finally found the GNOME/Nautilus commit that makes this work right

https://github.com/GNOME/nautilus/commit/b241fd28913e87fe01419e4922259949a11c973f
file: don't consider all desktop files nautilus links
Only those which match application/x-desktop directly.

Now index.theme files show their proper icon as in Nemo, open in Pluma by default when clicked on, and show both the "open" and "open with" right-click menus as well as the preferences/Open With tab. 

This matches Nemo's behavior for both index.theme and .desktop(launcher) files. If we want to only change behavior of index.theme files we could revert the other two commits, but in my workflow it is more common to have to edit a .desktop file than it is an index.theme file so I would prefer to keep them. At any rate, the whole thing is based on Nautilus commits so keeping them also reduces the degree of divergence between Nautilus and Caja

raveit65:
Nice finder and great job :-)
It works fine now and application/x-desktop files displays now like they are in browser, as a index.theme file.

raveit65:
Btw. i am fine with merging this state ;-)

lukefromdc:
Rebased and merged via
https://github.com/mate-desktop/caja/commit/5432a415dadbd8db34081037645e0c38457900ec
Show "open with" menu for desktop and theme file

https://github.com/mate-desktop/caja/commit/a4f26d060486684e8d37e922ab2fbc4dcc02a85e
Show "open with" tab in properties window when "open with" menu avail… 

https://github.com/mate-desktop/caja/commit/91dcd8a19665b6738f8304db127f45b503e00ce3
file: don't consider all desktop files caja links

raveit65:
Just cherry-picked those commits to 1.16, as it is a bugfix for a long outstanding issue.

PR #696, opened 2016-12-30
Fix cursor on single-click navigation mode.

torto9:
On this mode, after opening a folder, the hand cursor was used until the user moved the cursor over a file or folder.
Based on Nautilus.

### DIFF
--- a/libcaja-private/caja-icon-canvas-item.c
+++ b/libcaja-private/caja-icon-canvas-item.c
@@ -149,6 +149,8 @@ struct CajaIconCanvasItemDetails
     EelIRect bounds_cache_for_layout;
     EelIRect bounds_cache_for_entire_item;
 
+    GdkWindow *cursor_window;
+
     /* Accessibility bits */
     GailTextUtil *text_util;
 };
@@ -254,6 +256,12 @@ caja_icon_canvas_item_finalize (GObject *object)
     g_assert (CAJA_IS_ICON_CANVAS_ITEM (object));
 
     details = CAJA_ICON_CANVAS_ITEM (object)->details;
+
+    if (details->cursor_window != NULL)
+    {
+        gdk_window_set_cursor (details->cursor_window, NULL);
+        g_object_unref (details->cursor_window);
+    }
 
     if (details->pixbuf != NULL)
     {
@@ -2021,8 +2029,10 @@ caja_icon_canvas_item_event (EelCanvasItem *item, GdkEvent *event)
 {
     CajaIconCanvasItem *icon_item;
     GdkCursor *cursor;
+    GdkWindow *cursor_window;
 
     icon_item = CAJA_ICON_CANVAS_ITEM (item);
+    cursor_window = ((GdkEventAny *) event)->window;
 
     switch (event->type)
     {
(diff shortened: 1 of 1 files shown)


lukefromdc:
I was able to verify the described issue, and that this fix works. Waiting for input by others on the team before merging

monsta:
I confirm the issue and the fix, but... it seems like the issue is reversed now. :smile:
Now the cursor always becomes a pointer after opening some folder - even if there's another subfolder inside right under the cursor. You need to move it a bit (still staying over the subfolder) so it will change into hand.

monsta:
Upstream commit for reference:
https://git.gnome.org/browse/nautilus/commit/?id=2644c6ef4e7839e976d4861d6c86e16e34d46a13

raveit65:
Ok, i see the same new issue with this PR.
Well, it is very minor but i don't see that with nautilus-3.20.x
Anyway, we can merge it before we lost years to find the fix for the fix, if all agree.

lukefromdc:
What GTK version did you test with?

I just tested this in Nautilus, the reversed issue does not exist there with Nautlus and GTK 3.22. Also in Caja the pointer stayed a hand on my system if opening a folder put the cursor on another folder inside. Using GTK 3.22.5 plus recent commits to suppress some warnings.

lukefromdc:
I'm OK with merging as-is if others are

raveit65:
i just tested it with f24 and gtk+-3.20.

monsta:
I have GTK+ 3.18.9 on this machine and a VM with 3.22.5 (Debian Testing), both have the new issue. It's actually reproducible not only with subfolders, but with files inside as well.

Ok, the new issue is easier to get rid of (just a bit of mouse moving), so I'm fine with merging.
I suggest adding the upstream commit link to the commit message, just in case.

torto9:
I hadn't noticed the new problem...
After testing old Nautilus versions, I found that the issue was fixed by commit [https://git.gnome.org/browse/nautilus/commit/?id=d486af277f491ff5cfcfa254ebce618cd41a3edb](url)
To be honest, I have no idea why that fixes it... :)

lukefromdc:
Merged with
https://github.com/mate-desktop/caja/commit/f604ba59b9a1e45c1fc45c0e94869e416f5a94b0
and the GNOME/nautilus commit added to the extended description. I will now try out that second GNOME commit. We may need it anyway, because MATE can be used (or should be usable) with a touchscreen the same way people put Windows 8/10 back in desktop mode but still use their touchscreens.

torto9:
Thanks!

lukefromdc:
I was able to duplicate the hand cursor not holding in when opening a folder brought the cursor over a new folder or file, turns out I was still double-clicking most of the time, turning the cursor back to an arrow. That allowed me to apply and test https://git.gnome.org/browse/nautilus/commit/?id=d486af277f491ff5cfcfa254ebce618cd41a3edb and it worked, so committed it to Caja as
https://github.com/mate-desktop/caja/commit/dbce9dbdf16b7d8afc8e0c1b6166b26221e1466f

raveit65:
Yeap, that fixes the second issue.
Thanks guys.

monsta:
Good, I confirm it's fixed now. Thanks :smile:

Issue #649, opened 2016-10-05
[GTK+ >= 3.20] caja crash when changing from trash in list view to other directory in list view


viktorjk:
OS: Debian unstable (with GTK+ 3.22.1)
Caja version: 1.16.0

(I hope that I translated the expressions correctly to english.)
I could reproduce this with a new account and after a fresh install of caja. When I open the trash in list view, caja crashes when I go to another directory which was set to be displayed in list view. This is independent of the way to change the directory. I saw this for an empty trash and for one file in the trash.

The syslog entry for one crash is:
caja[8724]: gdk_window_get_window_type: assertion 'GDK_IS_WINDOW (window)' failed
caja[8724]: gtk_widget_set_parent_window: assertion 'GTK_IS_WIDGET (widget)' failed
caja[8724]: gtk_widget_set_parent: assertion 'GTK_IS_WIDGET (widget)' failed
kernel: [56828.208234] caja[8724]: segfault at 18 ip 00007f213cf76390 sp 00007ffd0d6fb5a8 error 4 in libgtk-3.so.0.2200.1[7f213cbec000+6f7000]
caja[9379]: Allocating size to GtkGrid 0x56258c28e140 without calling gtk_widget_get_preferred_width/height(). How does the code know the size to allocate?
caja[9379]: Allocating size to GtkGrid 0x56258c28e140 without calling gtk_widget_get_preferred_width/height(). How does the code know the size to allocate?
caja[9379]: Allocating size to GtkGrid 0x56258c28e140 without calling gtk_widget_get_preferred_width/height(). How does the code know the size to allocate?

See also https://bugs.debian.org/834943 (Caja 1.14.1 and GTK+ 3.20.7)

If I should give you more information about my system, just tell me. If you want me to debug this, please give me instructions or send a link to a site with instructions.
##

lukefromdc:
Confirmed here. The grid warnings should have nothing to do with it,
I am using  a branch that applies a Nemo/Nauitlus commit that fixes
those. I get only this segfault on the trash/other directory/both in list
mode transition:

Oct  5 19:08:15 ubuntu kernel: [ 6183.996530] caja[4921]: segfault at 18 ip 00007f5352d4b270 sp 00007ffc84fff2d8 error 4 in libgtk-3.so.0.2200.1[7f53529d5000+6e3000]

"Ubuntu" is just the machine name to obfuscate tracking by network
providers, OS is Debian Unstable with GTK 3.22 and a lot of locally
built packages.

[...]

PR #705, opened 2017-01-06
show/hide columns instead of appending/removing them

monsta:
fixes https://github.com/mate-desktop/caja/issues/649

adapted from:
https://git.gnome.org/browse/nautilus/commit/?id=e19f7cb3fec1e6c884d1741c3c00916c65ab035f

### DIFF
--- a/src/file-manager/fm-list-view.c
+++ b/src/file-manager/fm-list-view.c
@@ -1544,27 +1544,22 @@ apply_columns_settings (FMListView *list_view,
 
     view_columns = g_list_reverse (view_columns);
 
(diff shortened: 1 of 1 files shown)


lukefromdc:
Just did a test build, it worked and stopped the "trash crash." Reverting to master restored the crash, then reinstalling this version stopped it again. No adverse effects observed in a quick test

monsta:
Good, so no crashes on switching to list view like you mentioned previously in #649?

lukefromdc:
No problem switching to list view or from it, no trash issue.

Issue #245, opened 2014-01-27
icons are not spoken by Orca


the-kyle:
Running Mate 1.6.2 and Orca 3.10, the icons on the desktop and in the file manager are not being spoken by the Orca screen reader. Also, Orca is occasionally announcing that the list of icons contains 0 items, even when files are present, such as in the home directory or in the computer view. My guess is that the text of an icon as it is focused isn't getting passed to at-spi, which is the interface between the graphical toolkit and the Orca screen reader.
##

coffeeking:
this bug no longer exists in the gtk2 version of mate and hasn't since 1.8. However, it does exist in the gtk3 version of mate. This bug has taken on a new urgency for us at the sonar project because  the gtk2 version of mate will eventually be pulled outo f the manjaro repositories, leaving the gtk3 version. I'd suggest looking at nautilus for a fix, since their icon view is accessible. You could also look at nemo, mint's fork of nautilus if you'd rather. I'll do anything I can to help.

monsta:
@coffeeking: it seems to me that it's also discussed in https://github.com/mate-desktop/caja/issues/479 and/or https://github.com/mate-desktop/caja/issues/590. I guess we should close a report or two...

monsta:
> Also, Orca is occasionally announcing that the list of icons contains 0 items, even when files are present, such as in the home directory or in the computer view.

This part should be fixed now with https://github.com/mate-desktop/caja/commit/68bbfd2f66682f7fcbbf4791ac685f0425a4bef5 (see https://github.com/mate-desktop/caja/pull/706 for some details).

flexiondotorg:
I'll close this, please request a re-open if it is not completely fixed :-)

PR #652, opened 2016-10-08
Fix accessible object inherritance when used with GTK >= 3.2

TheMuso:
Also disable eel accessibility code that is not needed.

Fixes https://github.com/mate-desktop/caja/issues/590

PR #706, opened 2017-01-06
a11y: fix accessibility implementation for CajaIconCanvasItemAccessible

cwendling:
Fix CajaIconCanvasItemAccessible inheritance to properly inherit from
EelCanvasItemAccessible.  This fixes the ATK state machinery in
CajaIconCanvasItemAccessible, and adds AtkComponent support which
provides several useful features.

Part of https://github.com/mate-desktop/caja/issues/245 (the "0 items" part)

Based off https://git.gnome.org/browse/nautilus/commit/?id=6c5baeb7626eda6629fc6642c9eb513ef8bc5c8e
See https://bugzilla.gnome.org/show_bug.cgi?id=677509

### DIFF
--- a/eel/eel-canvas.c
+++ b/eel/eel-canvas.c
@@ -3992,21 +3992,6 @@ eel_canvas_item_accessible_ref_state_set (AtkObject *accessible)
     return state_set;
 }
 
-static GType eel_canvas_item_accessible_get_type (void);
-
-typedef struct _EelCanvasItemAccessible EelCanvasItemAccessible;
-typedef struct _EelCanvasItemAccessibleClass EelCanvasItemAccessibleClass;
-
-struct _EelCanvasItemAccessible
-{
-    GtkAccessible parent;
-};
-
-struct _EelCanvasItemAccessibleClass
-{
-    GtkAccessibleClass parent_class;
-};
-
 G_DEFINE_TYPE_WITH_CODE (EelCanvasItemAccessible,
                          eel_canvas_item_accessible,
                          ATK_TYPE_GOBJECT_ACCESSIBLE,
--- a/eel/eel-canvas.h
+++ b/eel/eel-canvas.h
@@ -558,6 +558,21 @@ extern "C" {
         GtkContainerAccessibleClass parent_class;
     };
 
+    GType eel_canvas_item_accessible_get_type (void);
+
+    typedef struct _EelCanvasItemAccessible EelCanvasItemAccessible;
+    typedef struct _EelCanvasItemAccessibleClass EelCanvasItemAccessibleClass;
+
+    struct _EelCanvasItemAccessible
+    {
+        GtkAccessible parent;
+    };
+
+    struct _EelCanvasItemAccessibleClass
+    {
+        GtkAccessibleClass parent_class;
+    };
+
 #ifdef __cplusplus
 }
 #endif
--- a/libcaja-private/caja-icon-canvas-item.c
+++ b/libcaja-private/caja-icon-canvas-item.c
@@ -3436,16 +3436,16 @@ caja_icon_canvas_item_accessible_text_interface_init (AtkTextIface *iface)
 }
 
 typedef struct {
-	AtkGObjectAccessible parent;
+	EelCanvasItemAccessible parent;
 } CajaIconCanvasItemAccessible;
 
 typedef struct {
-	AtkGObjectAccessibleClass parent_class;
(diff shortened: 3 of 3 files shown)


cwendling:
You can see in the linked GNOME report the effect it has.  Additionally, you can easily see with Orca or Accerciser:

# With Orca:

1. Launch orca
2. Open a non-empty directory in icon view (or go on the desktop)

## Expected result
Speaks *"\<name\> pane, Icon View layered pane"*

## Actual results (before this patch)
Speaks *"\<name\> pane; Icon View layered pane 0 items"*

# With Accerciser

1. Open Accerciser
2. Browse down a Caja window to find the Icon View
  As it's a bit deep, you might want to add the bookmark `1,0,1,1,0,0,1,0,0,1,0,0` for `caja` (the first element will be the window index inside the `caja` app, so it might not be 1, depending on whether you have the desktop enabled and other Caja windows)
3. Select any of the child icons
4. Switch to the *Interface browser* tab on the right

## Expected result
Under *Accessible*, most, if not all, icons should have *States* *visible* and *showing*.
Under *Component* you should see positions and sizes.

## Actual result (before this patch)
Under *Accessible*, the *States* is empty.
There's nothing under *Component*.

cwendling:
BTW, before the patch you can see these errors when accessing the accessible objects:

```console
invalid cast from 'CajaIconCanvasItemAccessible' to 'AtkComponent'
atk_component_get_position: assertion 'ATK_IS_COMPONENT (component)' failed
```

lukefromdc:
Asking for reviews prior to merge, but this worked in my tests with Accerciser.  No adverse results detected so far.

Does this need also to be tested by someone with a working Orca install, or do Accerciser results reliably track with behavior with Orca?

monsta:
Oh, it looks like follow-up to https://github.com/mate-desktop/caja/pull/652. However, this time we're GTK+3 only, so backporting this to 1.16 branch will require some more effort...

monsta:
Oooh... couldn't get Orca to speak UI elements in English, it always tries to use my language (Russian), so it took me a while to tweak some options in setup dialog before I could understand what it's trying to say... :smile:

monsta:
BTW, I didn't see the mentioned invalid cast warning with or without the patch. I've tested it with GTK+ 3.18.9, maybe it matters.

lukefromdc:
Rebased (after https://github.com/mate-desktop/caja/commit/81c03e738fd933702738264c75b0a4a0a72495b9) and merged 
https://github.com/mate-desktop/caja/commit/68bbfd2f66682f7fcbbf4791ac685f0425a4bef5

cwendling:
@lukefromdc 

> Does this need also to be tested by someone with a working Orca install, or do Accerciser results reliably track with behavior with Orca?

You probably need to test with Orca if you wanna see its behavior.  Accerciser and Orca both use the same information set, but what they do with it is totally different, and I don't think it's obvious that making icons properly report their showing state would fix that in Orca.  So if you wanna make sure it does as advertized, better check with it.

@monsta 
> However, this time we're GTK+3 only, so backporting this to 1.16 branch will require some more effort...

I need this backported on Caja 1.8/GTK2 for a little while, so I can submit the fix I come up with.  Should be fairly easy, even though the actual code will indeed be different, but the idea likely the same.
If so, against what should I create the PR?  The 1.16 branch?

> BTW, I didn't see the mentioned invalid cast warning with or without the patch. I've tested it with GTK+ 3.18.9, maybe it matters.

Not sure…  I only saw those when trying to browse the objects with Accerciser.  Under non-AT usage, those objects aren't created anyway, let alone used.

---
BTW, thanks for the review & merge!

cwendling:
> I need this backported on Caja 1.8/GTK2 for a little while, so I can submit the fix I come up with. Should be fairly easy, even though the actual code will indeed be different, but the idea likely the same.
If so, against what should I create the PR? The 1.16 branch?

I have a working Caja 1.8/GTK2 fix, that's likely relevant for 1.16 too.  If you tell me where to submit it I'd be happy to do so (with required modification for 1.16, if any).

monsta:
Yes, please create a PR for 1.16 branch, and we'll test it.

I didn't know some fix is needed for GTK+2 as well; I thought the inheritance has to be fixed only for GTK+3 code, same as in https://github.com/mate-desktop/caja/pull/652 (with ```GTK_CHECK_VERSION``` macros to keep GTK+2 code as is).

PR #707, opened 2017-01-07
force X11

leigh123linux:
This commit enables caja to draw desktop icons if used in a wayland session.

### DIFF
--- a/src/caja-main.c
+++ b/src/caja-main.c
@@ -266,6 +266,8 @@ main (int argc, char *argv[])
 
 	g_set_prgname ("caja");
 
+	gdk_set_allowed_backends ("x11");
+
 	if (g_file_test (DATADIR "/applications/caja.desktop", G_FILE_TEST_EXISTS)) {
 		egg_set_desktop_file (DATADIR "/applications/caja.desktop");
 	}


lukefromdc:
In a quick test, I found no adverse effects in an X11 session but it actually broke opening a Caja navigation window in a Weston session. The resulting error is 
`
Could not parse arguments: Cannot open display`

Without this commit a navigation window will open just fine in Weston, but launching with --force-desktop (Caja 1.17.2) will cause a segfault. With this commit the error above results with or without --force-desktop. Wonder if this is specific to Weston in some way?

raveit65:
I just tested it under fedora rawhide in gnome-wayland session. It works well.
A backport to 1.16 branch works well in f25 gnome-wayland session.
Without caja and other mate apps (ie. mate-terminal) crashes under wayland.
I am in doubt that anyone other can test this in another distro, so i will merge it and backport to 1.16.

raveit65:
Oh , i didn't saw your comment. But --force-desktop works here in gnome-wayland without any problem.
Looks like an issue with weston?

lukefromdc:
Could well be a Weston issue, to get to a GNOME waylaand session I have to temporarily uninstall lightdm so GDM3 can run. Anyway, no problems in X11 so go ahead with the merge

leigh123linux:
@lukefromdc 

Looks like weston doesn't set any display env judging by the error

lukefromdc:
Well that figures and is not a bug on this end.  What nautilus now does, BTW, is to force X11 for nautilus-desktop but I would assume they are actually using wayland for navigation windows directly, since GTK3 supports that. Doing this requires the split-out desktop binary, and is considered temporary until they come up with something better, No reason not to use X11 for all the Caja windows so long as it is needed for any of them(desktop).

lukefromdc:
I can confirm this works in a gnome-wayland session, though I had to launch it with alt-F2. This was also true of Nemo and even Nautilus, all of which generate the same error as above on attempt to launch from a terminal in gnome-wayland on my boxes.

My terminals have had environment mods in the past to work around various issues, so this may be local to my machines and in any event is a terminal issue, not a Caja issue so I will merge this.

lukefromdc:
Weston does some odd things as a window manager, with Caja and Nemo desktops coming up in transparent(icon container is transparent with compositing)  decorated windows on top of (and blocking interaction with until moved) any window previously drawn. Nautilus-desktop comes up undecorated but still in a non-fullscreen window. Anyway, Caja is working like the others and none of them have this issue in a gnome-wayland session

PR #711, opened 2017-01-10
a11y: fix accessibility implementation for CajaIconCanvasItemAccessible

cwendling:
Fix CajaIconCanvasItemAccessible inheritance to properly inherit from EelCanvasItemAccessible.  This fixes the ATK state machinery in CajaIconCanvasItemAccessible, and adds AtkComponent support which provides several useful features.

See also https://bugzilla.gnome.org/show_bug.cgi?id=677509 and https://github.com/mate-desktop/caja/issues/706.
Partly based off https://git.gnome.org/browse/nautilus/commit/?id=6c5baeb7626eda6629fc6642c9eb513ef8bc5c8e

### DIFF
--- a/eel/eel-canvas.c
+++ b/eel/eel-canvas.c
@@ -4297,21 +4297,6 @@ eel_canvas_item_accessible_ref_state_set (AtkObject *accessible)
 }
 
 #if GTK_CHECK_VERSION(3, 0, 0)
-static GType eel_canvas_item_accessible_get_type (void);
-
-typedef struct _EelCanvasItemAccessible EelCanvasItemAccessible;
-typedef struct _EelCanvasItemAccessibleClass EelCanvasItemAccessibleClass;
-
-struct _EelCanvasItemAccessible
-{
-    GtkAccessible parent;
-};
-
-struct _EelCanvasItemAccessibleClass
-{
-    GtkAccessibleClass parent_class;
-};
-
 G_DEFINE_TYPE_WITH_CODE (EelCanvasItemAccessible,
                          eel_canvas_item_accessible,
                          ATK_TYPE_GOBJECT_ACCESSIBLE,
@@ -4370,6 +4355,12 @@ eel_canvas_item_accessible_get_type (void)
         if (!parent_atk_type)
         {
             return G_TYPE_INVALID;
+        }
+        else if (parent_atk_type == ATK_TYPE_NO_OP_OBJECT)
+        {
+            /* use at least AtkGObjectAccessible as basis if nothing else
+             * is advertized in the registry */
+            parent_atk_type = ATK_TYPE_GOBJECT_ACCESSIBLE;
         }
         g_type_query (parent_atk_type, &query);
         tinfo.class_init = (GClassInitFunc) eel_canvas_item_accessible_class_init;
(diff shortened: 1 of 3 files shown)


eel/eel-canvas.c
```diff
@@ -4371,6 +4356,12 @@ eel_canvas_item_accessible_get_type (void)
         {
             return G_TYPE_INVALID;
         }
+        else if (parent_atk_type == ATK_TYPE_NO_OP_OBJECT)
+        {
+            /* use at least AtkGObjectAccessible as basis if nothing else
+             * is advertized in the registry */
+            parent_atk_type = ATK_TYPE_GOBJECT_ACCESSIBLE;
```
cwendling: This is required because ATK doesn't register an accessible type for `G_TYPE_INITIALLY_UNOWNED` in the registry, so we get an `ATK_TYPE_NO_OP_OBJECT`, yet we want to derive from something sensible, so at least `ATK_TYPE_GOBJECT_ACCESSIBLE` (like the GTK3 version).
Without this, not only some features are implemented in a fake way, but `CajaIconCanvasItemAccessible` relies on being a type derived from `ATK_TYPE_GOBJECT_ACCESSIBLE` using some of its features (like `atk_gobject_accessible_get_object()`).

cwendling:
Wait, actually the fix could be simpler as I see `EelCanvasItemAccessible` class and object should actually have members of type `AtkGObjectAccessible` (they derive from `ATK_TYPE_GOBJECT_ACCESSIBLE`) which is public in GTK2 too, not `GtkAccessible`.

cwendling:
See #712 for an alternative version.  Same end result, I let you choose which one you prefer.

flexiondotorg:
@lukefromdc I've seen this commit, is this PR done with now?

b4n:
@flexiondotorg No, PRs #711 and #712 are for the 1.16 branch and include the fix for GTK2 too, the commit you mentioned comes from #706, which is the GTK3-only fix that made it to master.

monsta:
@cwendling: so it was done somewhat incorrectly in https://github.com/mate-desktop/caja/pull/652? The code in question was added there (at [this point](https://github.com/mate-desktop/caja/commit/a8470c69525fd6b0c2c94755298753f90b813831#diff-edecdcd42dd3e3885b0e3276c5fe89c7R4299)). If your new implementation in https://github.com/mate-desktop/caja/pull/712 is more correct than that, I would certainly choose that PR over this one. It makes code simpler :smile:

cwendling:
@monsta yes, it is.  I guess the split code for GTK2/3 was due to [the `EelCanvasItemAccessible` structure erroneously inheriting from `GtkAccessible`](https://github.com/mate-desktop/caja/commit/a8470c69525fd6b0c2c94755298753f90b813831#diff-edecdcd42dd3e3885b0e3276c5fe89c7R4305), which is not public in GTK2.  But anyway, as you can see in [the `G_DEFINE_TYPE()` call](https://github.com/mate-desktop/caja/commit/a8470c69525fd6b0c2c94755298753f90b813831#diff-edecdcd42dd3e3885b0e3276c5fe89c7R4315), it should actually inherit from `AtkGObjectAccessible`, which is public.

So yeah, #711 and #712 should be perfectly equivalent.

cwendling:
And yeah, more correct as #712 gives the correct structure member to `EelCanvasItemAccessible`.

monsta:
Yes, they seem to be the same. I get the same issue... Orca doesn't say "0 elements" inside an empty folder.

monsta:
Closing as #712 is merged.

PR #712, opened 2017-01-10
a11y: fix accessibility implementation for CajaIconCanvasItemAccessible (v2)

cwendling:
Fix CajaIconCanvasItemAccessible inheritance to properly inherit from EelCanvasItemAccessible.  This fixes the ATK state machinery in CajaIconCanvasItemAccessible, and adds AtkComponent support which provides several useful features.

See also https://bugzilla.gnome.org/show_bug.cgi?id=677509 and https://github.com/mate-desktop/caja/issues/706.
Partly based off https://git.gnome.org/browse/nautilus/commit/?id=6c5baeb7626eda6629fc6642c9eb513ef8bc5c8e

---
Alternative simpler version of #711.

### DIFF
--- a/eel/eel-canvas.c
+++ b/eel/eel-canvas.c
@@ -4296,22 +4296,6 @@ eel_canvas_item_accessible_ref_state_set (AtkObject *accessible)
     return state_set;
 }
 
-#if GTK_CHECK_VERSION(3, 0, 0)
-static GType eel_canvas_item_accessible_get_type (void);
-
-typedef struct _EelCanvasItemAccessible EelCanvasItemAccessible;
-typedef struct _EelCanvasItemAccessibleClass EelCanvasItemAccessibleClass;
-
-struct _EelCanvasItemAccessible
-{
-    GtkAccessible parent;
-};
-
-struct _EelCanvasItemAccessibleClass
-{
-    GtkAccessibleClass parent_class;
-};
-
 G_DEFINE_TYPE_WITH_CODE (EelCanvasItemAccessible,
                          eel_canvas_item_accessible,
                          ATK_TYPE_GOBJECT_ACCESSIBLE,
@@ -4332,61 +4316,7 @@ static void
 eel_canvas_item_accessible_init (EelCanvasItemAccessible *accessible)
 {
 }
-#else
-static void
-eel_canvas_item_accessible_class_init (AtkObjectClass *klass)
-{
-    accessible_item_parent_class = g_type_class_peek_parent (klass);
 
-    klass->initialize = eel_canvas_item_accessible_initialize;
-    klass->ref_state_set = eel_canvas_item_accessible_ref_state_set;
-}
-
-static GType
-eel_canvas_item_accessible_get_type (void)
-{
-    static GType type = 0;
-
-    if (!type)
-    {
-        static const GInterfaceInfo atk_component_info =
-        {
-            (GInterfaceInitFunc) eel_canvas_item_accessible_component_interface_init,
-            (GInterfaceFinalizeFunc) NULL,
-            NULL
-        };
-        AtkObjectFactory *factory;
-        GType parent_atk_type;
-        GTypeQuery query;
-        GTypeInfo tinfo = { 0 };
-
-        factory = atk_registry_get_factory (atk_get_default_registry(),
-                                            G_TYPE_INITIALLY_UNOWNED);
-        if (!factory)
-        {
-            return G_TYPE_INVALID;
-        }
-        parent_atk_type = atk_object_factory_get_accessible_type (factory);
-        if (!parent_atk_type)
-        {
-            return G_TYPE_INVALID;
-        }
-        g_type_query (parent_atk_type, &query);
-        tinfo.class_init = (GClassInitFunc) eel_canvas_item_accessible_class_init;
-        tinfo.class_size = query.class_size;
-        tinfo.instance_size = query.instance_size;
-        type = g_type_register_static (parent_atk_type,
-                                       "EelCanvasItemAccessibility",
-                                       &tinfo, 0);
-
-        g_type_add_interface_static (type, ATK_TYPE_COMPONENT,
-                                     &atk_component_info);
-
-    }
-
-    return type;
-}
-#endif
 static AtkObject *
 eel_canvas_item_accessible_create (GObject *for_object)
 {
--- a/eel/eel-canvas.h
+++ b/eel/eel-canvas.h
@@ -575,6 +575,21 @@ extern "C" {
     };
 #endif
 
+    GType eel_canvas_item_accessible_get_type (void);
+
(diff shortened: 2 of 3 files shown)


lukefromdc:
This just built and ran fine over GTK 3.22, no idea how it would behave with GTK2 but I assume this is a GTK3 bugfix?  Same behavior in Accerciser as the most recent commits to master, looks good there. Last time I tried to install Orca I could not get it to work.

cwendling:
> This just built and ran fine over GTK 3.22, no idea how it would behave with GTK2 but I assume this is a GTK3 bugfix?

No, it's a fix for both GTK2 and GTK3.  On my end both worked fine.

> Same behavior in Accerciser as the most recent commits to master, looks good there. Last time I tried to install Orca I could not get it to work.

I think that now on #706 someone checked the Orca behavior, seeing Accerciser is behaving the same is probably enough.  But needs testing on GTK2 too if you wanna be thorough :)

lukefromdc:
I'll have to leave the GTK2 test to someone who has a GTK2 environment set up to do it in, as I do everything live on bare metal

flexiondotorg:
@lukefromdc I've seen this commit, is this PR done with now?

  * https://github.com/mate-desktop/caja/commit/68bbfd2f66682f7fcbbf4791ac685f0425a4bef5

lukefromdc:
This is for 1.16, which needs to work in GTK2. I have been waiting for someone to test that prior to merge. GTK3 works fine.  I checked the actual code changes, many of them affect all GTK versions and thus require a GTK2 test with accessability tested unless accessability in GTK2 is broken as-is.

monsta:
I also only have GTK+3 builds on baremetal now. I will try to make Orca work in a VM with GTK+2 build.
Are there any visible (or rather, audible) changes that I could notice immediately after applying this PR?

cwendling:
> Are there any visible (or rather, audible) changes that I could notice immediately after applying this PR?

Yes: when you get to an Icon View with some elements, without the PR you'll hear "0 items"; with it you won't.  See #706 (this PR here is a backport of #706, it should behave the same).

monsta:
Right, it was a stupid question, for some reason I thought the result would differ here :smile:

monsta:
Orca in Ubuntu 16.04 is completely buggy. Sometimes it works, but most of the time it randomly refuses to speak anything, or simply quits by itself (including setup dialog). In a VM, it seems to be even worse. Sometimes it doesn't like a particular speak engine and shuts up completely until I reboot the system.

alexarnaud:
Le 20/01/2017 à 18:48, monsta a écrit :
> Orca in Ubuntu 16.04 is completely buggy. Sometimes it works, but most
> of the time it randomly refuses to speak anything, or simply quits by
> itself (including setup dialog). In a VM, it seems to be even worse.
> Sometimes it doesn't like a particular speak engine and shuts up
> completely until I reboot the system.

You're right. You could use Debian testing to have a more stable 
environment to test a11y stuff.

Best regards.
-- 
Alex ARNAUD

monsta:
Ok, so without this PR it says "0 elements" in every folder, including non-empty ones, but with this PR it stops says "0 elements" for empty folders... In GTK+3 version it said it for empty folders just fine.

monsta:
Yes, I use Debian Testing as well, but it already has GTK+3 build of MATE. My only GTK+2 system with MATE 1.16 is a VM with Mint 18.1 (based on Ubuntu 16.04).

cwendling:
> Ok, so without this PR it says "0 elements" in every folder, including non-empty ones, but with this PR it stops saying "0 elements" for **empty** folders... In GTK+3 version it said it for empty folders just fine.

Is that so?  I just re-checked, and for me it still says "0 items" (well, OK, "0 éléments" as I'm also in french locale ;)) for empty folders under GTK2 just fine; and I can't see any reason why it would think there are elements if they are not.

If you revert this patch alone, does Orca say "0 items" for empty folders again?

monsta:
I have Russian locale, it's "0 элементов" which literally translates to "0 elements" (but can be translated as "0 items" as well). Interesting that Caja status bar uses yet another word, it says "0 объектов" which means "0 objects". :smile:

Yes, I rebuilt current 1.16 branch again to test it, then it started saying "0 elements" inside every folder (empty or non-empty).

cwendling:
> I have Russian locale, it's "0 элементов" which literally translates to
> "0 elements" (but can be translated as "0 items" as well).

Oh, I see :)

> Yes, I rebuilt current 1.16 branch again to test it, then it started
> saying "0 elements" inside every folder (empty or non-empty).

Damn, that's weird.  Could you tell me exactly what your setup is?  I'm
testing on Debian Sid, and Orca 3.22.2.   On my end, to test I created
an empty folder with `mkdir /tmp/empty`; then when to `/tmp/empty` and
`/tmp`, and it worked as expected.

Maybe you got a different Orca version?  Yet I doubt it should do that,
but who knows.

BTW, I backported the fix to 1.8, and it worked for me and @alexarnaud
there too (and I think with both Orca 3.16 and master, but that I'm not
totally sure right now).

monsta:
It's Mint 18.1, based on Ubuntu 16.04. Mint has most packages right from Ubuntu repos. Orca is 3.18.2-1ubuntu3 and - just in case - GTK+2 is 2.24.30-1ubuntu1.

monsta:
I would try it in Debian as well, but Testing and Sid have GTK+3 build by default, and I don't really want to rebuild everything with GTK+2 there...

cwendling:
@monsta I can't reproduce your issue under a Mint 18.1 VM.  For me, without the patch it always says "0 items", and with it it says it only in empty directories, but it *does* say it there.

What I did is:
1. run a Mint 18.1 ISO in a VM
2. install all the stuff
3. build Caja from the 1.16 branch
4. test and confirm the "0 items" in non-empty directories (so the branch has the problem)
5. switch to my branch (this PR)
6. test and confirm it only says "0 items" in empty directories

Again, for checking empty directories I created a new directory with `mkdir /empty` and went there with `caja /tmp/empty`.

But I see something similar to what you describe, but only when I navigate clicking/activating an icon in a view, thus changing the location.  If I use the location bar it works fine.  BTW, when it doesn't say "0 items", it's not saying "layered pane" either, so I guess it's just not announcing as much context for some reason.  Thus, I believe it's a separate issue where the content is either not properly refreshed for the AT layer, or Orca is not reacting to some changes the same as with a full view.
And I reproduce this both with an without my patch.

monsta:
Ok, now I see it might be something wrong with refreshing. I always clicked on a folder to enter it. Now I tried to use the location bar, but still couldn't make it say "0 elements" when using the location bar.

However, I found two methods to make it say that:

- open a folder directly from the terminal, like ```caja ~/empty```
- open Caja as usual, enter the empty folder, right-click on the empty space somewhere, invoke the folder properties dialog, then close the dialog

In both cases Orca said about "0 elements" in the end. So it seems to work, I will merge it, and we can dig into the refresh issue later.

PR #723, opened 2017-01-26
Add an option for asking confirmation before moving files to the trash

### LINKED ISSUE #632: [FR] An option to enable confirmation on moving (deleting) files to trash
Yes, we know... moving a file is not deleting it... ok

But you know too, that mate is used by several kind of people, even kids... so, it is not nice having to check trash every day to see if something was deleted by mistake.
In fact, anyone can hit delete key by mistake in caja having focus, and delete whatever it was there without too much notice... and pushing it for a few seconds will delete (move) a bunch of things. Because files are being selected on every moving, if you do not release the key you will be movingtotrash(deleting) all the files and directories on that level and below

If you find that behavior, moving to trash without confirmation, a nice feature ok, that is great. But let the other people to decide if they want to use it or not...
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/37670913-fr-an-option-to-enable-confirmation-on-moving-deleting-files-to-trash?utm_campaign=plugin&utm_content=tracker%2F651521&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F651521&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cwendling:
Add an option to show a confirmation dialog when moving files to the
trash, like there is for deleting files directly or emptying the trash.

Closes #632.

### DIFF
--- a/libcaja-private/caja-file-operations.c
+++ b/libcaja-private/caja-file-operations.c
@@ -1316,6 +1316,18 @@ should_confirm_trash (void)
 }
 
 static gboolean
+should_confirm_move_to_trash (void)
+{
+	GSettings *prefs;
+	gboolean confirm_trash;
+
+	prefs = g_settings_new ("org.mate.caja.preferences");
+	confirm_trash = g_settings_get_boolean (prefs, CAJA_PREFERENCES_CONFIRM_MOVE_TO_TRASH);
+	g_object_unref (prefs);
+	return confirm_trash;
+}
+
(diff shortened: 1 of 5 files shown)


monsta:
Looks like people were asking for this feature. Thanks, merged.

Issue #281, opened 2014-04-10
items are not spoken by orca when selecting multiple files/folders in icon view


am-dxer:
When moving through items in icon view with the arrow keys, the Orca screen reader speaks the item that has focus for me since the fix posted a while ago. However, when I hold down the control key to select multiple items and try to move around with the arrow keys, Orca doesn't give any feedback about the focused item. Orca should speak the item that has focus as well as whether it is selected or not. This works properly in the list view mode.
##

monsta:
@coffeeking: can you check if it's still relevant please?

lukefromdc:
Reported as resolved by
https://github.com/mate-desktop/caja/pull/724

PR #724, opened 2017-01-27
Use ATK_STATE_SELECTED and ATK_STATE_FOCUSED appropriately for highli…

TheMuso:
Allows the user to get feedback from Orca when additional icons are selected or deselected.

### DIFF
--- a/libcaja-private/caja-icon-container.c
+++ b/libcaja-private/caja-icon-container.c
@@ -856,10 +856,10 @@ clear_keyboard_focus (CajaIconContainer *container)
 }
 
 static void inline
-emit_atk_focus_tracker_notify (CajaIcon *icon)
+emit_atk_focus_tracker_notify (CajaIcon *icon, gboolean focused)
 {
     AtkObject *atk_object = atk_gobject_accessible_for_object (G_OBJECT (icon->item));
-    atk_focus_tracker_notify (atk_object);
+    atk_object_notify_state_change (atk_object, ATK_STATE_FOCUSED, focused);
 }
 
 /* Set @icon as the icon currently selected for keyboard operations. */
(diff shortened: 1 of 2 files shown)


libcaja-private/caja-icon-container.c
```diff
@@ -856,10 +856,10 @@ clear_keyboard_focus (CajaIconContainer *container)
 }
 
 static void inline
-emit_atk_focus_tracker_notify (CajaIcon *icon)
+emit_atk_focus_tracker_notify (CajaIcon *icon, gboolean focused)
```
cwendling: This function could use some renaming then I guess

cwendling:
Looks good, and it indeed fixes Orca not announcing anything while extending a selection for me. :+1:

TheMuso:
On Fri, Jan 27, 2017 at 07:44:29PM AEDT, Colomban Wendling wrote:
>  static void inline
> -emit_atk_focus_tracker_notify (CajaIcon *icon)
> +emit_atk_focus_tracker_notify (CajaIcon *icon, gboolean focused)
> 
> This function could use some renaming then I guess

Yeah good point.

Pushed a commit to do just that.

Luke

monsta:
Nice, it should fix that deprecation warning about ```atk_focus_tracker_notify```. Looks ok to me.

monsta:
The last two commits could be made simpler though, the emit function is always called with ```TRUE``` as second argument, so it doesn't seem to be needed.

monsta:
It seems to me that https://github.com/mate-desktop/caja/issues/281 might be fixed by this PR. Can you guys confirm it please?

alexarnaud:
@monsta it seems it PR resolved effectively the bug #281.

monsta:
Thanks, it's closed now.

Issue #506, opened 2016-01-22
GTK3: Backgrounds and Emblems dialog does not work [$10]


lukefromdc:
While emblems can stll be set from item preferences or the emblems sidebar, the "backgrounds and emblems" is is now mostly nonfunctional. When loaded for the first time, a few emblems can be seen-in a very short, 1 icon deep scrolled window that is difficult to control except by the end arrows. Dragging a pattern from it to a folder background has no effect.

In "colors" or "emblems" the scrolled window is initially empty. Same in emblems if you have clicked on colors or emblems at least once since Caja was started. Changing the GTK theme will bring up the contents of the scrolled window, and they will show after changing the theme back. Emblems can actually be applied to an item this way, but dragging and dropping anything else is ignored.

This is not limited to gtk3.19, I just tested in gtk3.18 as well and got the exact same results.
##

raveit65:
Yes, here is something to do.
Port to GtkGrid , GtkStyleContext............

raveit65:
This expand the GtkScrolledWindow and make the scrollbars usable.
https://github.com/mate-desktop/caja/commit/b058255bef4dab6ce19c6b056af2b32cd2e386a0
A possible style selector:
```
CajaPropertyBrowser .vertical .horizontal .vertical GtkScrolledWindow .frame {
    background-color: @theme_base_color;
}
```

lukefromdc:
For GTK3.20 I will have to work out a "widget chain" equivalent to the no-longer-recognized
CajaPropertyBrowser selector.

lukefromdc:
window.background>box>box>box>scrolledwindow>viewport.frame 

This widget chain will match for GTK3.19, it will show the scrolled window but it
stays 1 icon deep and the icons still don't appear except the patterns on initially
opening the window.  I will watch for any accidental matches to anything else in 
using my theme.

raveit65:
I noticed for other gtk+ branches that BlackMate already use a differrent bg without using this chain.

lukefromdc:
Won't need it, that kind of widget chain is only required in GTK3.19/20 because
GNOME took out support for old style widget names. Now only cssnodes, style
classes installed by the app, and #something widget names are supported.,

lukefromdc:
OK, that commit opens up the scrollbar windows (I was confused earlier, thinking the
theming work recommended was supposed to do that.) Themes easily enough-but the
issue of disappearing icons and ignored backround D &D is still there.

[...]

PR #731, opened 2017-02-02
Fixes for background drawing

XRevan86:
This fixes directory background drawing in #506, fixes compiz-reloaded/compiz#40 (desktop background issues with Compiz 0.8.x) and fixes background fading.
Caja now draws the desktop background again and the gap in code between GTK+ 3.22 and previous versions is removed.
Other parts of the change: mate-desktop/mate-desktop#249, mate-desktop/mate-settings-daemon#169.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4,7 +4,7 @@ dnl ===========================================================================
 
 m4_define(glib_minver,                 2.36.0)
 m4_define(gio_minver,                  2.26.0)
-m4_define(mate_desktop_minver,         1.17.0)
+m4_define(mate_desktop_minver,         1.17.3)
 m4_define(pango_minver,                1.1.2)
 m4_define(gtk_minver,                  3.14.0)
 m4_define(xml_minver,                  2.4.7)
--- a/eel/eel-background.c
+++ b/eel/eel-background.c
@@ -53,6 +53,7 @@ static guint signals[LAST_SIGNAL] = { 0 };
 struct EelBackgroundDetails
 {
     GtkWidget *widget;
+    GtkWidget *front_widget;
     MateBG *bg;
     char *color;
 
@@ -78,6 +79,10 @@ struct EelBackgroundDetails
     /* activity status */
     gboolean is_active;
 };
+
+#if GTK_CHECK_VERSION (3, 22, 0)
+static GList *desktop_bg_objects = NULL;
+#endif
 
 static void
 free_fade (EelBackground *self)
@@ -124,9 +129,16 @@ eel_background_finalize (GObject *object)
     }
 
     free_background_surface (self);
-
     free_fade (self);
 
+#if GTK_CHECK_VERSION (3, 22, 0)
+    if (self->details->is_desktop)
+    {
+        desktop_bg_objects = g_list_remove (desktop_bg_objects,
+                                            G_OBJECT (self));
+    }
+
+#endif
     G_OBJECT_CLASS (eel_background_parent_class)->finalize (object);
 }
 
@@ -358,11 +370,15 @@ void
 eel_background_draw (GtkWidget *widget,
                      cairo_t   *cr)
 {
-    int width, height;
-    GdkWindow *window = gtk_widget_get_window (widget);
-    GdkRGBA color;
-
     EelBackground *self = eel_get_widget_background (widget);
+    GdkRGBA color;
+    int width, height;
+
+    if (self->details->fade != NULL &&
+        mate_bg_crossfade_is_started (self->details->fade))
+    {
+        return;
+    }
 
     drawable_get_adjusted_size (self, &width, &height);
 
@@ -372,7 +388,8 @@ eel_background_draw (GtkWidget *widget,
 
     cairo_save (cr);
 
-    if (self->details->bg_surface != NULL) {
+    if (self->details->bg_surface != NULL)
+    {
         cairo_set_source_surface (cr, self->details->bg_surface, 0, 0);
         cairo_pattern_set_extend (cairo_get_source (cr), CAIRO_EXTEND_REPEAT);
     } else {
@@ -468,7 +485,7 @@ on_fade_finished (MateBGCrossfade *fade,
 
 static gboolean
 fade_to_surface (EelBackground   *self,
-                 GdkWindow       *window,
+                 GtkWidget       *widget,
                  cairo_surface_t *surface)
 {
     if (self->details->fade == NULL ||
@@ -479,7 +496,22 @@ fade_to_surface (EelBackground   *self,
 
     if (!mate_bg_crossfade_is_started (self->details->fade))
     {
+#if GTK_CHECK_VERSION (3, 22, 0)
+        mate_bg_crossfade_start_widget (self->details->fade, widget);
+#else
+        GdkWindow *window;
+
+        if (EEL_IS_CANVAS (widget))
+        {
+            window = gtk_layout_get_bin_window (GTK_LAYOUT (widget));
+        }
+        else
+        {
+            window = gtk_widget_get_window (widget);
+        }
+
         mate_bg_crossfade_start (self->details->fade, window);
+#endif
         if (self->details->is_desktop)
         {
             g_signal_connect (self->details->fade,
@@ -494,9 +526,7 @@ fade_to_surface (EelBackground   *self,
 static void
 eel_background_set_up_widget (EelBackground *self)
 {
-    GdkWindow *window;
     GtkWidget *widget = self->details->widget;
-    GdkRGBA color;
 
     gboolean in_fade = FALSE;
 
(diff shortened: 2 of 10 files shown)


raveit65:
Seems to work fine :-)
I tested all 3 PRs with f24 (gtk+-3.20) and fedora rawhide (f26) with gtk+-3.22.
As fedora rawhide is quite buggy in this early state i back ported the changes to 1.16 and created branches for every package. Than i can test this on stable f25 with 1.16.x, also i can ask some guys in fedora-forum to test all with compiz.
I am not sure if we want to merge it to 1.16 too as it is a massive change, but i really need patches for 1.16 to fix broken compiz in f25.
No need let the users wait 6 months for a fix which comes regular with mate-1.18 on f26.

lukefromdc:
I think I have figured out a way to make this work. That is to special-case any icon-view that is a desktop and keep that one transparent or with the new background code, but apply these changes to the rest of them.  Alternately it might be possible to force an immediate redraw of the icon-view on hover to stop the dark and back flash. I'll play with that option first, as it would stop the current situation of having to load the desktop wallpaper into RAM twice when compositing is not use. Also, we should probably prevent it from being possible to drag a background from backgrounds and emblems to the desktop as that works but then causes problems for mate-settings-daemon. It also could be easily done by accident, conflicts with setting the desktop bg the normal way, and would confuse users.

lukefromdc:
I did some experiments tonight-and was not sucessful in attempting to special-case the desktop using the icon view files also used in the navigation window. Thus, to make this work, some of what Nautilus did in splitting out the desktop may be needed. Would not need to be separate binary, but would need its own icon-view etc files. Also when applying only the eel changes things mostly worked but I got a black desktop background, that would have to be tracked down, appears to be an issue with an RGB value somewhere an RGBA is needed.

Alternative is if someone can find that icon flash problem directly, then we don't need to deal with special casing the desktop transparent.

raveit65:
I found another issue with this PR.
It breaks starting caja in a root terminal.
If i only revert the changes for caja but use changes for m-s-d and mate-desktop caja starts well as root again.
Now the the blinking icons are gone in normal user session but the desktop is broken again with compiz as window-manger.
Tested on f26 (bare metal) with mate-1.17.x.

lukefromdc:
In my testing the mate-desktop change for background fading works by itself. That seems to be ready to go as a standalone bugfix.

Concerning compiz: I have had no trouble with the desktop using the newer compiz 0.9 builds as the window manager, though since I am not using Ubuntu's patched GTK I have to patch compiz to falsely advertise GtkFrameExtents support so GTK 3.16 or later will treat it as a compositor. 

A few years ago Compiz did NOT work at all with Nemo due to problems with the transparent desktop, to this day I don't know what changed to fix that. A possible fix for 0.8 might be to detect compiz 0.8 as the WM and use the noncompositing background case for it.  If this is not possible, a configuration switch to force use of the noncomposited desktop rendering system might fix it.  Downside is having to load the background image into ram twice so it can be copied, which can be significant on small netbooks etc so it should not be unconditionally forced without an option.

I have compiz 0.9, not 0.8 installed but I propose this test for compiz 0.8: Build caja with the fallback noncomposited desktop rendering used unconditionally and test it. If it works add a configuration time switch for it with a name like "compiz-reloaded compatability."

raveit65:
caja starts fine now as root or sudo after a f26 mass rebuild :-)
So this wasn't related to the PRs.

lukefromdc:
This is something I'd really like to see working but I have no clue about how to fix the icon flash other than it appears to be a forced redraw after hover. I do recall a "don't chain up" line used in the original background code from GNOME 2 that referred to exactly this problem.

lukefromdc:
I ran some more tests today. If I use only the eel changes, I get a black desktop, normal icon behavior, and normal navigation windows but of course cannot use navigation window backgrounds. No icon flash issue this way. This is true in both the compositing and noncompositing modes, indicating that something that should be RGBA is RGB and rendered above the noncomposited case background in caja master.

Another build with a rough attempt to apply the transparent desktop to the full PR was very unstable. When it first started I had a working desktop background, but the desktop icons still flashed on hover. Desktop constantly restarting though until I moved an icon, then something changed and the desktop was stable but exibited the redraw bug.

XRevan86:
I did some further changes, now there are no redraws of the desktop window with no compositing.

lukefromdc:
I just built and tested this successfully, no more icon flash problem and backgrounds and emblems now work. This seems to be ready to go.

I am not merging it myself because it depends on the merge of the dev-bgfix branch of mate-desktop and should be used with the dev-bgfix branch of mate-settings-daemon as well. I don't have write access to those so leaving all of it until someone who does can merge them all at once.

lukefromdc:
Very nice-you can even drag and drop one of the backgrounds to the desktop without problems, at least in my quick test with compiz 0.9.13 and GTK 3.22.8

raveit65:
More test results:
with branch for 1.17, tested on f24 (Mate gtk3, 3.20) and fedora rawhide (Mate gtk3, 3.22)
- no redraws
- no desktop icon flickering
- background fade is working
- using different backgrounds for directories is working again, only bg and emblems needs a bit work and fg color is wrong for dark backgrounds.
But this is a old issue and not related to PR. We can ignore this for the moment.

with branch for 1.16, tested on f25 (Mate gtk3, 3.22)
- no redraws
- no desktop icon flickering
- background fade is working
- using different backgrounds for directories is working again, only bg and emblems UI needs a bit work and fg color is wrong for dark backgrounds.

with branch for 1.16, tested on f24 (Mate gtk2)
- here all is working like before
- wrong fg color for directories isn't a problem here. So we need to look at gtk2 code to see what is different here.
I recall that it was a issue or it was fixed, i can't remember.
@monsta 
please review, and it is important to build first mate-desktop, same for creating new tarballs.
mate-desktop needs to be updated first.

monsta:
Directory background works, both colors and patterns. Good job :+1: 

What's left is to fix bg/emblems UI and fg color (it's always black, even with dark bg). These are separate issues, so this PR seems good to go. I tested it with GTK+ 3.18.9, will try with 3.22 as well just in case.

monsta:
One weird thing I've noticed though: after I built and installed Caja from this PR (over git master version), and tried to log out, the session didn't end properly. Even the running apps like mate-terminal kept running. I only got GTK+ and icon themes reset - apparently m-s-d did quit after all. Next weird thing is that I made the session end after all... by telling Caja to quit (```caja -q```). :open_mouth: 

It only happened in one session. After I logged in back, and new version of Caja started, this problem didn't appear on next logout.

I reproduced it first on baremetal with GTK+ 3.18.9, then in a VM with 3.22.8.
Anyone knows what it is? Did we trigger some hidden bug with this PR? :smile:

monsta:
Well, the new Caja wasn't running at that time. I only built and installed it, but didn't restart it.

monsta:
Did some more excessive testing. Built mate-desktop PR (which is required for this one), then this PR, but left m-s-d as is, from git master. No logout issue, no crashes, and... directory background works. How? :open_mouth: 

This is on baremetal with GTK+ 3.22.8.

monsta:
Now I added m-s-d PR to that setup, all is good, and... no logout issue again.

monsta:
One more test, now in a VM with GTK+ 3.14.5. With mate-desktop PR + this PR the directory background works, and logout issue doesn't appear. But there's a regression: the desktop background is always white. (Adding m-s-d PR to that setup doesn't change anything because of old GTK+ version.)

lukefromdc:
If a fix for the GTK 3.14/white desktop issue is not easily found, there is always the option of limiting this to newer GTK 3 versions and removing backgrounds and emblems from the menu in the older GTK 3 versions. The background-fade issue for instance is GTK 3.22 only and this takes care of it, and that issue does not exist with older GTK 3 versions and the original Caja background code, which was never changed for those older GTK 3 versions.

Having done that someone who can still build and run Caja on GTK 3.14 could pursue that issue.

raveit65:
Confirmed, using all 3 PRs from 1.17 branch gives me a white desktop :-/ with gtk+-3.14 in rhel7/centos7 VM.
Adding a background to directories is working but not for the desktop, which is possible with 3.20/3.22
@XRevan86 
Any Idea ?
If not we should limit the change for >=3.20 for the moment and for 1.18 release.
I think @flexiondotorg is waiting for a release for ubunt-17-04 with gtk+-3.22.
And we can look for a solution after the release.
I don't think that i will update rhel7/centos7 before summer to 1.18 and switching to gtk3.

raveit65:
Btw. m-s-d writes the desktop bg correctly with 3.14.

XRevan86:
> . But there's a regression: the desktop background is always white
```css
.caja-desktop-window,
.caja-desktop:not(:selected):not(:active):not(.rubberband){
	background-color: transparent;
}
```
Indeed, it is caused by this code from ```caja-desktop.css``` **not** working with Gtk 3.14…
@lukefromdc, do you think it can be adapted to work without the ```:not```'s?

lukefromdc:
Using this code:

```
.caja-desktop-window,
.caja-desktop{
	background-color: transparent;
}

.caja-desktop:selected, .caja-desktop:active{
	background-color: @theme_selected_bg_color;
}

.caja-desktop.rubberband{
	background-color: alpha(@theme_selected_bg_color,0.35);
}
```

instead of the current

```
.caja-desktop-window,
.caja-desktop:not(:selected):not(:active):not(.rubberband){
	background-color: transparent;
}

Might fix the GTK 3.14 issue at the expense of hardcoding use of @theme_selected_bg_color (which thus must exist in the theme) for selected, active, and rubberband items.  There is already a hardcoded @theme_selected_bg_color in caja.css, so Raleigh (the only theme I've ever seen without these) is already broken anyway, The bigger issue is theme authors can't change the rubberband or desktop selected item styling if we do this, so we might want to have an alternate, GTK 3.14-only version of caja-desktop.css if that is possible.

This works by removes the :not references and instead hardcoding @theme_selected_bg_color for the items that must be excluded from the transparent background: selected desktop items, active desktop items, and rubberband selection.  I have never been able to simply specify anything like :normal so I have to use :not everything else instead.  Moving the rubberband and selected/active stuff back to caja.css to permit overriding it did not work in a quick test, I think that caused the more general specification in caja-desktop.css loaded at a higher priority to override it.
```

raveit65:
yeap,
```
.caja-desktop-window,
.caja-desktop {
	background-color: transparent;
}

.caja-desktop:selected, .caja-desktop:active {
	background-color: @theme_selected_bg_color;
}

.caja-desktop.rubberband {
	background-color: alpha(@theme_selected_bg_color,0.35);
}
```
works for gtk+-3.22 and 3.14 in a quick test.

raveit65:
As the css files becomes more and more important if caja works or not, we should consider to use it inside a gresource file, to prevent users from editing it.
Nautilus or nemo do the same. Should be easy to find the relevant commit in early nautilus git history.

raveit65:
> The bigger issue is theme authors can't change the rubberband or desktop selected item styling if we do this, so we might want to have an alternate, GTK+ 3.14-only version of caja-desktop.css if that is possible.

I found an example of using a different css for another gtk+ version.
https://git.gnome.org/browse/evince/commit/?id=bd68bf35fcabb57302841b1b713f2c1a382a8f21

monsta:
Possibly stupid question: maybe don't load CSS files at all with GTK+ 3.14?
It was added only a year ago in https://github.com/mate-desktop/caja/commit/e8be955849664584f1fb697fd904ae9f85945a5e, so before it Caja did work with various GTK+ versions somehow...

raveit65:
Css is updated https://github.com/mate-desktop/caja/commit/2a3d4cbef005553cff444aad4caa11f73484e1ea
So we can merge the stuff for including in 1.18 release.
We can decide later what we do with the css files.

monsta:
Confirmed, it brings desktop bg back for 3.14, and doesn't cause any issues with 3.18.

monsta:
Ok, latest commit fixes white bg in 3.14, no CSS change needed.

raveit65:
merged
https://github.com/mate-desktop/caja/commit/438682bef1a8fea49b4c6972350bede1bfad69d0
https://github.com/mate-desktop/caja/commit/4c10ec0fdac58498c8c9616b940ff78990c86c90

PR #745, opened 2017-03-08
search: add tag support (xattr::xdg.tags)

fbarriga:
This PR add the ability to search files by tags. The tags are stored in extended attributes under `user.xdg.tags`
There is already a plugin ([xattr-tags](https://github.com/mate-desktop/caja-extensions/pull/25)) merged in the official caja-extensions repo to display the tags of the files as a column in the file list. I've another plugin on the oven to be able to add and remove tags from caja.

caja-search

And here is a video demo:
https://felipebarriga.cl/owncloud/s/3b04UY9TL05584C

**How to Test**
```
$ touch file_with_tag.txt
$ setfattr --name=user.xdg.tags --value="foo,bar" file_with_tag.txt
```
Note: you can install setfattr:
- debian/ubuntu: `# apt-get install attr`
- archlinux: `# pacman -S attr`

### DIFF
--- a/libcaja-private/caja-query.c
+++ b/libcaja-private/caja-query.c
@@ -35,6 +35,7 @@ struct CajaQueryDetails
     char *text;
     char *location_uri;
     GList *mime_types;
+    GList *tags;
 };
 
 static void  caja_query_class_init       (CajaQueryClass *class);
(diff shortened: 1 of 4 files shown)


raveit65:
Ok, that works here,
It took some time to understand that the tag can be used like a filter for an existing search.
Search like a tag without an filename or putting the 'tag name' in the search field doesn't work.
But filtering an existing search result with tags works well for files and folders.
Nice work and no regressions so far in a quick test.

lukefromdc:
Merged with https://github.com/mate-desktop/caja/commit/96fdd4f17237dc55be5e2a9cde5a20a23ce53086

monsta:
I had to fix this a bit for older gcc version: https://github.com/mate-desktop/caja/commit/e3953e675e45d59ae0ef720e4bf65ea6bab97fb4

lukefromdc:
I can confirm this latest change also builds and runs fine with the newer GCC in current Debian Unstable.  C89 defaults made trouble for me in my (non-MATE) network-manager-gnome package with support for panel themes, so this is something I've seen before.

Issue #741, opened 2017-02-21
Rename broken in list view after keyboard navigation (GTK3)


cwendling:
# Steps to reproduce

* open Caja built against GTK3 (here tested with GTK 3.20 and 3.22, and Caja 1.16.2 and master)
* make sure it is in List View
* move focus to a file or folder with the keyboard (just press <kbd>DownArrow</kbd> as many times as needed to go to the file you like)
* hit <kbd>F2</kbd>

# Expected result

Enter rename mode for the selected file

# Actual result

Nothing seems to happen (but the displayed focus line disappeared)

---

Alternatively to reproduce, you can also:

* open Caja in List View
* right click on a file *in another column than the Name one*
* hit <kbd>Escape</kbd> to quit rename mode
* hit <kbd>F2</kbd>

My wild guess is that the name column isn't properly focused initially, or something like that (but I could be totally wrong)

lukefromdc:
I can confirm this

raveit65:
Confirmed, issue exists in compact mode too.

cwendling:
> Confirmed, issue exists in compact mode too.

You mean compact view?  I can't reproduce with compact view with current master, but still with list view.

monsta:
@raveit65: in compact view we have this: https://github.com/mate-desktop/caja/issues/670

PR #750, opened 2017-03-23
Fix broken rename in list (issue #741)

mpieuchot:
This is a backport from nautilus 5b09cedf434b53a975074b346d511335fc735cba

  list-view: explicitly set cursor on file name cell renderer on rename

  The file name column has two cell renderers, one for the icon and another for
  the name string. Thus, gtk_tree_view_set_cursor() isn't enough to reliably
  start editing the file name cell since it will try to edit whatever is the
  currently focused cell in the column.

  https://bugzilla.gnome.org/show_bug.cgi?id=656128

Fixes https://github.com/mate-desktop/caja/issues/741

### DIFF
--- a/src/file-manager/fm-list-view.c
+++ b/src/file-manager/fm-list-view.c
@@ -2891,9 +2891,10 @@ fm_list_view_start_renaming_file (FMDirectoryView *view,
                                   NULL,
                                   list_view->details->file_name_column,
                                   TRUE, 0.0, 0.0);
-    gtk_tree_view_set_cursor (list_view->details->tree_view,
+    gtk_tree_view_set_cursor_on_cell (list_view->details->tree_view,
                               path,
                               list_view->details->file_name_column,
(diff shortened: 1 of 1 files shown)


src/file-manager/fm-list-view.c
```diff
@@ -2891,9 +2891,10 @@ fm_list_view_start_renaming_file (FMDirectoryView *view,
                                   NULL,
                                   list_view->details->file_name_column,
                                   TRUE, 0.0, 0.0);
-    gtk_tree_view_set_cursor (list_view->details->tree_view,
+    gtk_tree_view_set_cursor_on_cell (list_view->details->tree_view,
                               path,
                               list_view->details->file_name_column,
+                              GTK_CELL_RENDERER (list_view->details->file_name_cell),
                               TRUE);
```
cwendling: I guess the alignment should be kept.  BTW upstream commit had it: https://git.gnome.org/browse/nautilus/commit/?id=5b09cedf434b53a975074b346d511335fc735cba

PR #760, opened 2017-04-13
IconCanvasItem: fix overlapping filenames on hover

raveit65:
Just a little fix for another annoying issue.

taken from: https://git.gnome.org/browse/nautilus/commit/?id=308f4ce

PR #759, opened 2017-04-13
fix scrolling and renaming in large folders

raveit65:
fixes https://github.com/mate-desktop/caja/issues/755
First time it's a bit hard to reproduce the issue, see this posts.
https://github.com/mate-desktop/caja/issues/755#issuecomment-292500277
https://github.com/mate-desktop/caja/issues/755#issuecomment-293630213
https://www.dropbox.com/s/8ftp9gois751892/caja-issue.ogv?dl=0

Note, i added a cosmetic back port of a nemo commit in another branch (dev-experimental), which reduce the width of the gap between grid of items and vertical right scrollbar.
Taken from https://github.com/linuxmint/nemo/pull/1257
But this dynamically grid change the look and feel of the icon-view and isn't really needed to fix the issue.
Please vote here if you like this behaviour or not.

PR #761, opened 2017-04-14
fix scrolling and renaming in large folders 1.16

raveit65:
Those commits are all bug fixes for gtk3 which belongs in 1.16 branch.
I made them all gtk3 only as i don't know if they are good for gtk2.
Not sure who have 1.16 with mate gtk3 running somewhere.
Please test.

### DIFF
--- a/eel/eel-canvas.c
+++ b/eel/eel-canvas.c
@@ -578,6 +578,15 @@ eel_canvas_item_move (EelCanvasItem *item, double dx, double dy)
 
 }
 
+#if GTK_CHECK_VERSION (3, 0, 0)
+static void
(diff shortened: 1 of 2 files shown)


lukefromdc:
I can confirm this works with GTK 3.22, but someone should test this with GTK2 because Mint shipped MATE 1.16 built with GTK2

raveit65:
Ok, fine :-)
I will squash and merge the branch. Are there any other commits to backport for new 1.16 and 1.18 releases?

monsta:
I think it's all... #759 and #760 can go to 1.18 and this one goes to 1.16.

PR #762, opened 2017-04-16
Use gtk+-3 bookmarks location

raveit65:
If it doesn't exist fallback to reading from the old location but
always write to the new location.

https://bugzilla.gnome.org/show_bug.cgi?id=674986

taken from:
https://git.gnome.org/browse/nautilus/commit/?id=ed90577

This is better since we use gtk3 filechooser. Filechooser use gtk3 location and you will never seen bookmarks from ~.gtk-bookmarks  there.
For a new installation or new user account this will make no different to old behaviour.
For a existing account the user needs to export the old bookmarks to new file.
Probably it's possible to read from both locations, but i don't think it's good to use 2 locations.

Also, we sync our location with nemo or nautilus. Helpful if users switch to using MATE.

### DIFF
--- a/src/caja-bookmark-list.c
+++ b/src/caja-bookmark-list.c
@@ -115,13 +115,30 @@ new_bookmark_from_uri (const char *uri, const char *label)
 }
 
 static GFile *
-caja_bookmark_list_get_file (void)
+caja_bookmark_list_get_legacy_file (void)
 {
     char *filename;
     GFile *file;
 
     filename = g_build_filename (g_get_home_dir (),
                                  ".gtk-bookmarks",
+                                 NULL);
+    file = g_file_new_for_path (filename);
+
+    g_free (filename);
+
+    return file;
+}
+
+static GFile *
+caja_bookmark_list_get_file (void)
+{
+    char *filename;
+    GFile *file;
+
+    filename = g_build_filename (g_get_user_config_dir (),
+                                 "gtk-3.0",
+                                 "bookmarks",
                                  NULL);
     file = g_file_new_for_path (filename);
 
@@ -559,6 +576,9 @@ load_file_async (CajaBookmarkList *self,
     GFile *file;
 
     file = caja_bookmark_list_get_file ();
+    if (!g_file_query_exists (file, NULL)) {
+            file = caja_bookmark_list_get_legacy_file ();
+    }
 
     /* Wipe out old list. */
     clear (self);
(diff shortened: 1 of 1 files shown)


raveit65:
The old ~.gtk-bookmarks file will only use if the gtk3 doesn't exits.
This  you can test with different bookmark files and simply removing ~.config/gtk-3.0/bookmarks.
Adding ~.config/gtk-3.0/bookmarks again will show the bookmarks of this file.

lukefromdc:
OK, that test works: if ~.config/gtk-3.0/bookmarks is removed, Caja reads the old file, then regenerates the new one at first write to bookmarks.

monsta:
Good idea, but won't we provide some migration for users who would upgrade from earlier versions to 1.20?

raveit65:
Yeah, a migration would be nice.
For some reasons ~.config/gtk-3.0/bookmarks is always installed nowadays, so the include migration (reading from old location if new location doesn't exists) won't work.
But i've no idea to do that.
Btw. i will use that for f26 with 1.18 . F26 will release in beginning of june, so adding this now is a good moment.

lukefromdc:
One possiblity would be to read once from the old file if the new version of Caja has not been run before, then set a gsettings preference or a file in ~/config/gtk3.0 to indicate that Caja has been migrated. If the old file does not exist, read from the new one first time around.  Clunky but workable.

lukefromdc:
I just rebased this on the latest changes in master but holding on merge due to request for review from https://github.com/clefebvre

monsta:
Yes, I'd like to hear @clefebvre's opinion on this and maybe some more ideas about the migration.

lukefromdc:
Rebased again after 
https://github.com/mate-desktop/caja/commit/55087066adcf0d4dcc575bf5b3cd09d7c073aa7c
still waiting for https://github.com/clefebvre and any ideas for migration

lukefromdc:
If we're currently using a gtk2 file location with gtk3, that's rather nonstandard, though it would not break in an install with gtk2 not present if it can write the file and the directory for it. We really should fix this

lukefromdc:
https://github.com/monsta ,should I go ahead and merge this? I've been using it without issue for a week, and we haven't heard back on the migration issue

raveit65:
I don't think that nemo used ever a migration in linuxmint or any other distro as nemo never used the old gtk2 file.
At least migration is a one liner ;-)
```
$ cat .gtk-bookmarks >> .config/gtk-3.0/bookmarks 
```

monsta:
> For some reasons ~.config/gtk-3.0/bookmarks is always installed nowadays, so the include migration (reading from old location if new location doesn't exists) won't work.

I just found that my Ubuntu 17.04 VM doesn't have that file at all, and I'm pretty sure I wasn't messing with it. :slightly_smiling_face: We need to take this scenario into account and probably copy the bookmarks to this new location (not waiting until user edits any bookmark - it might not happen).

Issue #410, opened 2015-04-28
Ubuntu MATE: Folder color not working in list view


ZrN-kAyDqY:
For Ubuntu MATE 15.04 -- The folder color is not able to be changed in "List View" (but it does work in icon and compact views).

System: Kernel: 3.19.0-15-generic x86_64 (64 bit gcc: 4.9.2)
  Desktop: MATE 1.8.2 (Gtk 3.14.12-0ubuntu2) Distro: Ubuntu 15.04 vivid
  Quad core AMD Phenom II X4 965 (-MCP-) 
  Graphics: Card: Advanced Micro Devices [AMD/ATI] Cedar [Radeon HD 5000/6000/7350/8350 
  Display Server: X.Org 1.17.1 drivers: ati,radeon (unloaded: fbdev,vesa) Resolution: 1920x1200@60.0hz
  GLX Renderer: Gallium 0.4 on AMD CEDAR GLX Version: 3.0 Mesa 10.5.2 Direct Rendering: Yes

caja: Installed: 1.8.2-3; Candidate: 1.8.2-3
          Version table: **\* 1.8.2-3 0
          500 http://us.archive.ubuntu.com/ubuntu/ vivid/universe amd64 Packages
##

stoermerjp:
i can reproduce in Linux Mint 17.1 w/ Caja 1.8.2
http://snag.gy/tKXrJ.jpg <-- "List View" mode with no folder icon color for /tmp

ZrN-kAyDqY:
That's true, when I used Mint, saw it there as well.

monsta:
I recall there were fixes for this in recent Mint versions... and as for Ubuntu, 15.04 is going EOL [in a week](https://lists.ubuntu.com/archives/ubuntu-announce/2016-January/000203.html). Is it reproducible in 15.10 or 16.04?

flexiondotorg:
Still not working in 16.04.

monsta:
@alexandervdm: can you please look at this issue?

flexiondotorg:
This is reproducible in Ubuntu MATE 17.04 with MATE 1.17.

lukefromdc:
This could get ugly with GTK 3.22 and the changes it needed to icon views to handle the desktop

monsta:
In the icon and compact views it still works even with GTK+ 3.22. It's something in the list view, probably some needed metadata isn't being read from the folders.

raveit65:
[...]

PR #765, opened 2017-04-19
file: make caja_file_get_gicon return custom icons too

monsta:
Finally found the relevant upstream commit. This should fix https://github.com/mate-desktop/caja/issues/410 in Mint (using ```caja-folder-color-switcher``` package) and Ubuntu (using ```folder-color-caja``` package). Maybe some other distros have it, but I don't know.
If you don't have folder-color available, please just check that I didn't break anything, in particular, in the list view and the bookmarks (that's where ```caja_file_get_gicon``` function is used, according to grep).

### DIFF
--- a/libcaja-private/caja-file.c
+++ b/libcaja-private/caja-file.c
@@ -4123,6 +4123,11 @@ caja_file_get_gicon (CajaFile *file,
 		return NULL;
 	}
 
+	icon = get_custom_icon (file);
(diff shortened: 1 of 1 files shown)


raveit65:
This does not fix using BG&Enblems gui for setting up a background color.
Other plugins i don't have installed.
To summarize i see no collateral damages in a quick test :-)

PR #776, opened 2017-05-02
CSS: Load a different caja-desktop.css with GTK 3.14

clefebvre:
In caja-desktop.css this line isn't compatible with GTK 3.14:

".caja-desktop:not(:selected):not(:active):not(.rubberband){"

It is needed for GTK 3.20/3.22 tough.

This PR checks the GTK version and loads the appropriate CSS.

This fixes desktop icons showing black text. That issue was
reproduced with Adwaita under LMDE with GTK 3.14.

### DIFF
--- a/data/Makefile.am
+++ b/data/Makefile.am
@@ -35,6 +35,7 @@ cajadata_DATA = \
 	caja-suggested.placeholder \
 	caja.css \
     caja-desktop.css \
+    caja-desktop-3.14.css \
 	$(NULL)
 
 # app data file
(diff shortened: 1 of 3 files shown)


monsta:
Confirmed, it fixes black desktop font with Adwaita with GTK+ 3.14, and doesn't affect GTK+ 3.22.
Seems good. Maybe just change one thing - replace compile-time version check with a runtime one?

raveit65:
I would prefer something like that to avoid installing both files.
http://git.mate-desktop.org/pluma/commit/?id=222cefd36299d85575c3f11f5224b2241b325215

monsta:
It's possible, but we'll need them both for a tarball anyway :slightly_smiling_face:

raveit65:
yes, distcheck needs both. take a look at pluma link. it is all there.

raveit65:
confirmed that this fixed font color with adwaita.
But honestly is isn't allowed to use such a theme inside MATE ;-)

monsta:
Well, Adwaita is just an example of theme which doesn't set the needed stuff.

Issue #771, opened 2017-04-25
The fdb_manager was not created


zuxunyang:
I find the fdb_manager was not created by caja_freedesktop_dbus_new in the range caja-1.17.* and caja-1.18.*.
So the Firefox download-page can not use caja to open a folder.

lukefromdc:
Testing today with caja 1.19 from git master 4-24-2017, firefox 53 (built by default with gtk3) can open and create folders just fine from the file save dialog. Firefox uses a lot of custom widgets but normally filechoosers and save dialogs are standard gtk items. These descended from Nautilus code but do not actually call Nautilus, Nemo, Caja et all-they are part of GTK 3.

raveit65:
```
[rave@mother ~]$ cat /usr/share/dbus-1/services/org.freedesktop.FileManager1.service 
[D-BUS Service]
Name=org.freedesktop.FileManager1
Exec=/usr/bin/nautilus --gapplication-service
```

raveit65:
Hmm, llooks like this is only a problem if nautilus is installed.
```
[root@mother rave]# dnf provides /usr/share/dbus-1/services/org.freedesktop.FileManager1.service
nautilus-3.24.0-1.fc26.x86_64 : File manager for GNOME
Quelle      : @System
```
I will test this.

monsta:
Now it works for me: https://github.com/mate-desktop/caja/pull/778.

monsta:
Ok, it's done now, and we'll release new 1.18.x version later.

PR #778, opened 2017-05-03
restore DBus interface initialization

monsta:
Fixes https://github.com/mate-desktop/caja/issues/771

Copied from 1.16 code. Not sure if it should be placed at that point or somewhere below in the code, so feel free to move it if there's a reason for it.

### DIFF
--- a/src/caja-application.c
+++ b/src/caja-application.c
@@ -2334,6 +2334,9 @@ caja_application_startup (GApplication *app)
      */
     G_APPLICATION_CLASS (caja_application_parent_class)->startup (app);
 
+    /* Start the File Manager DBus Interface */
+    fdb_manager = caja_freedesktop_dbus_new (app);
+
     /* Initialize preferences. This is needed so that proper
      * defaults are available before any preference peeking
      * happens.


clefebvre:
Looks good to me.

Just need to confirm whether it was removed by @lukefromdc by mistake or if there was a reason for it not to be there. Maybe he removed it to workaround another issue? I don't know, Luke should review/confirm.

lukefromdc:
I will be testing this momentarily, looks like a simple piece left out by accident-and that didn't show itself on a system with both Nemo and Nautilus also installed.

lukefromdc:
Rebased and merged with 
https://github.com/mate-desktop/caja/commit/00a679615d135db4016330c18ee731a668cdd8b4

lukefromdc:
I would now advise this be cherrypicked for 1.18 as it is a bugfix

monsta:
Thanks for testing. Yes, it's a fix for regression so it should go into 1.18 too.

monsta:
Picked for 1.18 in https://github.com/mate-desktop/caja/commit/6ece67ffa2b3c90f4c0ae69147b62ea255da3636.

PR #797, opened 2017-06-21
Fix off-by-one error in caja-pathbar.c

nokangaroo:
The pixel size difference between bold and normal fonts is not always even, so
it cannot be divided by 2. (Try not dividing by 2 and using padding only on the
right side).

And why use bold font at all? The buttons change when you click on them; that
should be enough. Or use CSS to change the font color to grey for inactive
buttons (use theme colors of course):
```
/* caja pathbar */
.caja-pathbar button label {
	color: #cbcbcb;
    /*font-weight: bold;*/
}

.caja-pathbar button:checked label {
    color: #ffffff;
}
```
This will work as long as you don't set different font-weights for button label
and button:checked label.

### DIFF
--- a/src/caja-pathbar.c
+++ b/src/caja-pathbar.c
@@ -1426,7 +1426,9 @@ set_label_padding_size (ButtonData *button_data)
     pango_layout_get_pixel_size (layout, &bold_width, &bold_height);
 
     pad_left = (bold_width - width) / 2;
-    pad_right = (bold_width - width) / 2;
+    pad_right = (bold_width + 1 - width) / 2; /* this ensures rounding up - the
+    pixel size difference between bold and normal fonts is not always even and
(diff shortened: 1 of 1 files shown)


raveit65:
@nokangaroo 
Is there an optical change which we can see?
Something like before and after?

raveit65:
> And why use bold font at all? The buttons change when you click on them; that
should be enough

In which theme you see that?

lukefromdc:
This works without issue in my testing, and my theme uses a bold font on the selected ( :checked) pathbar button only. I didn't see any obvious change but that's probably just my font working well either way. At any rate, this did not over-pad my buttons nor cause them to jump, and it makes sense by preventing any font that invokes a problem from running into the end of the label. Merging

Issue #801, opened 2017-06-30
Caja two-finger, touchpad, scrolling broken in Compact View 


galik:
#### Expected behaviour

In Caja window, two fingers on touchpad scroll left or right in Compact View

#### Actual behaviour

Only scrolls right

#### Steps to reproduce the behaviour

1. Set touchpad to left/right scrolling
```bash
synclient VertTwoFingerScroll=1
synclient HorizTwoFingerScroll=1
```
2. Select Compact View (Menu->View->Compact)
3. Try to scroll sideways using touchpad
#### MATE general version

From 1.16 - 1.18

#### Package version

From: caja-1.16.2-1.fc25.x86_64
To: caja-1.18.3-1.fc25.x86_64
(and counting)
#### Linux Distribution

Fedora 25

Bug registered with Fedora in February:
https://bugzilla.redhat.com/show_bug.cgi?id=1426130

[Originally posted to mate-desktop](https://github.com/mate-desktop/mate-desktop/issues/271)

galik:
Some additional info. Looking back in my package manager history I discovered the bug ppeared between these versions:
```
caja-1.16.1
caja-1.16.2
```
I have isolated what I believe may be the code change in question being here:
```
src/file-manager/fm-icon-view.c
```
It could be this commit:
```
commit d773ac2760052190cccb7ca1161a005fb2464b7d
Author: lukefromdc <lukefromdc@hushmail.com>
Date:   Thu Nov 3 23:31:30 2016 -0400

    GTK+3: fix mousewheel scrolling in compact view
    
    with patch by ZenWalker to fix horizontal wheel scrolling
    
    cherry-picked and squashed from:
    https://github.com/mate-desktop/caja/commit/f34590dd7a71251515b90cd9812878b7d26d30b6
    https://github.com/mate-desktop/caja/commit/ae053dbf2cc26ae74df5e782261c3a3bc7126763
    https://github.com/mate-desktop/caja/commit/3e93f938000cdb24a6137daecd77fe010ae93f36
```

lukefromdc:
Reverting that breaks mousewheel scrolling. I don't know how to do touchpad scrolling and the only laptop I own is a tiny netbook.  I normally scroll only with mousewheels or scrollbars so its no wonder I never saw that.

Who here is familiar with touchpad scrolling and can work on this?

If it works in one direction that should be a clue to the other. Since the offending commit came from Nautilus, there may also be a corrosponding Nautilus commit relating to this.

[...]

PR #803, opened 2017-07-02
Fix for Issue 801

galik:
I propose this for testing. This fixes the issue on my system.

https://github.com/mate-desktop/caja/issues/801

Changed delta scroll detection from `== 1.0` and `== -1.0` to ` > 0` and
`< 0` respectively as the delta values are not always exact. They are
some fraction.

### DIFF
--- a/src/file-manager/fm-icon-view.c
+++ b/src/file-manager/fm-icon-view.c
@@ -2499,15 +2499,13 @@ fm_icon_view_scroll_event (GtkWidget *widget,
 
             /* transform vertical integer smooth scroll events into horizontal events */
             if (scroll_event_copy->direction == GDK_SCROLL_SMOOTH && scroll_event_copy->delta_x == 0) {
-                if (scroll_event_copy->delta_y == 1.0) {
+                if (scroll_event_copy->delta_y > 0) {
                     scroll_event_copy->direction = GDK_SCROLL_DOWN;
-                } else if (scroll_event_copy->delta_y == -1.0) {
+                } else if (scroll_event_copy->delta_y < 0) {
                     scroll_event_copy->direction = GDK_SCROLL_UP;
                 }
             }
-            if ((scroll_event_copy->direction == GDK_SCROLL_UP) || (scroll_event_copy->delta_x == -1.0))
-
-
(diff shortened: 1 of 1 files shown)


monsta:
@sc0w: can you please test this? Need to make sure this doesn't break fix from https://github.com/mate-desktop/caja/pull/674 for mousewheel, in particular, the horizontal scroll with the wheel. My mouse doesn't have that feature, so I can't test it.

galik:
I am not 100% convinced this is the right fix now.

I notice that with this fix the scrolling is *much faster* than reverting to the `v1.16.1` code which scrolls much nicer. My current theory is that with this code, even very small "smooth" scrolls are being redirected as "whole step" scrolls. The `v1.16.2` code probably avoids that by only redirecting "smooth" scrolls that are exactly `1.0` (whole steps) in *value*. But that only works on systems that only generate "smooth" scrolls of exactly `1.0`.

raveit65:
It is possible that using libinput makes scrolling faster.
xinput --list
xinput --list-props <id>
will tell us the truths.

raveit65:
Btw. we need to change commit header before someone merge this, please.

raveit65:
Confirmed , that this fixes the scrolling in compact-view.
But this is definitely to fast. Scrolling on /usr/lib64 folder is like watching the matrix now ;-)

raveit65:
@monsta 
mouse-scrolling works mouse-wheel.

galik:
The function `fm_directory_view_handle_scroll_event()` uses a `static` variable to accumulate all the smooth scroll events, triggering a unit scroll after they reach `1.0` or `-1.0`. I wrote some code that does similar with this function but the difficulty is that the `y` scroll events interfere with the `x` scroll events so if I scroll down and right the down counteracts the right...

I think there needs to be a way to detect between up/down scrolling (with say a mouse wheel) and left/right scrolling (with a touch pad). But I think the problem with  left/right scrolling (with a touch pad) is you need to ignore up/down scrolling if left/right scrolling is enabled and use up/down scrolling if left/right scrolling is not enabled.

galik:
Please see new PR https://github.com/mate-desktop/caja/pull/804

PR #804, opened 2017-07-03
fix vertical Compact View scrolling Issue 801

galik:
**Offered for testing.**

Should fix [Issue: 801](https://github.com/mate-desktop/caja/issues/801).

Was always scrolling to right when using sideways scrolling devices.

Now scrolls left or right with sideways scrolling devices and adjusts
vertical scrolling devices to scroll sideways.

I have completely rewritten the function `fm_icon_view_scroll_event()`. It used to fire a new event as a means to map vertical scroll events to horizontal ones. That made it difficult to find a solution without causing an infinite loop. Now it simply modifies the current event's values as needed.

It assumes that zero valued horizontal scroll events imply a vertical only scrolling device. This works well on my system changing the touchpad abilities between horizontal scrolling, vertical scrolling and both at the same time.

I am unable to test this using a mouse at this time.

### DIFF
--- a/src/file-manager/fm-icon-view.c
+++ b/src/file-manager/fm-icon-view.c
@@ -2479,48 +2479,29 @@ fm_icon_view_scroll_event (GtkWidget *widget,
                            GdkEventScroll *scroll_event)
 {
     FMIconView *icon_view;
-    GdkEvent *event_copy;
-    GdkEventScroll *scroll_event_copy;
     gboolean ret;
 
     icon_view = FM_ICON_VIEW (widget);
-    if (icon_view->details->compact &&
-        (scroll_event->direction == GDK_SCROLL_UP ||
-        scroll_event->direction == GDK_SCROLL_DOWN ||
-        scroll_event->direction == GDK_SCROLL_SMOOTH)) {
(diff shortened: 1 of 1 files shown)


lukefromdc:
Scrolling left and right in compact view with a single scrollwheel mouse works fine with this both left and right, as does scrolling up and down in the icon and list views

raveit65:
During this test i noticed that there is a different with scroll-direction if you use the mouse-wheel over the compact-view or if the cursor is over the scrollbar itself.
If the the cursor is over the compact-view the scroll direction is the same like using 2-fingers-scrolling with the touchpad, but over the scrollbar the mouse-wheel scrolls in the other direction.
But this issue was there before with 1.18.3 release and isn't related to this PR.

monsta:
> If the the cursor is over the compact-view the scroll direction is the same like using 2-fingers-scrolling with the touchpad, but over the scrollbar the mouse-wheel scrolls in the other direction.

Just putting it here for reference: it turned out to be a GTK+ issue. We discussed it in https://github.com/mate-desktop/caja/issues/855.
It's in GtkRange, so it affects GtkScale and GtkScrollbar which are inherited from it.
(There was a report about scales as well, https://github.com/mate-desktop/mate-control-center/issues/305.)

Issue #807, opened 2017-07-08
[1.18]  "Browse media when inserted" option is broken


monsta:
#### Expected behaviour

When "Browse media when inserted" option (in Media tab in settings dialog) is enabled, inserted USB drives should be auto-opened in a new directory window.

#### Actual behaviour

The mentioned option does nothing. USB drives are only auto-mounted, but never auto-opened.

#### Steps to reproduce the behaviour

Well, insert some USB drive...

#### MATE general version

1.18

#### Package version

1.18.3

#### Linux Distribution

Linux Mint 18.2

@lukefromdc: can you please have a look at this? I traced it down to 9e5ea15d104720cfee752c193b77f8b03558c6b9.

Building from its parent commit (24bb72b321c044479728fd1d43c941afc0309f18) makes the feature work. I had to install libunique to build it of course... Building from 9e5ea15d104720cfee752c193b77f8b03558c6b9 breaks the feature, no matter if I use `--disable-libunique` option or not. I assume the libunique build is the default there (I don't remember that now, but the commit message hints about that). Looks like the feature is broken for both GtkApplication and libunique.

lukefromdc:
Libunique support was removed before 1.18 was released, as of that commit the default was to build with libunique. The code for libunique builds should not have changed, that may be a clue.

monsta:
Well, it seems like something was changed after all, as building from that commit with libunique already shows the problem.

lukefromdc:
If I can find that difference it should find the problem

lukefromdc:
I will also check the function of this in Nemo, as we now only use GtkApplication. My guess is that whatever watches for this is not getting started.

lukefromdc:
I just found this not working in GNOME and not working in Nemo either at the moment

lukefromdc:
https://github.com/mate-desktop/caja/pull/808 is a one-liner and I didn't find it in Nemo, so GNOME might have moved the handling for this to somewhere else somewhere around their origjnal Libunique/GtkApplication transition. Now broken there too it seems.

monsta:
Yeah, they removed it early in 3.0 development phase. I searched for "autorun" in Nautilus commits and there are some hints there. Autorun seems to be moved to g-s-d instead, and media handling settings to g-c-c.

lukefromdc:
[...]

PR #808, opened 2017-07-08
Fix browsing/autorun media on insertion

lukefromdc:
Fix https://github.com/mate-desktop/caja/issues/807

A one-liner, tested with GTK 3.22.16 and works. 
Looks like this was left out of Nautilus 3.0, as I didn't find it in Nemo either

Both browsing inserted flash drives and auto-opening GNOME MPV to play a video DVD worked.  Responded properly to the preference settings. 

Note that on first run, setting preferences back to "never prompt or start programs on media insertion" while the DVD was still in and GNOME MPV still open generated a segfault (in libgtk), but this was not repeated on any subsequent runs when letting the media player autostart, then setting the preferences back to "never..." with the disk either playing or stopped.

### DIFF
--- a/src/caja-application.c
+++ b/src/caja-application.c
@@ -1373,6 +1373,7 @@ mount_added_callback (GVolumeMonitor *monitor,
         caja_directory_force_reload (directory);
         caja_directory_unref (directory);
     }
+	caja_autorun (mount, autorun_show_window, application);
 }
 
 static CajaWindowSlot *


monsta:
Ok, so it seems another line not added to GtkApplication code, so it got lost during libunique removal (similar to https://github.com/mate-desktop/caja/issues/771). I'll check it.

About the segfault, which app crashed - Caja or the media player?

lukefromdc:
It tas caja that crashed

monsta:
I don't have a DVD to test this though... but with USB drive still plugged in, the first run of new Caja was fine, no crashes.

lukefromdc:
Crash could have been something unrelated in the preferences dialog

PR #815, opened 2017-07-16
avoid deprecated gdk_screen_get_number

sc0w:


lukefromdc:
I just tested this, it seems to work fine with one monitor or two, and icons can be put on the desktop on either screen. In a quick test of opening navigation windows and moving desktop icons, I could not find any behavioral difference from current master. Seems good to go.

PR #817, opened 2017-07-20
private: Ensure we allow launching .desktop files from trusted symlinks

ikeydoherty:
When a file is a symlink to one of the XDG data dirs, we'll allow that
link to work, as they're vendor provided and not world-writeable by a
malicious entity.

This lookup is handled by the "is_system_dir" logic, to ensure that whatever
the .desktop symlink is pointing at is already implicitly whitelisted. This
allows for vendor provided "default" symlinks on within user directories,
such as installer shortcuts for LiveCDs, etc.

Signed-off-by: Ikey Doherty <ikey@solus-project.com>

### DIFF
--- a/libcaja-private/caja-directory-async.c
+++ b/libcaja-private/caja-directory-async.c
@@ -3802,6 +3802,43 @@ file_info_start (CajaDirectory *directory,
     g_object_unref (location);
 }
 
+static gboolean is_trusted_system_desktop_file (GFile *file)
+{
+    gboolean res = FALSE;
+    GFileInfo *info;
+    const gchar *target = NULL;
+    GFile *location = NULL;
+
+    info = g_file_query_info (file,
+                              G_FILE_ATTRIBUTE_STANDARD_TYPE ","
+                              G_FILE_ATTRIBUTE_STANDARD_SYMLINK_TARGET,
+                              G_FILE_QUERY_INFO_NOFOLLOW_SYMLINKS,
+                              NULL,
+                              NULL);
+
+    if (info == NULL)
+    {
+        return FALSE;
+    }
+
+    target = g_file_info_get_symlink_target (info);
+    if (!target) {
+        goto done;
+    }
+
+    location = g_file_new_for_path (target);
+
+    res = caja_is_in_system_dir (location);
+
+done:
+    if (location) {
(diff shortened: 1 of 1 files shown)


ikeydoherty:
" to a file in /etc/xdg/autostart"

This isn't a trusted vendor directory from my understanding. Also /etc/xdg/autostart in reality is meant to be related to files in /usr/share/applications. The /usr/share is the vendor-domain, i.e. OS-shipped, and implicitly trusted

lukefromdc:
OK, this works fine with a symlink to a file in /usr/share/applications, will merge this

lukefromdc:
Holding merge because of request to hold other PRS in
https://github.com/mate-desktop/caja/pull/824#issuecomment-323568583

lukefromdc:
Remember that most binaries owned by root have user execution priviliges (such as 755), so that a user can run but not replace or alter them.  The .desktop files seem to be a special case as they normally do not set nor require execution priviliges to function (thus permissions 644 as no execute bit is set). 

Thus the ownership of the .desktop file does not determine who can click and run it, rather it determines who can replace it. This is important for things like preventing someone from renaming say "ransomlock_encrypt.desktop" as one of your normal .desktop files in the /usr/share/applications. Those are trusted because they are normally shipped by the distro and are not writeable by a user who does not have sudo priviliges. This matters even on a single-user machine, because a sucessful attack on your browser that does not also involve a privilige escalation attack does not gain the ability to write to these files as a way to trick you into executing malicious software, nor to write to one of the autostart directories.

This is why a symlink to one of the system .desktop files is safe, and why this PR should provide more user convenience without any less security. 

I'm not sure how caja-preferences is supposed to show permissions on symlinks (the link or the linked item?) in general but that is a different discussion from this PR and probably not a special case.

ikeydoherty:
It could be that Caja is handling the `.desktop` file type specially, given that it can edit the command lines and such. Thus it might resolve the link internally and find its a read only desktop file (as the current user)

raveit65:
yeah, go ahead.

lukefromdc:
Merged with 
https://github.com/mate-desktop/caja/commit/61fa153540c2931cbd7afe62752326d63b63b819

PR #825, opened 2017-08-08
Adds support for ctrl-tab and ctrl-shift-tab

### LINKED ISSUE #768: [Feature Request]: Switch tabs with CTRL+ TAB [$20]
It'd be nice to be able to switch open tabs the same way we can for Firefox: CTRL+TAB. and CTRL+SHIFT+TAB for reverse switch.

<bountysource-plugin>

---
There is a **[$20 open bounty](https://www.bountysource.com/issues/44432568-feature-request-switch-tabs-with-ctrl-tab?utm_campaign=plugin&utm_content=tracker%2F651521&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F651521&utm_medium=issues&utm_source=github).
</bountysource-plugin>

snaiperskaya96:
Fixes #768

### DIFF
--- a/src/caja-navigation-window-menus.c
+++ b/src/caja-navigation-window-menus.c
@@ -49,6 +49,7 @@
 #include <eel/eel-string.h>
 #include <libxml/parser.h>
 #include <gtk/gtk.h>
+#include <gdk/gdkkeysyms.h>
 #include <glib/gi18n.h>
 #include <libcaja-private/caja-file-utilities.h>
 #include <libcaja-private/caja-global-preferences.h>
@@ -819,6 +820,22 @@ action_tab_change_action_activate_callback (GtkAction *action, gpointer user_dat
     }
 }
 
+static gboolean
+on_key_press (GtkWidget *widget, GdkEventKey *event, gpointer user_data)
+{
+    if (event->keyval == GDK_KEY_ISO_Left_Tab && event->state == (GDK_CONTROL_MASK | GDK_SHIFT_MASK))
+    {
+        action_tabs_previous_callback(NULL, user_data);
+        return TRUE;
+    }
+    if (event->keyval == GDK_KEY_Tab && event->state == GDK_CONTROL_MASK)
+    {
+        action_tabs_next_callback(NULL, user_data);
+        return TRUE;
+    }
+    return FALSE;
+}
+
 static const GtkActionEntry navigation_entries[] =
 {
     /* name, stock id, label */  { "Go", NULL, N_("_Go") },
@@ -1044,6 +1061,8 @@ caja_navigation_window_initialize_actions (CajaNavigationWindow *window)
     g_signal_connect (window, "loading_uri",
                       G_CALLBACK (caja_navigation_window_update_split_view_actions_sensitivity),
                       NULL);
+
+    g_signal_connect (G_OBJECT (window), "key_press_event", G_CALLBACK (on_key_press), window);
 
     caja_navigation_window_update_split_view_actions_sensitivity (window);
 }
--- a/src/caja-navigation-window-menus.c
+++ b/src/caja-navigation-window-menus.c
@@ -823,12 +823,12 @@ action_tab_change_action_activate_callback (GtkAction *action, gpointer user_dat
 static gboolean
 on_key_press (GtkWidget *widget, GdkEventKey *event, gpointer user_data)
 {
-    if (event->keyval == GDK_KEY_ISO_Left_Tab && event->state == (GDK_CONTROL_MASK | GDK_SHIFT_MASK))
+    if (event->keyval == GDK_KEY_ISO_Left_Tab && event->state & GDK_CONTROL_MASK)
     {
         action_tabs_previous_callback(NULL, user_data);
         return TRUE;
     }
-    if (event->keyval == GDK_KEY_Tab && event->state == GDK_CONTROL_MASK)
+    if (event->keyval == GDK_KEY_Tab && event->state & GDK_CONTROL_MASK)
     {
         action_tabs_next_callback(NULL, user_data);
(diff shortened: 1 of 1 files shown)


lukefromdc:
How do I test this, what are these to be used for?

snaiperskaya96:
Hi,
please have a look at the ticket referenced on the first comment.
To test it compile it as usual and ctrl + tab/ctrl + shift + tab to switch tabs.

```
git clone https://github.com/snaiperskaya96/caja.git 
cd caja
git checkout add-support-for-ctrl-tab-and-ctrl-shift-tab
./autogen.sh
make
sudo make install
```

sc0w:
I don't know why, but it doesn't work here, can someone else test it?

lukefromdc:
@sc0w, what GTK version are you using? Wonder if that's the difference?

snaiperskaya96:
Please make sure that there's no another caja process running. You could also try to restart

sc0w:
@lukefromdc 

> what GTK version are you using? Wonder if that's the difference?

gtk 3.22.18 with the system updated in debian testing

@snaiperskaya96 

> Please make sure that there's no another caja process running. You could also try to restart

yes, I did, I restarted to make sure

raveit65:
Hmm,  old keybindings are defined here https://github.com/mate-desktop/caja/blob/master/src/caja-navigation-window-menus.c#L868
But where are the new ones in you commit?

snaiperskaya96:
@raveit65 In GTK you cannot define CTRL+SHIFT+TAB shortcut using an accelerator like normal bindings. So you have to intercept the key event and go ahead from there.
See `src/caja-navigation-window-menus.c`.

snaiperskaya96:
Btw it's probably the same problem as Pluma (https://github.com/mate-desktop/pluma/pull/243) , these bindings work with the Mac keyboard but not with standard ones. Will edit tonight

raveit65:
Not releated to your PR...
offtopic on
I am really missing the F6 key for switching panes in slit-view
https://github.com/mate-desktop/caja/blob/master/src/caja-navigation-window-menus.c#L851
Am i the only one where this doesn't work?
offtopic off

raveit65:
ok, forgot my last comment, now F6 is working.

snaiperskaya96:
If you have a moment, can you try this last commit please?

raveit65:
Last commit makes crtl+tab and crtl+shift+tab working, but old keybindings are still working.
I would really like to see only one working keybings, to avoid that users are able to learn 2 different ways of switching tabs ;-)

snaiperskaya96:
I kept both ways as per this comment https://github.com/mate-desktop/caja/issues/768#issuecomment-320886393

raveit65:
I do not agree
@monsta @lukefromdc @sc0w 
Your opinion please.

alexarnaud:
Le 24/08/2017 à 11:11, raveit65 a écrit :
> I do not agree
> @monsta <https://github.com/monsta> @lukefromdc 
> <https://github.com/lukefromdc> @sc0w <https://github.com/sc0w>
> Your opinion please.

IMO, it's not an issue if we could use both keyboard shortcuts. For 
example it's the behavior on Firefox. People accustomed to using 
ctrL+tab could used it and other that prefer ctrl+pgup/pgdown will be 
also happy.
It could also avoid to have a settings for choosing between them.

Best regards.
-- 
Alex ARNAUD

raveit65:
- ctrl+pgup/pgdown can be used for another function

snaiperskaya96:
Not sure if my opinion could worth anything (lol), but if it does i think
users could only benefit from having both the shortcut available.
Will make old and new users happy, so the first ones do not have to change
habit, and the second ones will find standardised shortcuts.
It's also good to have shortcuts on both, left and right sides, i
personally find myself using different shortcuts based on what is more
comfortable for that moment (for example pasting in the terminal via
ctrl+shift+v or via shift insert)

raveit65:
I think a teacher at school will be crazy for that, to teach 3 options for the same thing ;-)
But If everybody like 3 options for the same action i am not against it.
Personal i won't merge it before a statement from @monsta

snaiperskaya96:
>I think a teacher at school will be crazy for that, to teach 3 options for the same thing ;-)
That's 'cause of the lazy nature of students (...and teachers actually) 😄 
It's always good to have alternatives imo, especially because, again, some persons might be used to older shortcuts, and other persons just prefer newer ones

raveit65:
Btw. same problem here as with pluma PR.
Using crtl+tab for notebook tabs breaks using crtl+tab for the primary toolbar.
We need a gsettings key for switching from old to new behaviour.
Believe me, breaking old behaviour is the last thing what a linux user like ;-)

src/caja-navigation-window-menus.c
```diff
@@ -819,6 +820,22 @@ action_tab_change_action_activate_callback (GtkAction *action, gpointer user_dat
...
 
+static gboolean
+on_key_press (GtkWidget *widget, GdkEventKey *event, gpointer user_data)
+{
+    if (event->keyval == GDK_KEY_ISO_Left_Tab && event->state & GDK_CONTROL_MASK)
+    {
+        action_tabs_previous_callback(NULL, user_data);
+        return TRUE;
+    }
+    if (event->keyval == GDK_KEY_Tab && event->state & GDK_CONTROL_MASK)
```
mspeers: Should this be else if statement? One or the other not both of them. I'm new so this project?
Also like this personalty
 if (event->state & GDK_CONTROL_MASK)
{
     switch (event->keyal)
.....
}

sc0w:
already done in https://github.com/mate-desktop/caja/pull/1037

PR #824, opened 2017-08-07
avoid deprecated gdk_screen_get_width/height and gdk_screen_width/height

sc0w:
avoid deprecated:

- gdk_screen_get_width/height

- gdk_screen_width/height

PR #831, opened 2017-08-20
Fix: Scaling the desktop wallpaper doesn't work

sc0w:
Fixes https://github.com/mate-desktop/caja/pull/824#issuecomment-323568583

@raveit65 @lukefromdc please test

### DIFF
--- a/eel/eel-background.c
+++ b/eel/eel-background.c
@@ -311,13 +311,17 @@ drawable_get_adjusted_size (EelBackground *self,
 {
     if (self->details->is_desktop)
     {
-        gint sc_width, sc_height;
+#if GTK_CHECK_VERSION (3, 22, 0)
+        GdkRectangle geometry;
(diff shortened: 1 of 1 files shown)


lukefromdc:
OK, I just managed to duplicate the issue by selecting "stretch" for the wallpaper scaling option, and indeed this PR fixes it.

"Zoom" always works, "scale" always leaves uncovered areas due to the way  it works, both of them seem to work the same way with or without this PR due to not triggering the issue in some way. With stretch it is quite drastic without this PR and very ugly.

raveit65:
I think it depends on picture size which mode is affected. I can reproduce the issue with 'zoom' mode, which is the default and should work.

raveit65:
So, gdk_window_get_geometry () was missing for gtk+-3.22, we need to check that in all similar PRs which are merged.
I will test this PR a bit and merge it later.

raveit65:
Thanks for quick reaction.

PR #744, opened 2017-03-07
Changing the resolution will cause the desktop icons to not be adaptively arranged

nanxiongchao:
After changing the screen desktop resolution, the icon
can not be completely display and automatic arrangement

Get the actual screen height, determine whether the screen
"smaller", and only this situation rearrange the icon
Add the "screen size change" event and call the
rearranged icon function

Test Methods:
Create vertical two-row desktop icons close to the upper
and lower boundaries of the screen, increase and decrease
the resolution, observe whether all the icons are displayed properly

PR #837, opened 2017-08-28
make sure scaling the desktop wallpaper works in all monitors

### LINKED ISSUE #836: black wallpaper with 2 monitors
#### Expected behaviour
exactly the same wallpaper on all monitors

#### Actual behaviour
Wallpaper is black on one monitor

Install caja-1.19.1

#### MATE general version
1.19.x

#### Package version
master

#### Linux Distribution
fedora 26

not applicable,  as 1.19.1 isn't in downststream repos.

caja-desktop-2-monotors

Issue is caused by https://github.com/mate-desktop/caja/commit/ac12e6d16f4a51b3327d013781e72cfec17f3624
Removing that fix brings back the wallpaper for the second monitor.

Btw. icons are working on that monitor, so only drawing the background failed.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/48692009-black-wallpaper-with-2-monitors?utm_campaign=plugin&utm_content=tracker%2F651521&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F651521&utm_medium=issues&utm_source=github).
</bountysource-plugin>

sc0w:
Fixes #836 

It is difficult to me solve this, because I have only one monitor

Anyway, if we can't fix it, we can use the deprecated code at this point

@lukefromdc @raveit65 please test

### DIFF
--- a/eel/eel-background.c
+++ b/eel/eel-background.c
@@ -313,10 +313,26 @@ drawable_get_adjusted_size (EelBackground *self,
     {
 #if GTK_CHECK_VERSION (3, 22, 0)
         GdkRectangle geometry;
-        gdk_monitor_get_geometry (gdk_display_get_primary_monitor (gtk_widget_get_display (self->details->widget)),
-                                  &geometry);
-        *width = geometry.width;
-        *height = geometry.height;
+        GdkMonitor  *monitor;
+        GdkDisplay  *display;
+        int n = 0;
+        int i = 0;
+        int sc_width = 0;
+        int sc_height = 0;
+
+        display = gtk_widget_get_display (self->details->widget);
+        n = gdk_display_get_n_monitors (display);
+
+        for (i = 0; i < n; ++i)
+        {
+            monitor = gdk_display_get_monitor (display, i);
+            gdk_monitor_get_geometry (monitor, &geometry);
+            sc_width = sc_width + geometry.width;
+            sc_height = sc_height + geometry.height;
+        }
+
+        *width = sc_width;
+        *height = sc_height;
 #else
         GdkScreen *screen = gtk_widget_get_screen (self->details->widget);
         *width = gdk_screen_get_width (screen);


lukefromdc:
This works, some minor issues with backgrounds overlapping can occur, especially if you both add a second monitor and change the resolution of the primary monitor during the same run of Caja and without resetting the background image. It seems to get the background image to appear in the right place on both monitors, it needs to be set up once with both monitors running, then can be freely turned on and off with the same resolutions from then on.

Also, if the resolution is change, the desktop background image is rescaled but the actual caja-desktop widget is NOT rescaled, either being cut off on the right (one monitor) when scaled down or part of the desktop from the primary monitor appearing on the monitor to the right (two monitors). 

Will check to see if 1.18 also does this, but still a lot better then before. With one monitor, the background image now properly scales, and two monitors can be used and the background made to appear on both

lukefromdc:
Results: Failure to rescale the desktop is also present in 1.18 and not caused by this commit, the older code does however properly locate (center) the background on a secondary monitor while the new code requires this to be fiddles with. Could merge this as-is, then worry about the centering later if you can't find an easy fix, or fix it now if you have an idea to try right away

sc0w:
so, the behavior with 1.18 and here is the same?

lukefromdc:
Thw behavior that is the same is NOT the background scalling, but rather the scaling of the FOREGROUND, the grid of icons laid on the desktop.  On restarting caja, the foreground is resized to fit the resolution. 

To test, set up some icons all the way across the desktop. Scale the resolution down and some disappear off the end. Scale the resolution back and they return. Now scale the resolution down and restart Caja, the missing icons appear at the right hand edge of the screen as intended. Scale back up and you can see their new positions.

The fix for this would be to resize the desktop itself (not the backgrouind) when the resolution changes, so as to remove the need to restart Caja to scale it. Downside is of course that a game etc that changes the desktop resolution down would trash the right hand (and maybe bottom) part of the user's icon layout, so maybe we are better off NOT to fix this?

There are no really good solutions  at the moment. A PR to adaptively rearrange icons on resize 
https://github.com/mate-desktop/caja/pull/744
also trashes the user's layout of the icons on any change of desktop resolution, a real deal breaker with a desktop like mine and two monitors with one switched on and off-or any game that changes the resolution.

raveit65:
This works well :-)
No black background and the wallpaper spans always to the whole monitor if changing resolution,
and none deprecation warnings.

The problem with the icons is an old issue and not related to this PR, i am not sure if this really was working with MATE gtk2.
Sorry, my rhel7 installation  on bare metal use MATE-1.18 with gtk3 already, so i can't test this.
Normal rhel7 installation with Mate-1.16 (gtk2) i use only in a VM.
I will test this a bit more and merge it later.

raveit65:
Surprise, my rhel7 installation on my main box use MATE-1.16 (gtk2).
The problem with the desktop icons is still there if choose a lower resolution.
Some of the icons are out of the desktop area.
So , there isn't a different to MATE gtk3 version.

raveit65:
Thanks a lot.
https://github.com/mate-desktop/caja/commit/4c130e22b98bbc1a85747adac3b582e45d34514e

PR #835, opened 2017-08-25
Caja zoom in list view is broken

snaiperskaya96:
Fixes #790 

Hello, it's me again.
```
git clone https://github.com/snaiperskaya96/caja.git
git checkout caja-zoom-in-list-view-is-broken
...
```
or if you don't want to clone another repository 
```
git remote add snaiperskaya https://github.com/snaiperskaya96/caja.git
git fetch snaiperskaya
git checkout snaiperskaya/caja-zoom-in-list-view-is-broken
```

PR #838, opened 2017-08-28
list view: fix regression from #835

monsta:
With this PR, the behavior is as follows:

- column resizing works again
- if you don't resize a column manually before zooming, it will adapt its size on zoom in/out
- if you resize it manually before zooming, it will keep the size you've set (i.e. size won't change on zoom in/out)
- changing from list view to some other one and back will reset that behavior

I also checked the behavior in GTK+2 build of 1.16:

- column resizing works, but the minimum size of Name column is way too big, more than a half of the window
- if you don't resize a column manually before zooming, it will get bigger on zoom in, but doesn't get smaller on zoom out
- if you resize it manually before zooming, it will keep the size you've set (i.e. size won't change on zoom in/out)
- changing from list view to some other one and back will reset that behavior

Not sure if the two differences in GTK+2 build are bugs or not.

### DIFF
--- a/src/file-manager/fm-list-view.c
+++ b/src/file-manager/fm-list-view.c
@@ -2777,6 +2777,9 @@ fm_list_view_set_zoom_level (FMListView *view,
     gtk_cell_renderer_set_fixed_size (GTK_CELL_RENDERER (view->details->pixbuf_cell),
                                       -1, icon_size);
 
+    /* FIXME: https://bugzilla.gnome.org/show_bug.cgi?id=641518 */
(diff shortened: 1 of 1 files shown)


raveit65:
I can't make the first column (name) smaller as a specific size. Do set up a min .size for first columnn?

monsta:
First column indeed has a minimum size, I just pointed out the size is too big in GTK+2 build of 1.16. Not sure if it's GTK+2 problem or 1.16 one (maybe it was fixed later).

lukefromdc:
Very nice adjustment to minimum column width. Merged

PR #842, opened 2017-09-01
Fix: high cpu while loading background image

### LINKED ISSUE #606: caja use high cpu
I am running linux mint mate 18.0 , when I refresh desktop with `F5`, it keeps in refreshing status (the icons on desktop is not showing), and there is a process called `caja` use 100% cpu, I need to reboot to solve this, and it happens from time to time.

Mint mate 17.2 don't have this issue, but 18.0 do, any idea ?
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/36733835-caja-use-high-cpu?utm_campaign=plugin&utm_content=tracker%2F651521&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F651521&utm_medium=issues&utm_source=github).
</bountysource-plugin>

sc0w:
Fixes #606

with this PR the gssetings key `org / mate / background / picture-filename` only accepts an existing filename

### DIFF
--- a/eel/eel-background.c
+++ b/eel/eel-background.c
@@ -1062,6 +1062,14 @@ void
 eel_bg_load_from_gsettings (EelBackground *self,
 			    GSettings     *settings)
 {
+    char *keyfile = g_settings_get_string (settings, MATE_BG_KEY_PICTURE_FILENAME);
+
+    if (!g_file_test (keyfile, G_FILE_TEST_EXISTS) && (*keyfile != '\0'))
+    {
+        *keyfile = '\0';
+        g_settings_set_string (settings, MATE_BG_KEY_PICTURE_FILENAME, keyfile);
+    }
+
     if (self->details->bg)
         mate_bg_load_from_gsettings (self->details->bg,
         			     settings);


raveit65:
Merging this will close the report, here the user is speaking about a high load with using f5.
Sadly (happily) this i can't reproduce.
And he is using 1.14 which is pretty obsolete, and can only fixed by downstream maintainer with patching 1.14 if someone found a solution.
But i am really in doubt that will ever happens.
Sadly linuxmint users don't understand that only the distro itself can update old releases.
We have a lot of issues about old and obsolete mate packages.

raveit65:
Btw. report was filed out without using the template and filing out a downstream report.
As i am not willing any more to do the job of linuxmint to track down issues before users file out them here, it's save to close. ;-)

sc0w:
@raveit65 I can reproduce it with F5, with this steps:

- change the background image

- remove/rename the file

- go to the desktop, press F5, and see the CPU

so, for me, this PR fixes the issue :)

Issue #698, opened 2017-01-02
[List View] SHIFT+ENTER When creating new documents results in default names


epgfm:
When creating a new document using right click -> create document -> [Empty File or template] the new document is created and focus is immediately on the rename state, which allows one to immediately enter the new name.
However when one holds shift when pressing enter after writing the new name, what was entered is discarded and the default name is used instead.
This happens a lot to me when I create files such as README.TXT, would it be possible to change the code so that SHIFT+ENTER has the same behavior as ENTER ?
Edit: using Ubuntu MATE 16.04.1 with MATE 1.12.1

raveit65:
Hmm,
**if** it is possible i am in worry that other user blame us ' What you did with shift+ enter'.
Why do you press the shift button when you don't want to press this button?

epgfm:
I'm not doing it on purpose, it's just when I write a filename in allcaps, I almost always hit enter before  remembering I have to release the shift key. I'm not sure if Shift+Enter has any other uses when the rename mode is active.

kn00tcn:
i think there should be a whitelist of what escapes or cancels the text input... probably only the escape key?

for accepting text input, enter while ignoring anything else pressed at the same time as enter, & clicking out onto something else should also accept (unless it's to drag the caja window if that's possible, not sure if it depends on the window manager/decorator)

monsta:
@sc0w hmm I can't reproduce it with current master and 1.18 branches... maybe it was fixed as a side effect of https://github.com/mate-desktop/caja/pull/853?

sc0w:
@monsta 

I can't reproduce this issue with caja 1.18.4 (without https://github.com/mate-desktop/caja/pull/853)

monsta:
Ok, I tested some older releases as well. So I can't reproduce it even in 1.12.7 which is in Ubuntu 16.04 repos.

@epgfm can you still reproduce this?

sc0w:
...and I can't reproduce the issue with caja 1.8.2 and 1.16.6

epgfm:
I can still reproduce the issue in 1.18.4-0ubuntu2 (Ubuntu MATE 17.10), though the behaviour evolved somewhat. Now on top of the file falling back to default name, the new file is also opened.
I'm not sure if this package version comes without #853.

monsta:
This is even more weird - the new file shouldn't be auto-opened...

sc0w:
[...]

PR #868, opened 2017-11-04
Fix: wrong behavior with key "enter" while renaming files in list view

sc0w:
Fixes https://github.com/mate-desktop/caja/issues/698

### DIFF
--- a/src/file-manager/fm-list-view.c
+++ b/src/file-manager/fm-list-view.c
@@ -1100,6 +1100,7 @@ static gboolean
 key_press_callback (GtkWidget *widget, GdkEventKey *event, gpointer callback_data)
 {
     FMDirectoryView *view;
+    FMListView *listview;
     GdkEventButton button_event = { 0 };
     gboolean handled;
     GtkTreeView *tree_view;
@@ -1108,6 +1109,7 @@ key_press_callback (GtkWidget *widget, GdkEventKey *event, gpointer callback_dat
     tree_view = GTK_TREE_VIEW (widget);
 
     view = FM_DIRECTORY_VIEW (callback_data);
+    listview = FM_LIST_VIEW (view);
     handled = FALSE;
 
     switch (event->keyval)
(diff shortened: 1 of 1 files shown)


raveit65:
@monsta 
Sounds like this belongs in a ne caja 1.18 release?
... so i am waiting with a new tarball.

raveit65:
@monsta 
For me this PR belongs in a new 1.18 point release, so i am waiting for your review.

monsta:
Shift+Enter indeed does nothing now, but I'm pretty sure @epgfm meant it to act as just Enter.

sc0w:
@monsta yes, I know, but I tried and I don't know how to do that

anyway, this PR fixes the wrong behavior

monsta:
I'll try to find a way...

monsta:
Looks like it happens because GtkTreeView does nothing on Shift-Enter. Tested in sensors prefs dialog in mate-sensors-applet.

sc0w:
commit rebased

@raveit65 @lukefromdc @monsta 

please test again

now [shift+enter] and [control+enter] works like [enter]

lukefromdc:
Just retested this, works fine here too: ctrl and shift keys ignored when enter pressed.

monsta:
Very nice, it works 🙂 

One remark though: just in case, it's better to only remove the `GDK_SHIFT_MASK` or `GDK_CONTROL_MASK` bit, rather than set the whole `state` to zero. There are [lots of these modifiers](https://developer.gnome.org/gdk3/stable/gdk3-Windows.html#GdkModifierType).

src/file-manager/fm-list-view.c
```diff
@@ -1168,17 +1170,26 @@ key_press_callback (GtkWidget *widget, GdkEventKey *event, gpointer callback_dat
...
-        {
-            activate_selected_items_alternate (FM_LIST_VIEW (view), NULL, TRUE);
-        }
-        else
-        {
-            activate_selected_items (FM_LIST_VIEW (view));
-        }
-        handled = TRUE;
+	if (GTK_IS_CELL_EDITABLE (listview->details->editable_widget) &&
+	    (event->state & GDK_SHIFT_MASK) || (event->state & GDK_CONTROL_MASK))
```
monsta: Let's add one more pair of parentheses around this, as `||` has lower priority than `&&`.
Otherwise using Ctrl-Shift-Enter when **not** editing is completely ignored, while in git master it behaves like Shift-Enter.

I know it sounds minor, but let's keep the existing behavior, especially when the change is meant to go into 1.18 branch as well.

sc0w:
@monsta 

> One remark though: just in case, it's better to only remove the GDK_SHIFT_MASK or GDK_CONTROL_MASK bit, rather than set the whole state to zero.

I don't agree, just tell me what is the expected behavior and the actual behavior with this PR, it works like the other views, for example, [alt+shift+enter] and [alt+control+enter] works like [enter] with icon, compact, and list view with this PR

> Let's add one more pair of parentheses around this

yes, I forgot that, done!

monsta:
Ok, I might handle it later, for now let's merge it

Issue #859, opened 2017-10-19
Open in new window shortcut duplicated


riksoft:
#### Expected behaviour
SHIFT+CTRL+O should open a new window

#### Actual behaviour
It opens a new tab because the shortcut is the same

#### Steps to reproduce the behaviour
Put the focus on a folder and press SHIFT+CTRL+O

#### MATE general version
1.16.2-1+serena and 1.18.0-1+sonya

#### Package version
Caja 1.16.6-1 of mint 18.1 and Caja 1.18.3-1 on Mint 18.2

#### Linux Distribution
Mint 18.1. Same problem on Mint 18.2 (On 18.2 there is another strange behaviour: shift must be pressed before ctrl. On mint 18.1 doesn't matter the order.

#### Link to downstream report of your Distribution
I have no idea what it is the "downstream report"... peeping at the other reports I reckon  I'm not the only one :) )

#### Suggestion
Would be VERY nice and faster to have a new window opened with CTRL + double click.

raveit65:
> Link to downstream report of your Distribution

> I have no idea what it is the "downstream report"... peeping at the other reports I reckon I'm not the only one :) )

We like to see distro maintainers involved to be sure it isn't a distro problem, and 'shift must be pressed before ctrl' is a distro problem'.

So don't file first a report against your distro if you like that someone answer on your report here fast ;-)

monsta:
Fixed in 1.18.5. A different shortcut was used due to reasons explained in https://github.com/mate-desktop/caja/pull/869.

riksoft:
Thanks a lot!

PR #869, opened 2017-11-05
open folder in new window: [ctrl] + [shift] + [w]

sc0w:
Fixes https://github.com/mate-desktop/caja/issues/859

In the issue is suggested [ctrl+shift+o] to open new window

Currently [ctrl+shift+o] works to open in new tab, and it happens since years ago, so, I suppose some users prefer to keep it with no changes

I suggest the change in "open new window" to [ctrl+shift+w], it can be done faster, with one hand


monsta:
It's a simple fix, so it works. We have Ctrl-W to close the current tab though (or the whole window if there's only one tab). You can accidentally miss Shift and lose a tab 🙂 

If that's ok for you guys, merge it.

raveit65:
I assume this is a bug fix so i will cherry-pick it to 1.18 branch.

raveit65:
done!

PR #870, opened 2017-11-19
remove the shadows

eyelash:
This is obviously a very controversial change but IMHO makes caja look much more modern. You might want to do a poll with the community or so before merging this. Please note that also before these commits some of the sidebar tabs did not have a shadow (e.g. the information tab).
Anyway here are some screenshots with the default Ubuntu Ambiance theme:
before:
caja-before
after:
caja-after

### DIFF
--- a/src/caja-history-sidebar.c
+++ b/src/caja-history-sidebar.c
@@ -274,7 +274,6 @@ caja_history_sidebar_init (CajaHistorySidebar *sidebar)
                                     GTK_POLICY_AUTOMATIC);
     gtk_scrolled_window_set_hadjustment (GTK_SCROLLED_WINDOW (sidebar), NULL);
     gtk_scrolled_window_set_vadjustment (GTK_SCROLLED_WINDOW (sidebar), NULL);
-    gtk_scrolled_window_set_shadow_type (GTK_SCROLLED_WINDOW (sidebar), GTK_SHADOW_IN);
     gtk_container_add (GTK_CONTAINER (sidebar), GTK_WIDGET (tree_view));
     gtk_widget_show (GTK_WIDGET (sidebar));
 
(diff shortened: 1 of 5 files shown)


flexiondotorg:
Please could you post before and after screenshots using the Menta and Ambiant-MATE themes?

eyelash:
Ambiant-MATE before:
Ambiant-MATE-before
Ambiant-MATE after:
Ambiant-MATE-after

eyelash:
Menta before:
menta-before
Menta after:
menta-after

flexiondotorg:
Thanks for doing that. IMO those screenshots using Menta and Ambiant-MATE look better in the before images. Sorry. The Ambiance theme has not been adapted for MATE, so I think that is the actual underlying issue. @raveit65 Your thoughts?

eyelash:
I somewhat agree that for the MATE themes it looks better with shadows. I wonder if the shadows could be added in theme instead of the application. I believe the MATE applications should look good not only with the MATE themes.

raveit65:
Hmm, personal i like the design of the shadows.
And why you define using shadows as non-modern.
For my taste this is very modern :-)
.. for me this goes in design direction of Adwaita or there plagiarism Arc theme, and i don't like them very much.
Any ways, it is very simple to remove the shadow in gtk theme itself, or you can set up this in your user conf.
~.config/gtk-3.0/gtk.css
This overrides all themes which are installed at your box.
I can help you to find the setting or simply fire up gtk3-inspector on caja.

eyelash:
I call it modern because many GNOME 3 apps have removed their shadows and many new GTK 3 themes are primarily made for GNOME. But the definition of what is modern is certainly subjective :smiley:. I guess I close this now. Feel free to reopen if you still want to merge these changes.

raveit65:
@eyelash 
```
[rave@mother ~]$ cat /home/rave/.config/gtk-3.0/gtk.css
.caja-notebook scrolledwindow.frame,
.caja-side-pane notebook scrolledwindow.frame {
    border-style: none;
}
```
Works with most themes here.

raveit65:
Don't forget to restart caja or session.

lukefromdc:
I think as a rule we should not hardcode theme elements except where it is actually necessary to fix a problem. Artistic and theme style elements should remain under the control of the themes so that they remain controllable by end users.  These particular shadows don't show up in my theme for instance because the borders they are set on are not used, though by no means is UbuntuStudio_Legacy a "flat" theme.   Hardcoding style elements like this also means having to revisit that code if user tastes change and what is hardcoded becomes seen as "dated."

PR #878, opened 2017-12-04
WidthOfScreen and HeightOfScreen implementation

sc0w:
This commit reverts:

https://github.com/mate-desktop/caja/commit/4c130e22b98bbc1a85747adac3b582e45d34514e
https://github.com/mate-desktop/caja/commit/ac12e6d16f4a51b3327d013781e72cfec17f3624
https://github.com/mate-desktop/caja/commit/17b3c0ca924ffcd4bd06fded98227eba626d4f0b
https://github.com/mate-desktop/caja/commit/42c8d2c3373c25d6e321be6950997c0d7ce44279

And it applies an alternative to fix the deprecated functions:

gdk_screen_get_width
gdk_screen_get_height

gdk_screen_width
gdk_screen_height

In the tests, make sure you can't reproduce the issues: https://github.com/mate-desktop/caja/pull/824#issuecomment-323568583 and https://github.com/mate-desktop/caja/issues/836

### DIFF
--- a/eel/eel-background.c
+++ b/eel/eel-background.c
@@ -31,7 +31,7 @@
 #include "eel-lib-self-check-functions.h"
 #include <gtk/gtk.h>
 #include <eel/eel-canvas.h>
-# include <cairo-xlib.h>
+#include <cairo-xlib.h>
 #include <gdk/gdkx.h>
 #include <gio/gio.h>
 #include <math.h>
@@ -311,33 +311,9 @@ drawable_get_adjusted_size (EelBackground *self,
 {
     if (self->details->is_desktop)
     {
-#if GTK_CHECK_VERSION (3, 22, 0)
-        GdkRectangle geometry;
-        GdkMonitor  *monitor;
-        GdkDisplay  *display;
-        int n = 0;
-        int i = 0;
-        int sc_width = 0;
-        int sc_height = 0;
-
-        display = gtk_widget_get_display (self->details->widget);
-        n = gdk_display_get_n_monitors (display);
-
-        for (i = 0; i < n; ++i)
-        {
-            monitor = gdk_display_get_monitor (display, i);
-            gdk_monitor_get_geometry (monitor, &geometry);
-            sc_width = sc_width + geometry.width;
-            sc_height = sc_height + geometry.height;
-        }
-
-        *width = sc_width;
-        *height = sc_height;
-#else
         GdkScreen *screen = gtk_widget_get_screen (self->details->widget);
-        *width = gdk_screen_get_width (screen);
-        *height = gdk_screen_get_height (screen);
-#endif
+        *width = WidthOfScreen (gdk_x11_screen_get_xscreen (screen));
+        *height = HeightOfScreen (gdk_x11_screen_get_xscreen (screen));
     }
     else
     {
--- a/eel/eel-editable-label.c
+++ b/eel/eel-editable-label.c
@@ -36,6 +36,7 @@
 #include <pango/pango.h>
 #include <gtk/gtk.h>
 #include <gtk/gtk-a11y.h>
+#include <gdk/gdkx.h>
 #include <gdk/gdkkeysyms.h>
 
 
@@ -1051,7 +1052,6 @@ eel_editable_label_ensure_layout (EelEditableLabel *label,
             else
             {
                 gint wrap_width;
-                gint sc_width;
 
                 pango_layout_set_width (label->layout, -1);
                 pango_layout_get_extents (label->layout, NULL, &logical_rect);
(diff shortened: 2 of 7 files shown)


lukefromdc:
WidthOfScreen and HeightOfScreen apear on a quick Startpage search to be xlib macros
https://linux.die.net/man/3/widthofscreen
https://tronche.com/gui/x/xlib/display/screen-information.html
So does this give us another dependency on x11 that has to change or get a Wayland equivalent if we ever support Wayland directly? At the moment everybody still uses Xwayland for the desktop, as GNOME is not interested in desktop icon support so has not ported nautilus-desktop over to direct wayland use

sc0w:
@lukefromdc 

> So does this give us another dependency on x11 that has to change or get a Wayland equivalent if we ever support Wayland directly?

I don't know, I never used wayland

raveit65:
> WidthOfScreen and HeightOfScreen apear on a quick Startpage search to be xlib macros
 https://linux.die.net/man/3/widthofscreen
 https://tronche.com/gui/x/xlib/display/screen-information.html
 So does this give us another dependency on x11 that has to change or get a Wayland equivalent if we ever support Wayland directly? At the moment everybody still uses Xwayland for the desktop, as GNOME is not interested in desktop icon support so has not ported nautilus-desktop over to direct wayland use

> EDIT: if desktops stay x11 or xwayland only, this is a more elegant approach than before. Same if Wayland has or gets similar macros that could be runtime selected

I think this should be discussed here as this is the first PR of some followers in other projects.

@monsta @flexiondotorg @XRevan86 @clefebvre 
Thoughts?

Personal i think supporting Wayland or Xwayland is far away in future for us, so i agree if this is a more elegant and better solution.

monsta:
The first piece of commit seems to discard multi-monitor information, is that ok?

sc0w:
@monsta yes, is ok if you can't reproduce the issue https://github.com/mate-desktop/caja/issues/836

@raveit65  confirmed it in https://github.com/mate-desktop/caja/pull/878#pullrequestreview-81503463

monsta:
I don't have a second monitor to try that...

raveit65:
> I don't have a second monitor to try that...

I used the monitor from my main box and plugged them in my notebook.
Any way, the issue didn't come back here.
@monsta 
My question was an other one.
This PR and followers in other projects is a step back to support real Wayland,  not XWayland.
In this way i understand https://github.com/mate-desktop/caja/pull/878#issuecomment-349748233
It's more an theoretical question for future as there is no distro or DE which use Wayland at the moment.
So, do we want to go in this direction?

monsta:
Yes, from the point of the future porting to Wayland, such changes will just give us more work.
Do these PRs also fix some issues? If not, then maybe we shouldn't merge them.

sc0w:
@monsta 

> Do these PRs also fix some issues? If not, then maybe we shouldn't merge them.

Why not if this is a code improvement?

Question: currently wayland works in mate?

- If it doesn't work, I don't understand the wayland concerns

- If it works, feel free to test this PR with wayland and I will try to change

I noticed some problems with the previous fix to the deprecated functions, and my intention with all the PRs related is to prevent more problems

raveit65:
As we are a very small active team and some (core) team member don't react on emails any more, i am really i doubt that we ever use wayland.
** MATE + Wayland = disko 3000 **  :)
Also, it should be easy to revert those commits if there are some devs who want to make Wayland for Mate happen.

lukefromdc:
If we are unlikely to support Wayland anyway (it's on the roadmap but capacity said to be an issue), we can go ahead and merge these as it should make maintainance easier. If we do go to wayland, we will need to write or modify a wayland compositor, and it's in that compositor that the equivalent features would live and could even be created with the same name.

raveit65:
Ok, as it is a code improvement, nobody is working for wayland support, and most of the team vote for it, i will merge it.
Than we can go forward and work on other open similar PRs.

raveit65:
merged https://github.com/mate-desktop/caja/commit/2fea2c846689d3810def93e6e0e6eeb01c2ef718

clefebvre:
I'm not mad about this change personally. I'd rather stick with common toolkit and utility libs, rather than direct X11 calls.

lukefromdc:
I did find that these applications work in xwayland if they worked before this change. Some worked in wayland but many did not without xwayland prior to the commit to force X11 as the backend.  If MATE ever does convert to wayland, there would be the option to reimplement this from whatever the native Wayland calls are, possibly in mate-desktop and a runtime check would be needed. Alternately whatever wayland compositor was used might have similar functions that could be selected by runtime check if running under wayland.

I didn't think this was the best solution either but we have the reality of a small team, this might be easier to maintain, and some said they deemed a wayland port unlikely altogether due to the current size of the team. At any rate it would be a long time before xwayland goes away, if ever.

clefebvre:
Does this change affect HiDPI?

I know it's not implemented yet in MATE but it's coming... now that GTK3 is adopted we should be able to do it.

I was under the impression the scaling was set at toolkit/Gdk level and that X11 didn't really know about it. If you have an HiDPI branch somewhere, this is worth a test.

It might be OK, maybe I'm worrying for nothing here, but it crossed my mind so I thought I'd mention it.

raveit65:
> Does this change affect HiDPI?
> I know it's not implemented yet in MATE but it's coming... now that GTK3 is adopted we should be able to do it.
> I was under the impression the scaling was set at toolkit/Gdk level and that X11 didn't really know about it. If you have an HiDPI branch somewhere, this is worth a test.

You can forward your donations to MATE devs, then we can buy hdpi monitors and test this ;-)

Issue #880, opened 2017-12-11
Can't eject removable media without loading listing


unkilbeeg:
#### Expected behaviour
Click on sidebar eject button next to media icon and media is unmounted

#### Actual behaviour
Click on sidebar eject button next to media icon and the current view in the listing pane is replaced with the listing of that particular media.  A second click on the eject button will eject the media.

#### Steps to reproduce the behaviour
Launch  Caja with the file listing on your home directory (or any directory, really.)  Plug in usb drive or CD or any other removable medium.  Click on the eject button next to the icon for your removable medium and instead of ejecting, the listing changes to that of the medium you're trying to eject.  Normally the listing already showing is something I'm using -- I don't want it to shift just because I've ejected something else.

I did a search and ran across the following fixed bug:
https://github.com/mate-desktop/caja/issues/472?_pjax=%23js-repo-pjax-container
This bug involved ejecting any time you selected the icon of the mounted medium.  My problem is the opposite -- I want a portion of that field to be available to immediately eject. It looks like both issues come from the entire field being treated as one entity -- clicking on it used to cause an immediate eject, and now it causes an immediate load of the file pane.  I'd like it to work where the right side (the eject button) works to eject the medium and the left side (the icon for the filesystem) acts to load a file listing. 

#### MATE general version
Package: mate-common
Version: 1.18.0-1+sonya

#### Package version
Package: caja
Version: 1.18.5-1+sonya

#### Linux Distribution
Linux Mint 18.2

lukefromdc:
No matter how I fiddle with the coding for the over_eject_button function in caja-places-sidebar.c, I will always get either an occasional unwanted unmount or an occasional unwanted open rather than unmount. It looks like I can make this a bit less touchy about being over the eject button itself but I think there is a race condition between something else and over_eject_button returning.

Sometimes Caja will even first go to the root directory of the volume being unmounted, then unmount it. The code in Nemo works, but has a long string of changes and a simple attempt to apply the bulk of the changes for position sensing
https://github.com/GNOME/nautilus/commit/f294a4e80626793624d76d6002b094d1c0809de4
led to unconditional mount./umount no matter where in the drive's row the mouse was clicked. Will investigate further.

[...]

PR #881, opened 2017-12-12
Fix refusal to unmount when sidebar is wide

lukefromdc:
Fix refusal to unmount volumes from the sidebar when the sidebar is wide enough to expand the eject column.
Fixes https://github.com/mate-desktop/caja/issues/880

### DIFF
--- a/src/caja-places-sidebar.c
+++ b/src/caja-places-sidebar.c
@@ -265,6 +265,7 @@ get_eject_icon (CajaPlacesSidebar *sidebar,
         icon_set = gtk_style_context_lookup_icon_set (style, GTK_STOCK_MISSING_IMAGE);
         eject = gtk_icon_set_render_icon_pixbuf (icon_set, style, GTK_ICON_SIZE_MENU);
     }
+        g_object_unref (icon);
         gtk_icon_info_free (icon_info);
 
     return eject;
--- a/src/caja-places-sidebar.c
+++ b/src/caja-places-sidebar.c
@@ -232,16 +232,26 @@ G_DEFINE_TYPE_WITH_CODE (CajaPlacesSidebarProvider, caja_places_sidebar_provider
                                  sidebar_provider_iface_init));
 
 static GdkPixbuf *
-get_eject_icon (gboolean highlighted)
+get_eject_icon (CajaPlacesSidebar *sidebar,
+                gboolean highlighted)
 {
     GdkPixbuf *eject;
-    CajaIconInfo *eject_icon_info;
+    GtkIconInfo *icon_info;
+    GIcon *icon;
     int icon_size;
+    GtkIconTheme *icon_theme;
+    GtkStyleContext *style;
 
-    icon_size = caja_get_icon_size_for_stock_size (GTK_ICON_SIZE_MENU);
+    icon_theme = gtk_icon_theme_get_default ();
+    icon_size = caja_get_icon_size_for_stock_size (GTK_ICON_SIZE_SMALL_TOOLBAR);
+    icon = g_themed_icon_new_with_default_fallbacks ("media-eject-symbolic");
+    icon_info = gtk_icon_theme_lookup_by_gicon (icon_theme, icon, icon_size, 0);
 
-    eject_icon_info = caja_icon_info_lookup_from_name ("media-eject", icon_size);
-    eject = caja_icon_info_get_pixbuf_at_size (eject_icon_info, icon_size);
+    style = gtk_widget_get_style_context (GTK_WIDGET (sidebar));
+    eject = gtk_icon_info_load_symbolic_for_context (icon_info,
+                                                     style,
+                                                     NULL,
+                                                     NULL);
 
     if (highlighted) {
         GdkPixbuf *high;
@@ -250,7 +260,8 @@ get_eject_icon (gboolean highlighted)
         eject = high;
     }
 
-    g_object_unref (eject_icon_info);
+    g_object_unref (icon);
+    gtk_icon_info_free (icon_info);
 
     return eject;
 }
@@ -375,7 +386,7 @@ add_place (CajaPlacesSidebar *sidebar,
     }
 
     if (show_eject_button) {
-        eject = get_eject_icon (FALSE);
+        eject = get_eject_icon (sidebar, FALSE);
     } else {
         eject = NULL;
     }
@@ -1022,6 +1033,9 @@ over_eject_button (CajaPlacesSidebar *sidebar,
         eject_button_size = caja_get_icon_size_for_stock_size (GTK_ICON_SIZE_MENU);
 
         if (x - total_width >= 0 &&
+            /* fix unwanted unmount requests if clicking on the label */
+            x >= total_width - eject_button_size &&
+            x >= 80 &&
             x - total_width <= eject_button_size) {
             return TRUE;
         }
@@ -2897,7 +2911,7 @@ update_eject_buttons (CajaPlacesSidebar *sidebar,
 
             gtk_list_store_set (sidebar->store,
                         &iter,
-                        PLACES_SIDEBAR_COLUMN_EJECT_ICON, get_eject_icon (FALSE),
+                        PLACES_SIDEBAR_COLUMN_EJECT_ICON, get_eject_icon (sidebar, FALSE),
                         -1);
             gtk_tree_model_filter_refilter (GTK_TREE_MODEL_FILTER (sidebar->filter_model));
 
@@ -2919,7 +2933,7 @@ update_eject_buttons (CajaPlacesSidebar *sidebar,
                      path);
         gtk_list_store_set (sidebar->store,
                     &iter,
-                    PLACES_SIDEBAR_COLUMN_EJECT_ICON, get_eject_icon (TRUE),
+                    PLACES_SIDEBAR_COLUMN_EJECT_ICON, get_eject_icon (sidebar, TRUE),
                     -1);
         gtk_tree_model_filter_refilter (GTK_TREE_MODEL_FILTER (sidebar->filter_model));
 
--- a/src/caja-places-sidebar.c
+++ b/src/caja-places-sidebar.c
@@ -1022,15 +1022,9 @@ over_eject_button (CajaPlacesSidebar *sidebar,
         eject_button_size = caja_get_icon_size_for_stock_size (GTK_ICON_SIZE_MENU);
 
         if (x - total_width >= 0 &&
-            /* fix unwanted unmount requests if clicking on the label */
-            x >= total_width - eject_button_size &&
-            x >= 80 &&
             x - total_width <= eject_button_size) {
             return TRUE;
         }
-        /* Fix refusal to unmount when sidebar is wide enough to expand the eject column */
-        if (x >= 180)
-        return TRUE;
     }
 
 out:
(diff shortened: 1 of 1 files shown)


lukefromdc:
Just updated this to use the same code that works well in Nautilus and Nemo minus the use of a symbolic icon. Also included the bugfixes Nautilus pushed when they got a couple issues on first changing this code.

lukefromdc:
This can all be squashed for final commit

lukefromdc:
I'll find the commits where GNOME fixed those deprecations and apply them too

raveit65:
Works great and  it seems that the sidebar is a bit smaller.
Only if you start caja as root than the scrollbar is nearly over the whole eject icon.
Maybe it's possible to add a right-margin only for icon?
Doing that or adding a padding with the theme is really weird with the treeview.

lukefromdc:
The sidebar goes to whatever size you set it to. What is different is that the eject icon itself is always all the way to the right. Mounting one drive so the eject icon first appears does not make the sidebar resize, so it should not be any smaller on a change of Caja version. If it is somehow different, the sidebar is easily dragged back. Will take a look on my test user (default) user but have to log out to do so.

lukefromdc:
On a totally default, first run of Caja, I found an exactly 4px difference in the width of the sidebar if the startup was with one ejectable drive mounted by another user. Treeviews are very hard to work on because GtkInspector does not drill down into them, I have no idea how to add a 4px right pad. Do we want to block a bugfix over that? Any user can drag it back and it will stay where they set it.

raveit65:
At least it is only a problem with overlay-scrollbars which are disabled as default in normal user environment. As it is set via an env var, disabling does not work as root.
But luckily i was able to disable overlay-scrolling at CajaPlacesSidebar with gtk3-inspector.
So it should be possible to  use gtk_scrolled_window_set_overlay_scrolling () to disable this  in code in general.
https://developer.gnome.org/gtk3/stable/GtkScrolledWindow.html#gtk-scrolled-window-set-overlay-scrolling
I don't think that anyone from team is sad about missing overlay-scrolling ;-)

lukefromdc:
Latest change disables overlay scrollbars in the places sidebar and pads the icon an addtional 4px away from the edge to match the previous width. Both Nemo and Nautilus share the problem of burying the eject icon under overlay scrollbars, so no fix from there. If I pad the icon enough to keep it out from under the wide scrollbar set by Menta it just plain takes up too much space when not using overlay scrollbars, so had to use the suggestion to hard disable overlay scrolling.

lukefromdc:
All but one of the new deprecations come from 
https://github.com/mate-desktop/caja/pull/881/commits/655557f9e9258a36c9aa12ddb7c8a4ee2ab27808

Which is the routine to handle a missing icon. Turns out that does not work, with or without that commit a missing icon does NOT crash Caja, at least with GTK 3.22, instead no icon is rendered and clicking on the empty space will eject the volume. I spent the last half hour trying to get that to work with the eject icon renamed, no success on that. I could rewrite that-or just revert the offending commit as the behavior on a missing icon is exactly the same and no crash results

lukefromdc:
https://github.com/mate-desktop/caja/pull/881/commits/d2c0bbefd13e28806f1b279f6d9462b2c1e5e8ee
fixes the four new deprecations and also fixes handling the missing icon case.  For some reason GNOME's code for handling the missing icon case just didn't work: It found the icon easily enough, and reversing the logic so the deprecated code ran used that icon to indicate a missing icon, but actually making the icon missing didn't return NULL for some reason and the if logic failed. Didn't crash, just rendered no icon at all. 

Returning to our origjnal code for actually finding and rendering the icon but leaving in place the rest of the GNOME changes to the first of their series of commits on this brought back the previous missing icon handling, removed the deprecated stuff-and still kept the rest of the changes working so the eject button works consistantly, does not get overlay scrollbars covering it up, etc.

raveit65:
I am using always 'git format-patch commit-id' to create patches for my rpms for testing. So squashing this before merging in less commits would be very helpfull.
Otherwise i have to that for my own for every new test :/

raveit65:
Btw. for my point of view we don't need the extra 4px padding if we disable overlay-scrolling.
But your decision, i am fine w/o using a padding.

lukefromdc:
Squashed, cleaned up and merged with 
https://github.com/mate-desktop/caja/commit/c6b81f64c4c4212e414c04e584380774d6ac83ca
note that removing a couple variables left unused by returning to the original eject icon lookup had the effect of removing all of a couple commits from the history, helping clean up the changelog as well.

raveit65:
@lukefromdc 
Weird, what you merged is not what i tested yesterday.
I merged all commits in one from my local repo and it looks different than yours
see https://gist.github.com/raveit65/f5127897350b731145b075f7556c6edb
Forget about the padding, but I am missing those parts:
```
@@ -232,25 +229,31 @@ G_DEFINE_TYPE_WITH_CODE (CajaPlacesSidebarProvider, caja_places_sidebar_provider
                                  sidebar_provider_iface_init));
 
 static GdkPixbuf *
-get_eject_icon (gboolean highlighted)
+get_eject_icon (CajaPlacesSidebar *sidebar,
+                gboolean highlighted)
 {
     GdkPixbuf *eject;
-    CajaIconInfo *eject_icon_info;
+    CajaIconInfo *icon_info;
     int icon_size;
+    GtkIconTheme *icon_theme;
+    GtkStyleContext *style;
 
+    icon_theme = gtk_icon_theme_get_default ();
     icon_size = caja_get_icon_size_for_stock_size (GTK_ICON_SIZE_MENU);
+    icon_info = caja_icon_info_lookup_from_name ("media-eject", icon_size);
 
-    eject_icon_info = caja_icon_info_lookup_from_name ("media-eject", icon_size);
-    eject = caja_icon_info_get_pixbuf_at_size (eject_icon_info, icon_size);
+    style = gtk_widget_get_style_context (GTK_WIDGET (sidebar));
+
+    eject = caja_icon_info_get_pixbuf_at_size (icon_info, icon_size);
 
     if (highlighted) {
         GdkPixbuf *high;
         high = eel_create_spotlight_pixbuf (eject);
         g_object_unref (eject);
         eject = high;
-    }
+        }
 
-    g_object_unref (eject_icon_info);
+    g_object_unref (icon_info);
 
     return eject;
 }
@@ -375,7 +378,7 @@ add_place (CajaPlacesSidebar *sidebar,
     }
 
     if (show_eject_button) {
-        eject = get_eject_icon (FALSE);
+        eject = get_eject_icon (sidebar, FALSE);
     } else {
         eject = NULL;
     }
```
```
@@ -2900,7 +2883,7 @@ update_eject_buttons (CajaPlacesSidebar *sidebar,
 
             gtk_list_store_set (sidebar->store,
                         &iter,
-                        PLACES_SIDEBAR_COLUMN_EJECT_ICON, get_eject_icon (FALSE),
+                        PLACES_SIDEBAR_COLUMN_EJECT_ICON, get_eject_icon (sidebar, FALSE),
                         -1);
             gtk_tree_model_filter_refilter (GTK_TREE_MODEL_FILTER (sidebar->filter_model));
 
@@ -2922,7 +2905,7 @@ update_eject_buttons (CajaPlacesSidebar *sidebar,
                      path);
         gtk_list_store_set (sidebar->store,
                     &iter,
-                    PLACES_SIDEBAR_COLUMN_EJECT_ICON, get_eject_icon (TRUE),
+                    PLACES_SIDEBAR_COLUMN_EJECT_ICON, get_eject_icon (sidebar, TRUE),
                     -1);
         gtk_tree_model_filter_refilter (GTK_TREE_MODEL_FILTER (sidebar->filter_model));
 
```
Or do i miss something?

lukefromdc:
I did a final cleanup just before merge, removing a couple variables that has become unused when I reverted a GNOME change that broke rendering the fallback eject icon, and returning to the variable name eject_icon_info rather than icon_info for consistancy with the prior code. None of this should change behavior.

I did not push the squashed final changes to the remote branch (have them locally) prior to merge. The diff between that and what is merged is the missing variable cleanup and reverting the rename of one variable. After I squashed, the resulting diff as viewed in git-gui caused the unused variables to immediately stand out, and I figured merging without removing them would have been bad practice, thus the cleanup and a final retest locally.

If you want me to to revert this and reopen the PR for further testing that is simple enough to do.

lukefromdc:
I just committed the final pre-merge cleanup back to the unsquashed fix-eject branch here so the exact diff could be seen. This is the same file as was merged at this point and shows the unused variables and the change of padding back to 6px.

https://github.com/mate-desktop/caja/commit/22a0733a920bb076d820f7e20788819c7e947ef2

raveit65:
In the meanwhile i retested the merged commit. Everything is fine :-)
I was only wondering about the last minute changes.
Only commit messages is a bit missleading.
```
caja-places-sidebar: revert all kludges for eject behavior: Revert
https://github.com/mate-desktop/caja/commit/d14fef384a435471860130f7337d5dea5bb824a0
https://github.com/mate-desktop/caja/commit/9b2e152876067aaee3baa699c51c6c6174d4954e
and
https://github.com/mate-desktop/caja/commit/30f19d0aed3e40a3682f820059ebb8b3038ba84e
```
Second link was only in PR and code from third link is in merged commit.
But no reason to revert it.
All is fine and thanks for the hard work.

raveit65:
Btw. i think we can disable overlay-scrollbars for views too, to have same behaviour.

lukefromdc:
That raises an interesting question: We already disable overlay scrollbars by default in a mate-session, though a user can override that in dconf-editor. Doesn't work when called as root (and thus ignoring the session config on this) without special hacks though. This is true of all MATE applications such as Pluma etc. Do we want to go through all of MATE to disable overlay scrollbars unconditionally (with no user override) or to only do this when there is a functional problem with the overlay scrollbars as in the places sidebar?

Keep in mind that the overlay scrollbars blocks eject icons in both Nemo and Nautilus, GNOME never fixed that problem.  If we disable them unconditionally in the caja views, that  leaves Caja inconsistant with all other apps if a user sets overlay scrollbars. As it was we either had to pad the heck out of the eject icon (wasting a lot of space with normal scrollbars) or make the sidebar inconsistant with the view when the user selects overlay scrollbars. We can leave that as it is, make Caja consistant with itself but nothing else, or make all of MATE consistant with itself but nothing else when a user enables overlay scrollbars.

GNOME probably allows disabling in the code specifically for cases such as this places-sidebar where it creates functional issues.

I am happy with this either way, what do others here want to do?

lukefromdc:
The 3ed commit mentioned in the changelog was only reverted because it was a subset of what was about to be applied later, and had been first used by itself. Yes, this got confusing as the road to getting the right mix of what GNOME did was rather complicated.

raveit65:
Ok, i agree with you not to disabling (hardcoding)overlay-scrollbar for whole the MATE desktop as it is inconsistant to other applications.
But i think we should set the same behaviour in side one application.
When we say we need to disable overlay-scrolling for sidebar we should do the same for the views.
@monsta 
thoughts?

monsta:
Hmm, I don't see 22a0733a920bb076d820f7e20788819c7e947ef2 anywhere in branches...

lukefromdc:
That's what happens in those themes that normally put the eject icon very close to the scrollbar. Since overlay scrollbars allocate no space, they partially or completely cover eject icons. Nautilus and Nemo have never fixed this. Adwaita just allocates more space to the right of the eject icons, so GNOME gets away with it. This does not look so good though when overlay scrolling is NOT used, and Adwaita in general is a theme that pads up everything rather large.

Issue #867, opened 2017-11-03
Search Performance Slow


oscardssmith:
#### Expected behaviour
When I search for python from the root of a file system it should take almost no time

#### Actual behaviour
It takes ~45 seconds and uses 100% cpu (only 1 core)

#### Steps to reproduce the behaviour
search for something from root

#### MATE general version
1.18.0

#### Package version
1.18.4

#### Linux Distribution
Ubuntu Mate 17.10

jhasse:
Would it make sense to use something as https://github.com/sharkdp/fd#benchmark as a backend?

Issue #849, opened 2017-09-14
search does not work with wildcards (*, ?)


caltrop-3:
#### Expected behaviour
use wildcards for search

#### Actual behaviour
does not recognize wildcards

#### Steps to reproduce the behaviour
use it with wild cards

#### MATE general version
1.18.0

#### Package version
1.18.3

#### Linux Distribution
Linux Mint 18.2 Sonya
Kernal Linux 4.10.0-33-generic x86_64

#### Link to downstream report of your Distribution
unknown

chexmo:
This still happens. Using Caja in Ubuntu MATE 19.10.

smithtim:
Confirmed that this still does not work as of caja 1.24.0.

As a partial workaround, if you just want to find all files, you can search with `.` and that will return all files that have an extension.

Issue #467, opened 2015-09-27
Caja closes when the last window closes.


maandree:
Please add an option to revert to the old behaviour: Caja stays alive until it is manually killed,
not until the last window closes.

monsta:
That behavior is still there. Of course, that's if you use Caja in MATE and not as root. :smile:

maandree:
But I don't want to use MATE.

maandree:
I guess it is due time to try get MATE 1.4 running.

raveit65:
why not gnome-2

maandree:
Because MATE 1.4 was better. MATE 1.4 was the best release of desktop environment yet.

PR #883, opened 2017-12-13
Add (back) tracker search support to caja.

darkshram:
Tracker 2.0 deprecated tracker-needle and tracker-preferences. This leaves without a graphical client for it any DE that hasn't integrated Tracker. Caja has code to use tracker for search. It's disabled because the needed macros are absent in configure.ac becuase most of the code is obsolete —even with an unconditioned include.

I have updated the tracker code to work with tracker 1.x and 2.x (borrowed some code from Nemo and Nautilus). I'm currently doing testing in 5 computers. Seems to works perfect, no crashes, no errors, no warnings. Unlike regular search, tracker searches show almost instant results.


jhasse:
Should fix #867, right?

Does this also fix #849?

darkshram:
@jhasse: */? are ignored by tracker, unless they are part of a filename. Tracker is meant to be used for desktop search, not for filesystem search. mate-search-tool does a decent job for that.

#467 is fixed if a tracker index exists. This means: tracker must be installed and tracker-miners must have been run at least once to create the indexes for search.

darkshram:
@jhasse 
vokoscreen-2017-12-13_17-05-13

raveit65:
I am not really happy that tracker tracked my box for no reason. And i don#t like to install tracker on my boxes.
Can we have a gesettings key which enable tracker and default is disabled?
And again, your spanish videos are useless for 3/4 of the world ;-)

darkshram:
I just commited with --enable-tracker=no (now default), originally commited with 'yes' (missed that one). Copied macros from nautilus and then just added tracker 2.0 support.

Would be pointless to build caja with --enable-tracker=yes and then have it disabled. But let think a good approach. Currently Caja has no setting to choose search backend on runtime, the only setting available to choose search engine it's done during build, same as Nautilus and Nemo do.

```
CajaSearchEngine *
caja_search_engine_new (void)
{
    CajaSearchEngine *engine;

#ifdef ENABLE_TRACKER
    engine = caja_search_engine_tracker_new ();
    if (engine)
    {
        return engine;
    }
#endif

    engine = caja_search_engine_beagle_new ();
    if (engine)
    {
        return engine;
    }

    engine = caja_search_engine_simple_new ();
    return engine;
}
```

So, Caja has support to use _caja_search_engine_tracker_new_, _caja_search_engine_beagle_new_ (dead project) and _caja_search_engine_simple_new_.

Nautilus removed beagle suport since 2010-12-03. Nemo was born without it. Beagle is basically dead since 2008, Fedora stopped packaging it since 2012. That's code that may be safely removed.

I'm currently working on forking tracker-settings and tracker-needle, but that's something that probably will happen until the next year.

darkshram:
@raveit65 I'll install a VM with english as default. Next videos will be in english. Promise.

jhasse:
Wouldn't it be possible to detect if Tracker is running at runtime?

darkshram:
@jhasse theorically yes. But that I think that would mean redesign/rewrite the whole search support of caja. Currently search engine is picked during build.

raveit65:
Thanks, that means with **--enable-tracker=yes** an installation of caja will require tracker libs as dependency.
Does **--enable-tracker=yes** install an extra file/lib for caja?
Than it should be possible to spilt tracker support to a sub-package for an distro maintainer and a user can decide for himself if he like to install the sub-package for tracker support.

darkshram:
@raveit65 Yes. Requires pkgconfig(tracker-sparql-2.0). It does not install any extra lib or file.

raveit65:
> I'm currently working on forking tracker-settings and tracker-needle, but that's something that probably will happen until the next year.

Why, tracker-settings open fine here in mate session of fedora ?

darkshram:
@raveit65 because tracker preferences an tracker-needle are gone with tracker 2.0. They deprecated it asuming DE already have a tool in their control-centers. GNOME3 and Cinnamon have it.

Current Fedora uses tracker 1.12, wihc still includes tracker-needle and tracker-prefrences. That's why I'm not in a hurry with the fork. When distros start shipping tracker 2.0, then it will be meaningful.

darkshram:
@raveit65 this is part of the changes between tracker 1.12 and 2.0:

>   Overview of changes since 1.12:
> 
>   * Tracker core (store, libraries) and miners have been split, this source
>     tree contains the former.
>   * Added support for domain ontologies, these can be used to set up private
>     databases managed and populated by a standalone set of daemons. It was
>     devised for sandboxing environments like Flatpak, but can be used
>     independently.
>   * Added API to manage application-private databases with (optional) custom
>     ontologies. The application is in charge of content population as well as
>     queries.
>     No daemons are required for this usecase, so this becomes a SPARQL
>     endpoint equivalent to SQLite.
>   * libtracker-miner objects can now be used on the application side in
>     combination with application-private endpoints. An additional object
>     has been added to make TrackerMiner implementations exported through
>     DBus.
>   * The filesystem TrackerMiner base implementation has received some
>     notable speedups.
>   * Running multiple concurrent SELECT queries is now lock-free on most
>     situations.
>   * Meson build instructions have been added.

The removal of tracker-needle and tracker-prefrences ocuren during 1.99.0 release:

> NEW in 1.99.0 - 2017-07-06
> ==========================
> 
>   WARNING: This is unstable development towards 2.0. There are API and
>            ABI incompatibilities that might affect you.
> 
>   * Domain ontologies: it is now possible to create domain-specific SPARQL
>     endpoints with customizable ontologies and data locations. It is possible
>     to do so either in-process using the traditional Tracker daemons to
>     do the work (Note: miners rely on Nepomuk ontology).
>   * **Deleted nautilus/firefox/evolution/thunderbird plugins**. They were in
>     a pitiful state of maintainership, the latter 2 being known broken for
>     some time already.
>   * **Deleted tracker-needle**. It has been in strict maintenance state for years,
>     since then other UIs have integrated Tracker search and offer a more
>     compelling and modern experience.
>   * **Deleted tracker-preferences**. Tracker has been using GSettings for years,
>     so doesn't specifically require an UI for it, there's of course room
>     for desktop integrated configuration, gnome-control-center does this
>     for example.
>   * Made portions of libtracker-miner useful on the client side. Miners don't
>     mandate anymore an org.freedesktop.Tracker1.Miner DBus interface/name,
>     this is instead delegated on TrackerMinerProxy. TrackerMinerFS has been
>     made largely independent of the ontology in use, all high level operations
>     that do require SPARQL updates are now exposed in TrackerMinerFSClass.
>   * Removed deprecated and duplicate API on libtracker-miner
>   * Added documentation about creating and maintaining ontologies.
>   * Removed maemo leftovers.
>   * libtracker-sparql: Protect sqlite3_bind* with interface lock
>   * Improvements to meson build

raveit65:
@darkshram 
You can use 
`git rebase -i last-good-commit`
for edit, squash, sort or reword a commit in your local repo. 
(i = interactive) After pushing the changes to github with `git push -f`
the PR will be updated automatic.

darkshram:
I have been testing a patch for simple-search engine. So far works acceptable. Can't expect too much from a search engine ot this type. Definitively has better performance than unpatched simple-search engine. Will PR after tracker support  PR. Meanwhile, if someone wants to test, here's the patch.
[caja-search-engine-simple-fix.patch.zip](https://github.com/mate-desktop/caja/files/1565730/caja-search-engine-simple-fix.patch.zip)

raveit65:
Ok, after adding MATE to OnlyShowIn from tracker desktop files searching with tracker works now.
But i run into another problem and iam pretty sure a lot of users will affected too if a distro maintainer switched to tracker.
Tracker indexing only the home folder with first run  and ihnore other partitions which are mounted somewhere. In my case ie.
/media/Jackass.
Without adding this dir to tracker config the search is complete broken at this folder and subfolders.

Another bigger problem for me that 'case insensitive' search is ignored , and i didn't found any tracker setting for this.
Sorry, i can't approve it like this.

raveit65:
Another problem with using tracker.
I 've symlink my rpmbuild folder in /home with same folder in dropbox to work easily on different boxes/installations.
If i search insid the symlink folder i found nothing. If i do the same search directly under /home/my-account/ i found files which a re in the symlink folders.
Sorry, i can't work like this, i need to drop this patch from my local build.

darkshram:
@raveit65 tracker-preferences (and gsettings) have a feature to index removable media and other mount points and other directories. By default, tracker only indexes Documents, Images, Videos and Desktop folders. You need to set aditional folders from tracker-prefrences or use gsettings.

No, tracker does not care for upper and lower case. It's designed for **Desktop Search**, not filesystem saearch. If the user needs a tool that features same things as 'find' tool, user should stick with simple search engine (current code **does not support** wildcards and follow symlinks, which make caja simple search unconsistent with find tool) or mate-search-tool.

pantallazo-2017-12-21 09-08-20

* Here you may set aditional folders to index
pantallazo-2017-12-21 09-08-29

pantallazo-2017-12-21 09-08-39

pantallazo-2017-12-21 09-08-45

raveit65:
Thanks for english images, that's a real progress :)
> No, tracker does not care for upper and lower case.

Hmm, in my github folder for all git projects.
searching for readme > no results
searching for README > gives me a really a lot results.

But the problem with with linked folders makes tracker unusable for me (personal+fedora maintainer).
https://github.com/mate-desktop/caja/pull/883#issuecomment-353336917
And using caja without the PR fixes the problem, i just re-tested it.

darkshram:
@raveit65 I have a patch for simple-search engine. I made it to be **consistent** with find tool. Supports wildcards. Do you want to test it? has _slightly_ better performance than current outdated code. I will pull it after this PR.
[caja-search-engine-simple-modernize-code.zip](https://github.com/mate-desktop/caja/files/1580618/caja-search-engine-simple-modernize-code.zip)

If you want it to follow symlinks, apply patch, then edit libcaja-private/caja-search-engine-simple.c and replace:

```
                        G_FILE_QUERY_INFO_NOFOLLOW_SYMLINKS,
                        data->cancellable, NULL);
```

With:

```
                        0, data->cancellable, NULL);
```

raveit65:
Normal simple-search  follow symlinks here.
I posted only things which do normal work here.
Don't understand me wrong , i am not against tracker support.
But i see so much issues comparing to normal search, that i can't enable it for fedora.
I am still wondering how gnome or cinnamon users can live with it.
At least that needs to be reviewed by other people which want to enable it for their distros.

darkshram:
@raveit65 Obviously tracker was designed to be used by final users as a '[desktop search](https://en.wikipedia.org/wiki/Desktop_search)' engine, not for 'filesystem search' or a find tool alternative/replacement.

In my case, I can live without caja-simple-search-engine because I **never** use caja-simple-search-engine because the poor performance. Find tool (and even mate-search-tool) do a better job for these kind of tasks.

Because the performance issues, GNOME removed recursive search from nautilus simple-search for to address the same issue caja has (see [https://mail.gnome.org/archives/commits-list/2016-August/msg08078.html](https://mail.gnome.org/archives/commits-list/2016-August/msg08078.html) and [https://bugzilla.gnome.org/show_bug.cgi?id=766174](https://bugzilla.gnome.org/show_bug.cgi?id=766174)). Yes, a find tool without recursiveness is 'not-very-useful', that's why they don't use simple-search and use tracker as default instead. Yes, advanced users complained about this, but simple-search engine is far from have a permanent solution for thew performance issue and most final users seem to be happy with an 'instant results' search enginee.

monsta:
I haven't looked into this yet, but I think we can already pick and merge the commit that drops Beagle support.

lukefromdc:
Any progress on this? Do we want to hold it open as a long term project, close it, or something else? Further work will require fixing the file conflicts that have build up over time

PR #891, opened 2017-12-31
Fix eject notifications

lukefromdc:
Prevent "it is now safe to remove the drive" notification from being shown if unmounting fails or is cancelled by the user(e.g when the drive is busy). Also remove function in src/caja-places-sidebar.c for sending the notification as the same function in libcaja-private/caja-file-operations.c already sends it.


lukefromdc:
This is important: wrongly telling a user a drive is safe to eject when it is busy could cause data loss.

lukefromdc:
What triggered this was I found that if I had a terminal open on a folder on a flash drive and tried to eject the disk, the disk busy cancel/unmount anyway dialog would come up, and choosing "cancel" would then show the "safe to remove" notification even though the drive remained mounted. 

A disk simply being written to held in correctly and showed the notification when and only when it was done as intended.

Issue #813, opened 2017-07-15
Places sidebar — show home dir when desktop disabled


TomaszGasior:
#### Expected behaviour
If "Desktop" directory path is set to user home directory, it means that Desktop directory is disabled. In this case CajaPlacesSidebar should not show "Desktop" element.

#### Actual behaviour
If "Desktop" directory is disabled, CajaPlacesSidebar hides home directory element and shows only "Desktop" button. This behavior should be inverted. 

#### Package version
Caja 1.18.3

TomaszGasior:
I prepared patch which fixes this bug.

```
--- /home/tomasz/caja/src/caja-1.19.3/src/caja-places-sidebar.c.org
+++ /home/tomasz/caja/src/caja-1.19.3/src/caja-places-sidebar.c
@@ -490,42 +490,42 @@
                              _("Computer"));
 
     /* add built in bookmarks */
-    desktop_path = caja_get_desktop_directory ();
 
     /* home folder */
-    if (strcmp (g_get_home_dir(), desktop_path) != 0) {
-        char *display_name;
-
-        mount_uri = caja_get_home_directory_uri ();
-        display_name = g_filename_display_basename (g_get_home_dir ());
-        icon = g_themed_icon_new (CAJA_ICON_HOME);
-        last_iter = add_place (sidebar, PLACES_BUILT_IN,
-                               SECTION_COMPUTER,
-                               display_name, icon,
-                               mount_uri, NULL, NULL, NULL, 0,
-                               _("Open your personal folder"));
-        g_object_unref (icon);
-        g_free (display_name);
-        compare_for_selection (sidebar,
-                               location, mount_uri, last_uri,
-                               &last_iter, &select_path);
-        g_free (mount_uri);
-    }
-
-    /* desktop */
-    mount_uri = g_filename_to_uri (desktop_path, NULL, NULL);
-    icon = g_themed_icon_new (CAJA_ICON_DESKTOP);
+    char *display_name;
+
+    mount_uri = caja_get_home_directory_uri ();
+    display_name = g_filename_display_basename (g_get_home_dir ());
+    icon = g_themed_icon_new (CAJA_ICON_HOME);
     last_iter = add_place (sidebar, PLACES_BUILT_IN,
                            SECTION_COMPUTER,
-                           _("Desktop"), icon,
+                           display_name, icon,
                            mount_uri, NULL, NULL, NULL, 0,
-                           _("Open the contents of your desktop in a folder"));
[...]

PR #892, opened 2017-12-31
places-sidebar: show user's home dir when it is also the desktop

lukefromdc:
Current behavior is to show Desktop but hide the user's home directory in the sidebar when desktop-is-home-directory is set. This is opposite what would be expected so invert it.

Patch by https://github.com/TomaszGasior

Fix https://github.com/mate-desktop/caja/issues/813

### DIFF
--- a/src/caja-places-sidebar.c
+++ b/src/caja-places-sidebar.c
@@ -490,42 +490,42 @@ update_places (CajaPlacesSidebar *sidebar)
                              _("Computer"));
 
     /* add built in bookmarks */
-    desktop_path = caja_get_desktop_directory ();
 
     /* home folder */
-    if (strcmp (g_get_home_dir(), desktop_path) != 0) {
-        char *display_name;
+    char *display_name;
 
-        mount_uri = caja_get_home_directory_uri ();
-        display_name = g_filename_display_basename (g_get_home_dir ());
-        icon = g_themed_icon_new (CAJA_ICON_HOME);
-        last_iter = add_place (sidebar, PLACES_BUILT_IN,
-                               SECTION_COMPUTER,
-                               display_name, icon,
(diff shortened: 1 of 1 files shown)


raveit65:
Ok, it seems that changing the gsettings key for it is broken without this PR.
- enable desktop is home dir + a caja window is open = the caja window closed + home dir is desktop
- enable desktop is home dir + any caja window is open = nothing happens
same behaviour with disabling the key.

lukefromdc:
I don't have any problem with holding this for 1.21, given that this key is rarely set. I also noticed that changing the gsettings key did not change what was actually displayed as the desktop right away either, logged out of the test account and back into it to see it changed but no doubt could just have killed caja.

Not only the sidebar but the desktop itself need to be refreshed anytime this key is changed-and this is not new, going back all the way to Nautilus in GNOME 2. A long time ago I had an issue with exactly this: desktop-is-home-directory got accidently set and changing it back seemed to do nothing. I was unaware that killing what was then Nautilus would fix it, so it was trial and error until I had my desktop back.

raveit65:
I set this to approved as the problem with applying the gsetting key still exists without the PR.
If you think the change doesn't break anything than merge it. Personal i would leave it for 1.21 as it isn't perfect, a seldom used key and we're a few week before release.
Maybe we wait for a approval from a third reviewer.

lukefromdc:
I will wait to see what someone else says here, favor merging as either way it will be little seen and it does fix one of two bugs with this gsettings key.

flexiondotorg:
I say merge it and refine it in 1.21.

lukefromdc:
Rebased and merged with
https://github.com/mate-desktop/caja/commit/8926c42fd69d02ea4ff977039e37fd64d98ee039

PR #898, opened 2018-01-10
fix bug: the search_dispose and search_finilize can't be run after search finished. Closes #895

### LINKED ISSUE #895: why the search_dispose about search directory don't run?
#### Expected behaviour
when search behavior finished, the search_dispose, search_finalize, caja_directory_finalize should be called to run. 
#### Actual behaviour
all finalize functions about search directory don't  be call to run.

1. add log info in search_dispose, search_finalize, caja_directory_finalize
2. make & make install
3. open caja file manager
4. search 
5. close caja file manager
6. view log information

#### Package version
caja-1.18.3

chendonghai01:
there is g_object_ref_sink in caja_bookmark_new of caja-bookmark.c , so The caja_bookmark_finilize will not be run to free object resources. because there is CajaSearchDirectory  reference in CajaFile of CajaBookmark, so the search_dispose , search_finalize will be not called to run.

### DIFF
--- a/libcaja-private/caja-bookmark.c
+++ b/libcaja-private/caja-bookmark.c
@@ -572,7 +572,6 @@ caja_bookmark_new (GFile *location, const char *name, gboolean has_custom_name,
     CajaBookmark *new_bookmark;
 
     new_bookmark = CAJA_BOOKMARK (g_object_new (CAJA_TYPE_BOOKMARK, NULL));
-    g_object_ref_sink (new_bookmark);
 
     new_bookmark->details->name = g_strdup (name);
     new_bookmark->details->location = g_object_ref (location);


chendonghai01:
the system user must be root, the test case is below:
1. add the log code as below in search_dispose, search_finalize of caja_search_directory.c
   syslog(LOG_ERR, "======== func: %s", __func__);
2. add "#include <syslog.h>" in caja_search_directory.c 
3. open caja file manager
4. enter keywords and search
5. please monitor system log.  (ex: journalctl -f on fedora 25 mate desktop)
6. close caja file manager after search finished.
7. please check if the search_dispose and search_finalize above is present in system log

result:
1. if the patch is used, the log of search_dispose and search_finalize  above is present 
2. if the patch is not used. the log of search_dispose and search_finalize  above is not present 

why the system user must be root? 
if the system is not root, because there is desktop function in caja, the caja never be exit, so the eel_debug_shut_down in src/caja-main.c don't be run.  the bookmark  object   in history_list also don't be free, so the search_dispose, search_finalize log will be not present in system log.

i think we should be to fix it for non-root user

thanks

chendonghai01:
the system user must be root, the test case is below:

add the log code as below in search_dispose, search_finalize of caja_search_directory.c
syslog(LOG_ERR, "======== func: %s", func);
add "#include <syslog.h>" in caja_search_directory.c
open caja file manager
enter keywords and search
please monitor system log. (ex: journalctl -f on fedora 25 mate desktop)
close caja file manager after search finished.
please check if the search_dispose and search_finalize above is present in system log
result:

if the patch is used, the log of search_dispose and search_finalize above is present
if the patch is not used. the log of search_dispose and search_finalize above is not present
why the system user must be root?
if the system is not root, because there is desktop function in caja, the caja never be exit, so the eel_debug_shut_down in src/caja-main.c don't be run. the bookmark object in history_list also don't be free, so the search_dispose, search_finalize log will be not present in system log.

i think we should be to fix it for non-root user

thanks

monsta:
> why the system user must be root?
if the system is not root, because there is desktop function in caja, the caja never be exit, so the eel_debug_shut_down in src/caja-main.c will not be run. the bookmark object in history_list also don't be free, so the search_dispose, search_finalize log will be not present in system log.

You can tell the current instance of Caja to quit (`caja -q`) to work around this. A new instance of Caja will be started by the session manager after that.

chendonghai01:
if the caja file manager be closed, should the bookmark object in history_list  be free ?

monsta:
Also you can auto-close the bug report by your commit, as described [here](https://help.github.com/articles/closing-issues-using-keywords/). In general, it's preferred to have a small summary in the first line of the commit message, and some explanation (if needed) in the next lines. The fix/close keywords usually go into that explanation. An example: https://github.com/mate-desktop/mate-control-center/commit/afcb324eb181a9c6cb0b3c8e8589ed99391a63a0.

raveit65:
merged https://github.com/mate-desktop/caja/commit/9b12785db1907b279fb7d6d80fd1d9fd4bb04bfb
Thanks

PR #908, opened 2018-01-19
Add Travis CI configuration

jhasse:
This will build Caja on Fedora 27.

Output would look like this: https://travis-ci.org/jhasse/caja/builds/330877194

The repository would need to be activated here: https://travis-ci.org/mate-desktop

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,6 @@
+language: c
+sudo: required
+services:
+  - docker
+script:
+  - docker run -v$PWD:/w -it fedora:27 /bin/sh -c "cd w && dnf install -y mate-common gobject-introspection-devel mate-desktop-devel libxml2-devel libnotify-devel redhat-rpm-config libSM-devel -yq && ./autogen.sh && make"


raveit65:
Can you please give some instructions to use https://travis-ci.org/mate-desktop for a f27 build?

jhasse:
All that is needed is to active the mate-desktop/caja repo in Travis settings. Then, after this PR is merged, the instructions in the `.travis.yml` will automatically be picked up.

The commands in the `script:` section are being run in a checkout of a commit to test it. This would be

```
docker run -v$PWD:/w -it fedora:27 /bin/sh -c "..."
```

`$PWD` is the current working directory which get's mounted into the docker container at `/w` via the `-v` (for volume) parameter. Why use Docker at all? Travis CI uses Ubuntu 14.04 which is a little too old to build Caja.

So now in the Docker container the following gets executed:

```
cd w && dnf install -y mate-common gobject-introspection-devel mate-desktop-devel libxml2-devel libnotify-devel redhat-rpm-config libSM-devel -yq && ./autogen.sh && make
```

Change to the `/w` directory where the sources are, install the build dependencies, run autogen, make. If any of these commands fail, this will result in an email to the person who pushed the commit or created the PR (I think to the email he set-up in his GitHub account).

The result of the test run will also be shown on the PRs and the commits at GitHub.

There's no particular reason I chose Fedora, it just happens to have a new version of mate-desktop and they don't diverge too much from upstream.

raveit65:
So, than i need to install docker?
And where the rpm will be build?
At travis-CI or github server or at my box in a docker container?
I never used docker.....
I think we leave that for after mate-1.20 release,  we're quite busy with release preparation.
And this needs to be more discussed as most MATE ppl don't have a benefit from it as they still use deb installer package.
Not to mention that i like it as fedoras MATE maintainer ;-)

jhasse:
Ah sorry for the confusion: This isn't for creating the RPM package ;) This doesn't even help with the packaging at all, it's just a nice way for you to see if a commit fails to build or not. For example you'll see for every pull request if it succeeds to build on Travis CI.

Docker will be ran at Travis CI and is only an implementation detail, nothing changes for the way Caja is build.

raveit65:
OK, thanks for clarification.
So this works only if a distro provides a docker image like fedora for uploading by Travis CI ?
At least we see a normal build log at https://travis-ci.org/jhasse/caja/builds/330877194
So i think the other guys which are using debian or ubuntu have a benefit too.
Maybe it's better to use fedora rawhide instead of f27, because latest dev version of Mate landing always first in rawhide.

@flexiondotorg @monsta 
Thoughts?

jhasse:
> So this works only if a distro provides a docker image like fedora for uploading by Travis CI ?

Yes, nearly all distros provide a Docker image though. You can search for them here: https://store.docker.com/

> So i think the other guys which are using debian or ubuntu have a benefit too.

Indeed. There isn't anything RPM or Fedora specific, it really is just for seeing that it builds ;)

> Maybe it's better to use fedora rawhide instead of f27, because latest dev version of Mate landing always first in rawhide.

Good point. The reason I chose 27 is that I found that Rawhide sometimes breaks (e.g. a shared library dependency is missing). It can maybe be changed to rawhide or Arch Linux at a later point.

monsta:
> Travis CI uses Ubuntu 14.04 which is a little too old to build Caja.

Good reason to stay away from that service.

jhasse:
Could you elaborate? This isn't using Ubuntu 14.04 to build but Fedora 27.

raveit65:
>> Travis CI uses Ubuntu 14.04 which is a little too old to build Caja.

> Good reason to stay away from that service.

Why is this a reason not to use docker technologie for CI to check our commit?
You can use every distro which provides docker images.
If you feel better we can use a debian docker image if debian provide such modern stuff.
I don't care which docker images are used,

raveit65:
any way, i am +1 for merging it.

raveit65:
Closing in favor of https://github.com/mate-desktop/caja/pull/1128

PR #912, opened 2018-01-27
avoid deprecated GtkStock

sc0w:
The expected with this PR: the same icons like before

Note: I think the translations are affected, so, this is not for 1.20


sc0w:
fixed some keyboard shortcuts with latest changes

lukefromdc:
I just retested, can confirm the broken cut and paste by keyboard shortcut with the original PR code, that it was not broken in master, and that this update to the PR fixes it. Didn't catch that in initial testing because I usually use the right click menus for this

PR #917, opened 2018-02-11
fix backgrounds and emblems dialog content rendering

lukefromdc:
Fix https://github.com/mate-desktop/caja/issues/506

Note that it still takes a few seconds for new content to show when switching between patterns, colors, and emblems but now they at least appear. Also it was necessary to use gtk_widget_queue_allocate and not gtk_widget_queue_resize to avoid massive amounts of  "Allocating size to<all widgets using eel_wrap_table> without calling gtk_widget_get_preferred_width/height(). How does the code know the size to allocate?" log spam.

### DIFF
--- a/eel/eel-wrap-table.c
+++ b/eel/eel-wrap-table.c
@@ -53,6 +53,7 @@ struct EelWrapTableDetails
 
     guint is_scrolled : 1;
     guint cols;
+    gboolean drawn;
 };
 
 /* Private EelWrapTable methods */
@@ -252,6 +253,11 @@ eel_wrap_table_draw (GtkWidget *widget,
                                       cr);
     }
 
+    /*Redraw the table once and only once to ensure it is displayed */
+    if (wrap_table->details->drawn == FALSE){
+        gtk_widget_queue_allocate (GTK_WIDGET(widget));
+        wrap_table->details->drawn = TRUE;
+        }
     return FALSE;
 }
 
@@ -797,6 +803,8 @@ eel_wrap_table_new (gboolean homogeneous)
     wrap_table = EEL_WRAP_TABLE (gtk_widget_new (eel_wrap_table_get_type (), NULL));
 
     eel_wrap_table_set_homogeneous (wrap_table, homogeneous);
+
+    wrap_table->details->drawn = FALSE;
 
     return GTK_WIDGET (wrap_table);
 }


lukefromdc:
My test was with compiz-and a very fast CPU, though it still worked when the CPU was pinned to idle speed. Will test again with marco. The function involved can take 3-10 seconds to get invoked again on my system

lukefromdc:
Update: I just found the contents are not redrawn until you move your mouse off the just-pressed button. Once you do with this PR and in my test on my machine (compiz in this test), the contents are redrawn immediately.

lukefromdc:
Latest commit should fix the delayed rendering. By using a bunch of g_message commands temporarily inserted, I found that the code to draw the wrap table was running a second time when the mouse leaves the button, no idea why. I checked the "toggle" signal, it is sent and its callback run when the button is pressed, not when it is left. 

Moving the gtk_widget_queue_allocate to eel_wrap_table_draw made the table contents show up immediately, but also made the draw function run over and over again, wasting CPU cycles. Setting the "drawn" boolean variable was the fix for that, so it would be redrawn only if not redrawn before. Since changing the category makes a new wrap table, this starts back at FALSE and redraws the table each time you change categories.

raveit65:
Without testing the latest commit i found out that it depends on themes.
Only with the first commit it works perfect with Adwaita themes.
Nearly perfect with Menta themes.
Ok with UbuntuStudio legacy theme.
Bad with submarine themes.

raveit65:
Cool, you did it :)
This works fine now with every theme.

lukefromdc:
Squashed for merge

raveit65:
Cherry-picked to 1.20/18 branches.

monsta:
Oops. This can't be used as is for 1.18 - `gtk_widget_queue_allocate` is only available starting from GTK+ 3.20. So it doesn't build with older GTK+ versions. Can we try some other function?

raveit65:
Ohh, it depends on where 1.18 is used.
I use it for f26 with gtk+-3.22, see https://src.fedoraproject.org/rpms/caja/c/a476e808821242f9d30664bb35a174070b1db01b?branch=f26
Is there really a distro with 1.18 with
a) < gtk+-3.20 ?
b) Where the Mate maintainer isn't blocked by distro policies or to lazy to update mate-1.18 ?

sc0w:
I see caja 1.18 in stretch-backports, but stretch uses gtk 3.22 , so, there is no problem in debian

raveit65:
Or add a lovely GTK_VERSION_CHECK to 1.18 branch to be sure :)

monsta:
I need this to work in Mint 18.3 (GTK+ 3.18) and LMDE 2 (GTK+ 3.14). I'd add a version check if it was possible, but here that function is needed for the fix.

lukefromdc:
Trying to use gtk_widget_queue_resize got enough "Allocating size to without calling gtk_widget_get_preferred_width/height(). How does the code know the size to allocate?" to fill up logs. Might work with GTK 3.18, but my current setup no longer allows easily testing older GTK versions

monsta:
It's ok, I have VMs for testing. So, `gtk_widget_queue_resize` works with GTK+ 3.14 and 3.18, there are no runtime warnings, and the fix still works.

Looks like only I can test this with older GTK+ versions... so should I make a PR or just push the update with `GTK_CHECK_VERSION` added?

I also think about force-pushing the branch for one reason: to get rid of the commit that doesn't build with all GTK+ versions that are supported by 1.18.

raveit65:
> Looks like only I can test this with older GTK+ versions... so should I make a PR or just push the update with GTK_CHECK_VERSION added?

Push it.

> I also think about force-pushing the branch for one reason: to get rid of the commit that doesn't build with all GTK+ versions that are supported by 1.18.

Please revert the culprit commit for 1.18, no need to break other ppl local repos with 'git push -f'

monsta:
Ok, pushed a new commit for that, https://github.com/mate-desktop/caja/commit/a7e1cd759da81eb50b28b2c114fd686d45053837.

PR #944, opened 2018-03-03
caja-file-management-properties.ui: avoid deprecated GtkVBox and GtkHBox

sc0w:
the expected: the same look and behavior like before


sc0w:
merged: https://github.com/mate-desktop/caja/commit/683e944d5b12fc72717f36588184cfe8dde7390f

PR #959, opened 2018-03-28
modify code error: end the statement with a semicolon

zhangwen2:
modify code error: end the statement with a semicolon

### DIFF
--- a/libcaja-private/caja-search-directory.c
+++ b/libcaja-private/caja-search-directory.c
@@ -525,9 +525,9 @@ search_engine_hits_added (CajaSearchEngine *engine, GList *hits,
             caja_file_monitor_add (file, monitor, monitor->monitor_attributes);
         }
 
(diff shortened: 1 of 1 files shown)


zhangwen2:
it’s not a problem under normal circumstances，but the writing is not the intention of this code。

lukefromdc:
As I said, what surprised me is that the code with the mistake EVER worked, this somehow was missed and your fix is long overdue. Hoping for another review before I merge this

will-kong:
of course it can work well.
as the rule in C/CPP language, two expressions can be written on single one line with a comma/semicolon as a delimiter. these two expressions wil be executed sequentially. (Sorry for i forgetting the order of executing ^_^)
in this case, two function will be invoked sequentially, this is why there is no functional error ocurrs. But i think this maybe a bad style.

monsta:
Wiki has pretty good summary for this operator:
https://en.wikipedia.org/wiki/Comma_operator

Issue #952, opened 2018-03-24
Feature request: Sort by extension


Smirftsch:
I know this is a VERY old story and somewhat hated? topic, but please let me explain before it goes unread into trash.
I had a look myself in adding it, but got quickly lost inside the dependency libs and use of their functions.

Reasoning: I have to interact often with windows files and I have a cross OS project Linux/Windows. In this case neither type nor mime-type sort does work out if I want to copy just some specific files.

Examples:  
Sort by "Type" does mix up localization files like .int, .itt, .frt and even worse, it does mix in also .ini files here. Mime type again does sort them all as text/plain.
By "Type" mixes up .bsc, .mhr, .u and some more development files. 
In the Linux case it also does mix up the symlinks .so and the libs .xx.xx - this does make sense, no doubt, but is also not needed or wanted if one wants f.e. to copy only plain .so files. 

Please don't argue this is something which has to be fixed by mime recognition, that's no solution, it's comparing apples and oranges. Mime type never will be able to recognize and differentiate properly everything especially when it comes to localization files like the above mentioned. 

Sorting by extension does make sense if one has to work with multiple OS files especially from the Windows side and seen technically it should be very simple to add it for one familiar with all the libs involved in caja development. I really don't get it why such a thing is discussed that often, is always denied and that in literally all desktop environments- as if the community is afraid of being to much "Windows alike" or something. It's really just about workflow and functionality here.
Please consider adding this functionality.

muesli4:
> I had a look myself in adding it, but got quickly lost inside the dependency libs and use of their functions.

The actual functionality is in *libcaja-private* folder.

> In this case neither type nor mime-type sort does work out if I want to copy just some specific files.

I think it is a reasonable request.

> Please don't argue this is something which has to be fixed by mime recognition, that's no solution, it's comparing apples and oranges.

Is there a problem with mime recognition?

> Please consider adding this functionality.

Maybe, I will look into this. But I'm not sure how one should handle filenames with multiple dots in it.

* How would I identify the proper extension? First dot? It's probably irrelevant.
* Does it have an extension at all or just dots in the name? Not our problem then?
* Should we compare basename based on dot segments in reverse order? It looks to me that this would work properly.

muesli4:
The trickiest part was to figure out how everything connects. It really looks like caja needs a clean rewrite. Anyway, this is how it looks now:
[...]

PR #757, opened 2017-04-02
new feature: Add size on disk along with size of files

fbarriga:
Currently caja only display (and count) the size of files without considering the size of the file on the filesystem. If you work with sparse files (or multiple small files) you will found handy to check the space that the files are actually using on the disk. Sparse files can be created when you start a torrent and the client pre-allocate the space of the whole file.

In order to test this, you can create 2 files: one with 3 bytes and 1 big sparse file:
```
$ echo "fo" > 3_bytes_file.txt
$ dd of=sparse_file.txt bs=1k seek=3120000 count=0
$ echo "f" >> sparse_file.txt
```

**Note:** you need to use a filesystem that support sparse files (e.g.: ext2/3/4, btrfs)
**Note2:** It has been tested on NFS and works without issues.

Screenshots:
size on disk
directory_size
sparse_file
3_bytes_file

PR #960, opened 2018-03-28
add sort criterion by reversed extension segments

muesli4:
This implements a new way to sort filenames. Entries are sorted based on their extension segments (e.g., `foo.tar.gz` has extension segments `gz`, `tar` and `foo`). This allows a better overview of files when filenames are more important (or correct) than MIME information.

This implements https://github.com/mate-desktop/caja/issues/952.

I also fixed an error that caused setting the gsettings key related to the default sort order to the wrong value (in the caja preferences dialog).


muesli4:
At the moment I'm not using a locale-dependent comparison. Please let me know how I could improve that. (I.e. what is the *collate key*?)

raveit65:
It seems to work well, but why removing "trash-time" and  "size_on_disk" here.
```
 static const char * const sort_order_values[] =
 {
     "name",
     "directory",
     "size",
-    "size_on_disk",
     "type",
     "mtime",
     "atime",
     "emblems",
-    "trash-time",
+    "extension",
     NULL
```

muesli4:
> It seems to work well, but why removing "trash-time" and "size_on_disk" here.

The array lists enum values that correspond to the default sort order (i.e. the attribute that files should be sorted with by default) which can be selected from a combo box:

* I removed `size_on_disk` because it was not present in the ui file (i.e. not selectable in the combo box). But it shifted the index so it was actually a bug that I fixed. I can add it back in if desired (then I will add an entry to the UI for `size_on_disk`).

* I removed the trash time, because it doesn't make sense to use it as a default sort criterion. Most files don't even have that attribute.

The feature is probably so rarely used that no one encountered it yet (I did while testing). My apologies for the mess. Let me know how I should proceed. (Maybe a descriptive commit that explains the bug?)

I'm also preparing another commit that provides a heuristic for identifying file extensions. My initial results are promising and the column in the list view looks better (without false positives).

raveit65:
I noticed that those options are selectable in dconf-editor, so it can be possible that someone use it.
Honestly i never uses those options, but i am sure someone will blame us if something doesn't work like before.
I am agree with you with you to remove the  'trash time' option, but this should be removed from gsettings too or not?

muesli4:
@raveit65 It still supports those values. They are just not correctly displayed in the preferences GUI. I added those two values back in.

> I am agree with you with you to remove the 'trash time' option, but this should be removed from gsettings too or not?

As long as I don't fully understand the code I don't want to remove something. This should probably part of another fix.

libcaja-private/caja-file.c
```diff
@@ -3286,6 +3474,9 @@ caja_file_compare_for_sort (CajaFile *file_1,
 				result = compare_by_full_path (file_1, file_2);
 			}
 			break;
+		case CAJA_FILE_SORT_BY_EXTENSION:
+			result = compare_by_basename_segments (file_1, file_2);
+			break;
```
vkareh: All other sort criteria have a fallback just in case:
```c
if (result == 0) {
    result = compare_by_full_path (file_1, file_2);
}
```
It might make sense to add that to this as well to keep it consistent (unless there's a valid reason to not fall back to full path?)

src/caja-file-management-properties.ui
```diff
@@ -83,6 +83,9 @@
       <row>
         <col id="0" translatable="yes">By Size</col>
       </row>
+      <row>
+        <col id="0" translatable="yes">By Size on Disk</col>
+      </row>
```
vkareh: Thanks for adding this - would be interesting to find out when/why this got removed

src/caja-file-management-properties.c
```diff
@@ -94,6 +94,12 @@ static const char * const zoom_values[] =
     NULL
 };
 
+/*
+ * This array corresponds to the object with id "model2" in
+ * caja-file-management-properties.ui. It has to positionally match with it.
+ * The purpose is to map values from a combo box to values of the gsettings
+ * enum.
+ */
```
vkareh: Thanks for adding this comment! It should reduce future bugs like that missing "By Size on Disk" :slightly_smiling_face:

libcaja-private/caja-file.c
```diff
@@ -3160,6 +3162,192 @@ compare_by_full_path (CajaFile *file_1, CajaFile *file_2)
...
+					has_letters = TRUE;
+				}
+				/* fail if it is neither digit nor letter */
+				else if (!isdigit (c)) {
+					result = FALSE;
+					break;
+				}
+
+				/* only accept short segments */
+				if (n > 3) {
```
vkareh: This `n` value is confusing - I'm assuming that since you're inside a `do`, the whole thing is added 1? I have some `foo.debian.tar.xz` and some `foo.orig.tar.xz` files: I can see the `orig` part in the extensions column, but not the `debian`.

I guess I don't mind where the cut-off is since if you're using `.debian.tar.xz` to do anything meaningful you should probably be in a terminal :smile_cat: - anyway, my issue is that this could use a `#define` value for the number of characters that still get treated as an extension, say for example `#define SORT_BY_EXTENSION_MAX_LENGTH`, and then here you can do
```c
if (n > SORT_BY_EXTENSION_MAX_LENGTH - 1) { ... }
```
(I also noticed that there's a similar `while (segment_index < 4);` below, so a defined macro number will make sense for future reference/debugging/updating)

src/caja-file-management-properties.c
muesli4: That was the intention. :slightly_smiling_face:

I thought about moving all those GSettings enum values to the GUI file as a column in the `GtkListStore`, which would avoid future errors. But I'm not sure it is the best way and also not so much related to this pull request.

lukefromdc:
I really don't know enough about this one to evaluate it.

libcaja-private/caja-file.c
muesli4: > This n value is confusing - I'm assuming that since you're inside a do, the whole thing is added 1?

You are absolutely correct. I will change it to `char_offset` to properly reflect its purpose.

> I have some foo.debian.tar.xz and some foo.orig.tar.xz files: I can see the orig part in the extensions column, but not the debian.

Neither should actually show (maximum length of following segments is 3). It is a bug, which I will fix in an instant. Now, I would be more than happy if someone actually came up with a heuristic that would work better. But I fear dots are used for other purposes too. Another approach would be to use a list of known extension segments (`tar`, `pkg`, `debian`). But it could be very time-consuming.

I will replace the hard-coded constants with macros.

muesli4:
> I mentioned a few nice-to-have changes in the code, but overall I'm approving this change, as it works just fine as is.

@vkareh Thanks a lot for your suggestions. I implemented all of them.

> It's a bit hard to get used to how odd it looks when sorting files with compound extensions, but the behavior is appropriate (and I guess anyone who needs this feature would probably want it working that way).

I originally intended it that way. Oddly enough, I got confused at one point and was convinced that I introduced a bug (which I did not).

raveit65:
@muesli4 
Can you squash the commits and rebasing the PR please?
Or does that make sense to have so much single commits?

muesli4:
I squashed all commits into one and rebased on the *master* branch.

src/caja-file-management-properties.ui
monsta: Maybe it was just an oversight, wasn't included in https://github.com/mate-desktop/caja/pull/757...

monsta:
The extension column in the list view looks a bit weird, given that the name column already has the full name. I understand this column is needed to allow sorting by extension in the list view, but still...

src/caja-file-management-properties.ui
muesli4: @monsta It can be added by providing an entry here.

https://github.com/mate-desktop/caja/blob/9ea3b711a6a24d4b68231711129982a34938479b/src/file-manager/fm-icon-view.c#L1729

muesli4:
> The extension column in the list view looks a bit weird, given that the name column already has the full name.

It is true. However, since I'm using it by myself I find it very useful to compare file extensions without having to read a long name. (I implemented it out of fun at first.) I thought about a different format for the segments, e.g.:
```
gz, tar
xz, tar, pkg
```
or
```
gz / tar
xz / tar / pkg
```
Not sure whether it would be more or less confusing.

>  I understand this column is needed to allow sorting by extension in the list view, but still...

And of course as with every sort criterion the column can be simply not shown (i.e., no overhead).

PR #965, opened 2018-03-31
caja-navigation-window-menus.c: avoid deprecated GtkImageMenuItem

sc0w:
avoid deprecated 'gtk_image_menu_item_set_always_show_image'

cajamenugo


lukefromdc:
Recalling core team to the stand...
Rebased against master, but holding merge a bit in case any further discussion is needed of the behavioral change of making the "icons in menus" gsettings preference actually work.

raveit65:
Does that warning comes from the commit?
```
caja-navigation-window-menus.c:458:16: warning: variable 'menuitem' set but not used [-Wunused-but-set-variable]
     GtkWidget *menuitem;
                ^~~~~~~~
```

sc0w:
@raveit65

> Does that warning comes from the commit?

yes, fixed in https://github.com/mate-desktop/caja/pull/966

PR #968, opened 2018-04-02
caja-side-pane.c: avoid deprecated GtkImageMenuItem

sc0w:
avoid deprecated:

gtk_image_menu_item_new_with_label
gtk_image_menu_item_set_image

sidebar_menuitem


raveit65:
Looks like before.

PR #970, opened 2018-04-05
caja-spatial-window.c: avoid deprecated GtkImageMenuItem

sc0w:
avoid deprecated:

gtk_image_menu_item_set_always_show_image
gtk_image_menu_item_new_with_label
gtk_image_menu_item_set_image

press_spatial

release_spatial

### DIFF
--- a/eel/eel-gtk-extensions.c
+++ b/eel/eel-gtk-extensions.c
@@ -494,3 +494,27 @@ eel_image_menu_item_new_from_icon (const gchar *icon_name,
 
     return menuitem;
 }
+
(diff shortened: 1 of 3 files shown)


lukefromdc:
Note that gtk_image_menu_item_set_always_show_image now remains in two other files:
caja-window-menus.c
and 
fm-directory-view.c
and with this commit, users can now turn icons in menus from the menubar in a spatial window on and off but the icons in the combobox in the status bar are still shown unconditionally. Thus those are probably controlled by one of those two files, but that's another PR anyway

Issue #903, opened 2018-01-13
Infrequent but non-zero segfaults, often but not always on clicking a remote drive's desktop icon


lukefromdc:
#### Expected behaviour
Caja does not segfault

#### Actual behaviour
Caja segfaults, restarts leaving messages like

Jan 10 16:47:24 ubuntu kernel: [ 7807.615727] caja[32683]: segfault at 557997263ab0 ip 0000000037b25f92 sp 0000000016cf9415 error 4 in libgobject-2.0.so.0.5500.0[7f86adfb0000+52000]

in the logs

#### Steps to reproduce the behaviour
Start a mate session, keep doing various file jops, mounting and using external drives, etc.  Be sure to click on external drive icons on the desktop, this has the best chance of triggering a segfault and crash.  You might see a segfault or two in a long session-or maybe not

#### MATE general version
1.19.0 from git master

#### Package version
1.19.3 from git master current as of 1-10-2016

#### Linux Distribution
Debian Unstable

#### Link to downstream report of your Distribution
None, this is a locally compiled package

lukefromdc:
Here is the sum total of all "errors" and "warnings" found by CppCheck in examing Caja's build directory. Style warnings, portability warnings, and information messages are excluded as there are tons of style warnings none of which would be likely to cause a segfault:


[Caja_CppCheck_errors_and_Warnings.txt](https://github.com/mate-desktop/caja/files/1628043/Caja_CppCheck_errors_and_Warnings.txt)

monsta:
Do you have `systemd-coredump` package installed? If yes, check if there are any crash dumps in `/var/lib/systemd/coredump/` directory. You can then use `coredumpctl` tool to run gdb and get the backtrace.

lukefromdc:
Didn't have it installed, just got it. Will look for a core dump next time Caja crashes on this

lukefromdc:
Here's what I got from systemd-coredump after one of these crashes:

```
  PID: 3531 (caja)
           UID: 1000 (luke)
           GID: 1000 (luke)
        Signal: 11 (SEGV)
     Timestamp: Thu 2018-01-25 22:16:38 EST (4min 30s ago)
  Command Line: caja
    Executable: /usr/bin/caja
 Control Group: /user.slice/user-1000.slice/session-2.scope
          Unit: session-2.scope
         Slice: user-1000.slice
       Session: 2
     Owner UID: 1000 (luke)
[...]

PR #973, opened 2018-04-06
Fix warnings from HiDPI commits

lukefromdc:
*caja-pathbar: fix warning,and in the process maybe fix occassional segfaults on opening a mounted device from the desktop

*caja-icon-info: Stop  gdk_cairo_surface_create_from_pixbuf: assertion 'GDK_IS_PIXBUF (pixbuf)' failed warning by handling NULL return from function finding the pixbuf

caja-icon-canvas-item-c: do not use g_object_unref where it does not work Stop the g_object_unref: assertion 'G_IS_OBJECT (object)' failed warnings on closing icon views

### DIFF
--- a/src/caja-pathbar.c
+++ b/src/caja-pathbar.c
@@ -1538,9 +1538,13 @@ setup_file_path_mounted_mount (GFile *location, ButtonData *button_data)
     GFile *root, *default_location;
     gint scale;
 
+    /* Return false if button has not been set up yet or has been destroyed*/
+    if (!button_data->button)
+        return FALSE;
     result = FALSE;
     volume_monitor = g_volume_monitor_get ();
     mounts = g_volume_monitor_get_mounts (volume_monitor);
+
     scale = gtk_widget_get_scale_factor (GTK_WIDGET (button_data->button));
     for (l = mounts; l != NULL; l = l->next)
     {
--- a/libcaja-private/caja-icon-canvas-item.c
+++ b/libcaja-private/caja-icon-canvas-item.c
@@ -280,7 +280,7 @@ caja_icon_canvas_item_finalize (GObject *object)
 
     if (details->rendered_surface != NULL)
     {
-        g_object_unref (details->rendered_surface);
+    cairo_surface_destroy (details->rendered_surface);
     }
 
     if (details->editable_text_layout != NULL)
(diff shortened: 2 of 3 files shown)


src/caja-pathbar.c
```diff
@@ -1541,6 +1541,10 @@ setup_file_path_mounted_mount (GFile *location, ButtonData *button_data)
     result = FALSE;
     volume_monitor = g_volume_monitor_get ();
     mounts = g_volume_monitor_get_mounts (volume_monitor);
+    /* Return false if button has not been set up yet or has been destroyed*/
```
monsta: I would place this closer to the beginning, because `mounts` apparently should be freed (see `g_list_free_full` at the end of the function), so it's better to return before it's allocated.

monsta:
BTW, this will unconditionally close #903, because Github sees "fix #903" in "maybe fix #903". Just in case you'd still prefer to keep that report open.

src/caja-pathbar.c
monsta: This might also go into a separate commit, as it's not related to HiDPI, and I'd pick it for 1.18 branch then.

monsta: Hmm, maybe it won't help in 1.18... `button_data->button` is only used to get the scale.

lukefromdc:
Just updated PR and force-pushed after applying @monsta 's suggestion to use cairo_surface_destroy (details->rendered_surface) to free that variable, upon which g_object_unref is not valid

src/caja-pathbar.c
lukefromdc: This is actually needed here, as one of the warnings was coming from it. Otherwise it would be a separate PR. Moving though as you suggested

lukefromdc:
Just force-pushed again after moving the conditional return in caja-pathbar.c to return prior to allocating mounts. This should now be ready for testing, I will prepare another PR for unconditionally dealing with https://github.com/mate-desktop/caja/issues/903, hope it will apply clean both with and without this one

lukefromdc:
Not sure now that https://github.com/mate-desktop/caja/pull/973 is affected by this, as it turns out to be slightly harder to trigger than I thought and require having a navigation window or two already open. Still stops the warnings from the HiDPI work though. Working on the other issue now

lukefromdc:
https://github.com/mate-desktop/caja/pull/974 handles 
https://github.com/mate-desktop/caja/pull/973 ,
is completely separate from this, and merges clean into master and into this so can be handles separately. Both are ready for testing, I an using a "dev-all" branch locally for now

lukefromdc:
I got the warnings, and so did @monsta and we both found the warnings gone

raveit65:
> I got the warnings, and so did @monsta and we both found the warnings gone

Np, it's not the first time that we have differences between distros.
I will cherry-pick this to 1.20 branch and release 1.20.2 tonight.

PR #974, opened 2018-04-06
pathbar: fix random segfaults on opening mounts from Desktop

lukefromdc:
Apply Nautilus commit
https://github.com/GNOME/nautilus/commit/e1ad3c05a6cd08c8cbf18ae53701dd742249d5fd#diff-f896071d07d34e87af94a18de95e4ea2
"pathbar: add a path-event signal"

and then apply most of

https://github.com/GNOME/nautilus/commit/875efc324f8e91f2d157c7532fe5570c1de421c7#diff-f896071d07d34e87af94a18de95e4ea2
pathbar: remove unused code

Which removes the function that was randomly segfaulting on creating certain navigation windows, usually on clicking a mounted volume from the desktop, probably by being called before the window was created. Replacement in button-event should never be called before the buttons exist

### DIFF
--- a/src/caja-navigation-window-pane.c
+++ b/src/caja-navigation-window-pane.c
@@ -261,63 +261,21 @@ location_button_create (CajaNavigationWindowPane *pane)
 }
 
 static gboolean
-path_bar_button_pressed_callback (GtkWidget *widget,
-                                  GdkEventButton *event,
-                                  CajaNavigationWindowPane *pane)
-{
-    CajaWindowSlot *slot;
-    CajaView *view;
-    GFile *location;
-    char *uri;
+path_bar_path_event_callback (CajaPathBar *path_bar,
+                   GFile *location,
+                   GdkEventButton *event,
+                   CajaWindowPane *pane)
 
-    caja_window_set_active_pane (CAJA_WINDOW_PANE (pane)->window, CAJA_WINDOW_PANE (pane));
-
-    g_object_set_data (G_OBJECT (widget), "handle-button-release",
-                       GINT_TO_POINTER (TRUE));
-
-    if (event->button == 3)
-    {
-        slot = caja_window_get_active_slot (CAJA_WINDOW_PANE (pane)->window);
-        view = slot->content_view;
-        if (view != NULL)
-        {
-            location = caja_path_bar_get_path_for_button (
-                           CAJA_PATH_BAR (pane->path_bar), widget);
-            if (location != NULL)
-            {
-                uri = g_file_get_uri (location);
-                caja_view_pop_up_location_context_menu (
-                    view, event, uri);
-                g_object_unref (G_OBJECT (location));
-                g_free (uri);
-                return TRUE;
-            }
-        }
-    }
-
-    return FALSE;
-}
-
-static gboolean
-path_bar_button_released_callback (GtkWidget *widget,
-                                   GdkEventButton *event,
-                                   CajaNavigationWindowPane *pane)
 {
     CajaWindowSlot *slot;
     CajaWindowOpenFlags flags;
-    GFile *location;
     int mask;
-    gboolean handle_button_release;
+    CajaView *view;
+    char *uri;
 
-    mask = event->state & gtk_accelerator_get_default_mod_mask ();
-    flags = 0;
-
-    handle_button_release = GPOINTER_TO_UINT (g_object_get_data (G_OBJECT (widget),
-                            "handle-button-release"));
-
-    if (event->type == GDK_BUTTON_RELEASE && handle_button_release)
-    {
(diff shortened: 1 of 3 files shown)


monsta:
Hmm, this generates a runtime warning...

```
sys:1: Warning: ../../../../gobject/gsignal.c:2423: signal 'path_set' is invalid for instance '0x56162a3ee810' of type 'CajaPathBar'
```

lukefromdc:
Just found a problem here requiring more work, so hold off on testing or merging:
Right-click on pathbar disabled, need to bring it back

lukefromdc:
Right now I am having nothing but trouble with this: the function removed and supposedly replaced by the GNOME commits is the one that displays the right-click menus, can't seem to figure out what else it does. The replacement code does not work here, and I using the code from master I have not been able to find any conditional test to abort caja_path_bar_update_path on clicking a desktop mount and prevent the segfault.

https://github.com/mate-desktop/caja/issues/903
can usually (but not always) be invoked by having a flash drive mount(I'm using an encrypted one but camera cards have also done this) showing on the desktop, opening two navigation windows, ensuring nothing on the desktop is selected, then selecting one folder in one of the navigation windows and then clicking several times as fast as possible on the flash drive's icon on the desktop. With the PR I will get as many copies of its window as there were clicks, also true if I remove
g_signal_emit (path_bar, path_bar_signals [PATH_SET], 0, file_path);
from  caja_path_bar_update_path unconditionally and change nothing else. With the unchanged code from master I often get the segfault.

It looks like I need to revert the code from Nautilus, remove the signal and the callbacks linked to it,  and then find a way to use only the "button-clicked" callback, adding code to detect a right-click on a pathbar button and show the right-click menu.

lukefromdc:
I just reverted the Nautilus commits, and finally found conditionals that stop the segfaults. Two functions had to be put behind conditionals, and one problem remained: the second attempt to open a desktop icon for a mount after the first try fails (as it sometimes does) still sometimes (not always) leaves path_bar->current_path as NULL . I had to make button_data_file_changed return ahead of the g_return_if_fail for this to avoid spamming logs with warnings, sometimes still get three other warnings on the second try at opening a desktop icon for a mount, but at least the navigation window opens. Here are the warnings in question that appear some of the time on that second double-click of a desktop mount icon:

```
(caja:16075): GLib-CRITICAL **: 01:34:44.722: g_hash_table_foreach: assertion 'version == hash_table->version' failed

(caja:16075): GLib-GObject-CRITICAL **: 01:34:44.728: g_object_ref: assertion 'G_IS_OBJECT (object)' failed

(caja:16075): GLib-GObject-CRITICAL **: 01:34:44.732: g_object_ref: assertion 'G_IS_OBJECT (object)' failed

```
Normally the pathbar is now responsive to the right-click, with one exception that needs to be found and fixed if possible:  In and only in those cases that generate these errors, the pathbars on already open navigation windows will ignore the right-click until the directory is changed and the function in question gets re-run.

vkareh:
@lukefromdc - to be honest I'm not sure how to reproduce the original issue. I'm using a normal USB drive and repeatedly clicking on the desktop icon for that drive. It opens multiple windows and they are all responsive. What am I doing wrong/different?

In any case, this commit builds and runs just fine for me, but it would be nice if I can figure out how to reproduce the original issue.

monsta:
I wonder how Nautilus changes were supposed to work. Added `path-event`, removed `path-set`, ok, but who connects to `path-event` after that? A signal without listeners isn't useful...

monsta:
Hmm, maybe it should be https://github.com/GNOME/nautilus/commit/2759def4968f89c1a2370ca5b2b91af84dc4afd3 and https://github.com/GNOME/nautilus/commit/cfa51e6702ade6a8ca9045791773e6e10560262f - that's what searching for "path-event" in their repo gave me.

lukefromdc:
Original problem appears to be some kind of race condition, and that means it might be machine dependent. The exact condition under which I get it is I click on the desktop icon for a mount and nothing happens, within a second or so click it again and the problems result. I will now test the later GNOME commits and see what happens, this sort of thing is why I'm not squashing until this is done

lukefromdc:
Those Nautilus commits worked, though the linked diffs were not quite the whole of the commits and missed a crucial g_signal_connect_object which I found by looking at the whole commits. With that in place, right-click and middle click on the pathbar now work with the Nautilus commits.

 I made (and pushed)  a backup of the branch, then reset it to before last night's revert of the previous nautilus commits, added these and now I've got the pathbar right click menus working and no segfaults on clicking the desktop mount icons.  The newest commit is the two Nautilus commits applied together by hand, would have been more work to split them up.

lukefromdc:
Squashed and the four Nautilus commits linked in the commit message prior to merge

PR #975, opened 2018-04-10
sidebar popup menu: show "media-eject" icon

sc0w:


monsta:
Couldn't figure out why I don't see it... then I realized I need to insert USB flash drive to get "Eject" menu item. My HDD only has "Unmount" item which doesn't have the icon. Should we add it there too?

sc0w:
@monsta 

> Couldn't figure out why I don't see it... then I realized I need to insert USB flash drive to get "Eject" menu item.

I have tested the icon with one cd

> My second internal HDD (SATA) only has "Unmount" item which doesn't have the icon. Should we add it there too?

In my opinion, yes, I think it is a good idea, and the icon is understandable there

I can't see the "unmount" item, because I don't have a second internal hdd

raveit65:
I don't see that icon if i plug in an external usb-disk-drive.
Where is that mysterious menu?
Picture please?

raveit65:
Ok, got it.
If i plugin an usb-stick i see the icon in menu.
Still wondering why caja makes an different between an usb-stick and an external usb-disk-drive?

monsta:
IIRC it's decided at gvfs level (I don't remember the details).

sc0w:
@monsta 

> My second internal HDD (SATA) only has "Unmount" item which doesn't have the icon. Should we add it there too?

done in https://github.com/mate-desktop/caja/pull/977

PR #976, opened 2018-04-11
fm-icon-view: Show size on disk as sort criterion

muesli4:
This was introduced in commit b28445b3d2c42a7d2dbac97983d007e4ab58684d
but has never been added to some parts of the UI.

The GUI part was only implemented for the list view (but still selectable as default in the preferences for other views).

### DIFF
--- a/src/file-manager/caja-icon-view-ui.xml
+++ b/src/file-manager/caja-icon-view-ui.xml
@@ -13,6 +13,7 @@
 				<placeholder name="Auto Layout">
 					<menuitem name="Sort by Name" action="Sort by Name"/>
 					<menuitem name="Sort by Size" action="Sort by Size"/>
+					<menuitem name="Sort by Size on Disk" action="Sort by Size on Disk"/>
 					<menuitem name="Sort by Type" action="Sort by Type"/>
 					<menuitem name="Sort by Modification Date" action="Sort by Modification Date"/>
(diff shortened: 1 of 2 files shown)


monsta:
BTW, it's the only sort menu item without accelerator. Should we add one (to both `sort_criteria` and `arrange_radio_entries` arrays)?

muesli4:
@monsta Damn, I wanted to do that but forgot it. Is there some scheme how to pick them (or just any letter in the word that is not colliding with others)?

monsta:
I think they just don't have to repeat in the same menu, otherwise you'll get jumping between two (or more) menu items when pressing the accelerator key. Usually they're picked from first letter of the menu item name, or second if the first is already used, or some other... here we don't have much choice anyway.

raveit65:
It works well if i use the context-menu of icon-view.
But if i use the drop-down combobox menu in preferences, i see no change in icon-view.
caja-preferences
Or is this a complete other function?

monsta:
This is supposed to set the default sort order for all new windows... but I also found out it doesn't work for me. Well, it works for some folders sometimes, I can't figure out the pattern. It's not related to this PR or https://github.com/mate-desktop/caja/pull/960, it happened in master before.

monsta:
Ok, with rebased commit menu item now has accelerator.

PR #979, opened 2018-04-20
Emit accessible-name-change signal for icons only when icons are renamed

joanmarie:
caja_icon_canvas_item_set_property was emitting accessible-name-change
signals whenever the PROP_EDITABLE_TEXT was being set for an icon. This
method is called by caja_icon_container_request_update_all when the icon
view is first loaded. Thus irrelevant events are being emitted (the name
of each icon didn't change, as far as the user is concerned). Furthermore,
this floods assistive technologies any time the user opens a folder with
a large number of files.

Moving the notification out of caja_icon_canvas_item_set_property and
into end_renaming_mode (where ICON_TEXT_CHANGED is also emitted) solves
the problem of floods of irrelevant notifications while still emitting
the signal when the name actually changes.

### DIFF
--- a/libcaja-private/caja-icon-canvas-item.c
+++ b/libcaja-private/caja-icon-canvas-item.c
@@ -379,7 +379,6 @@ caja_icon_canvas_item_set_property (GObject        *object,
         {
             gail_text_util_text_setup (details->text_util,
                                        details->editable_text);
-            g_object_notify (G_OBJECT(accessible), "accessible-name");
         }
 
         caja_icon_canvas_item_invalidate_label_size (item);
--- a/libcaja-private/caja-icon-container.c
+++ b/libcaja-private/caja-icon-container.c
@@ -9078,6 +9078,7 @@ end_renaming_mode (CajaIconContainer *container, gboolean commit)
 {
     CajaIcon *icon;
     const char *changed_text;
+    AtkObject *accessible_icon;
 
     set_pending_icon_to_rename (container, NULL);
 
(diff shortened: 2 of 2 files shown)


alexarnaud:
Thank you @joanmarie for the fix. Do you think we could do the same for the list view?

Best regards,
Alex.

raveit65:
Honestly i don't see any different with or without both PRs in accerciser app.
But i am complete clueless to use this app.
Also, this app produces abrt alams about python3 in fedora 26 and causes a high proc load.
So i stop testing this.
Building works fine and i see no new build warnings.
So, if it works for you.....

alexarnaud:
Le 21/04/2018 à 14:22, joanmarie a écrit :
> @alexarnaud <https://github.com/alexarnaud> I am not seeing that event 
> spam for the list view. Are you?

Not really, I observe such events:
> object:state-changed:sensitive(1, 0, 0)
> 	source: [menu item | Open]
> 	application: [application | caja]
> object:state-changed:enabled(1, 0, 0)
> 	source: [menu item | Open]
> 	application: [application | caja]
> object:property-change:accessible-name(0, 0, Open)
> 	source: [menu item | Open]
> 	application: [application | caja]
> object:visible-data-changed(0, 0, 0)
> 	source: [menu item | Open]
> 	application: [application | caja]
> object:property-change:accessible-name(0, 0, Open)
> 	source: [menu item | Open]
> 	application: [application | caja]
> object:visible-data-changed(0, 0, 0)
> 	source: [menu item | Open]
> 	application: [application | caja]
> object:state-changed:visible(1, 0, 0)
> 	source: [menu item | Open]
> 	application: [application | caja]

FYI, the list view is the most active view on the blind and 
visual-impaired community because it is represented on list, not on column.

What do you observe Joanie on list view?

Best regards,
Alex.

alexarnaud:
@joanmarie How many files do you have in your test folder? Are you testing Caja or Nautilus ?

Best regards,
Alex.

alexarnaud:
@joanmarie Should we use the Orca patch to makes it fast or is it already fast for you without the orca patch?

Best regards,
Alex.

PR #981, opened 2018-04-21
Eliminate accessible children-changed event flood during initial load

joanmarie:
Accessible children-changed events should not be emitted during the
initial load of the icon container. They should only be emitted when
a user-triggered action (creation of file, copying an item) causes
an item to be added.

By setting container->details->is_loading (which does not seem to
be used anywhere) and adding a call to finish_adding_new_icons when
all icons have been added, we can make the needed distinction and
emit the signal only when it makes sense to do so.

### DIFF
--- a/libcaja-private/caja-icon-container.c
+++ b/libcaja-private/caja-icon-container.c
@@ -8025,6 +8025,7 @@ caja_icon_container_request_update_all (CajaIconContainer *container)
 
     g_return_if_fail (CAJA_IS_ICON_CONTAINER (container));
 
+    container->details->is_loading = TRUE;
     for (node = container->details->icons; node != NULL; node = node->next)
     {
         icon = node->data;
@@ -8032,6 +8033,7 @@ caja_icon_container_request_update_all (CajaIconContainer *container)
     }
 
     redo_layout (container);
+    container->details->is_loading = FALSE;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


joanmarie:
Thanks @lukefromdc. Who else needs to review it?

FWIW, testing this change combined with the change in #979 make Orca way more responsive with your icon view.

monsta:
When switching very quickly between folders (via sidebar or location bar), there is a weird flickering in the top left corner of the icon view. It's like some of the icons from the folder are flashing there, then disappear. Happens when switching into some folder with lots of files and subfolders. Might take several tries to reproduce.

raveit65:
Honestly i don't see any different with or without both PRs in accerciser app.
But i am complete clueless to use this app.
Also, this app produces abrt alams about python3 in fedora 26 and causes a high proc load.
So i stop testing this.
Building works fine and i see no new build warnings.
So, if it works for you.....

flexiondotorg:
I can confirm the "flickering" that @monsta observed. I enabled Icon view and switching from `/usr/bin' to `/usr/lib` using the location bar. The issue is present most of the time, but occasionally a few switches don't exhibit the issue.

flexiondotorg:
I can confirm the "flickering" that @monsta observed. I enabled Icon view and switching from `/usr/bin' to `/usr/lib` using the location bar. The issue is present most of the time, but occasionally a few switches don't exhibit the issue.

joanmarie:
Ok, I'll take another stab at it. The fix I did for Nautilus was different because of the changes since the fork. But hopefully a similar approach will work in Caja.

Thanks for the review!

joanmarie:
@monsta and @flexiondotorg: Mind trying with the latest commit? Thanks again!!

monsta:
Oh, no need for another PR. It can be done on your current branch, and this PR will be updated after you force-push it. We can also do it ourselves, via "Squash and merge" button here on Github.

joanmarie:
@monsta Cool. In my experience, doing a force push auto-closed pull requests. But maybe that's repo-specific (or no longer a thing, or something I dreamed up). Anyoo, done. Thanks again!!

monsta:
Thanks, merged.

PR #984, opened 2018-04-26
Eliminate accessible children-changed event flood on container repopulation

joanmarie:
When the icon container is repopulated (e.g. when using back/forward)
finish_adding_new_icons is called leading to an accessible children-
changed event flood. We can distinguish these unwanted notifications from
desired notifications by comparing the number of new icons to add with
the size of the container's icon set.


joanmarie:
@lukefromdc and/or @monsta, please review. Thanks in advance!

monsta:
Icon view works for me as before (I don't use any a11y features), so this looks ok to me.

I'm not familiar with this part of code though, so I don't quite understand the logic... So the same length of that list and that table means that the icon view is still being filled with items?

PR #989, opened 2018-05-05
avoid deprecated 'gtk_style_context_get_background_color'

sc0w:


raveit65:
I just start testing, but why not using the suggestion from gtk+ as replacement?
```
warning: 'gtk_style_context_get_background_color' is deprecated: Use 'gtk_render_background' instead
```

sc0w:
because I think it can be fixed easily without 'gtk_render_background'

PR #990, opened 2018-05-06
Fix some gdk deprecations

raveit65:
Please test.

### DIFF
--- a/libcaja-private/caja-autorun.c
+++ b/libcaja-private/caja-autorun.c
@@ -753,15 +753,17 @@ static gboolean
 is_shift_pressed (void)
 {
     gboolean ret;
+    GdkDisplay *display;
     XkbStateRec state;
     Bool status;
 
     ret = FALSE;
 
-    gdk_error_trap_push ();
-    status = XkbGetState (GDK_DISPLAY_XDISPLAY (gdk_display_get_default ()),
-    			  XkbUseCoreKbd, &state);
-    gdk_error_trap_pop_ignored ();
+    display = gdk_display_get_default ();
+    gdk_x11_display_error_trap_push (display);
+    status = XkbGetState (GDK_DISPLAY_XDISPLAY (display),
+                          XkbUseCoreKbd, &state);
+    gdk_x11_display_error_trap_pop_ignored (display);
 
     if (status == Success)
     {
--- a/src/file-manager/fm-desktop-icon-view.c
+++ b/src/file-manager/fm-desktop-icon-view.c
@@ -149,6 +149,7 @@ static void
 net_workarea_changed (FMDesktopIconView *icon_view,
                       GdkWindow         *window)
 {
+    GdkDisplay *display;
     long *nworkareas = NULL;
     long *workareas = NULL;
     GdkAtom type_returned;
@@ -165,7 +166,8 @@ net_workarea_changed (FMDesktopIconView *icon_view,
      * workareas array is going to be (each desktop will have four
      * elements in the workareas array describing
      * x,y,width,height) */
-    gdk_error_trap_push ();
+    display = gtk_widget_get_display (GTK_WIDGET (icon_view));
+    gdk_x11_display_error_trap_push (display);
     if (!gdk_property_get (window,
                            gdk_atom_intern ("_NET_NUMBER_OF_DESKTOPS", FALSE),
                            gdk_x11_xatom_to_atom (XA_CARDINAL),
@@ -177,7 +179,7 @@ net_workarea_changed (FMDesktopIconView *icon_view,
     {
         g_warning("Can not calculate _NET_NUMBER_OF_DESKTOPS");
     }
-    if (gdk_error_trap_pop()
+    if (gdk_x11_display_error_trap_pop (display)
             || nworkareas == NULL
             || type_returned != gdk_x11_xatom_to_atom (XA_CARDINAL)
             || format_returned != 32)
@@ -193,7 +195,7 @@ net_workarea_changed (FMDesktopIconView *icon_view,
      * request does not correspond to the number of bytes you get
      * back, and is the reason for the workaround below.
      */
-    gdk_error_trap_push ();
+    gdk_x11_display_error_trap_push (display);
     if (nworkareas == NULL || (*nworkareas < 1)
             || !gdk_property_get (window,
                                   gdk_atom_intern ("_NET_WORKAREA", FALSE),
@@ -208,7 +210,7 @@ net_workarea_changed (FMDesktopIconView *icon_view,
         workareas = NULL;
     }
 
-    if (gdk_error_trap_pop ()
+    if (gdk_x11_display_error_trap_pop (display)
             || workareas == NULL
             || type_returned != gdk_x11_xatom_to_atom (XA_CARDINAL)
             || ((*nworkareas) * 4 * sizeof(long)) != length_returned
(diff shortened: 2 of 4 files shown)


libcaja-private/caja-autorun.c
```diff
@@ -753,15 +753,16 @@ static gboolean
...
     Bool status;
 
     ret = FALSE;
 
-    gdk_error_trap_push ();
+    gdk_x11_display_error_trap_push (display);
     status = XkbGetState (GDK_DISPLAY_XDISPLAY (gdk_display_get_default ()),
     			  XkbUseCoreKbd, &state);
-    gdk_error_trap_pop_ignored ();
+    gdk_x11_display_error_trap_pop_ignored (display);
```
sc0w: here the variable 'display' isn't defined, maybe this line is needed before?

display = gdk_display_get_default ();

src/file-manager/fm-desktop-icon-view.c
```diff
@@ -350,7 +353,7 @@ fm_desktop_icon_view_handle_middle_click (CajaIconContainer *icon_container,
     }
 
     /* Stop the event because we don't want anyone else dealing with it. */
-    gdk_flush ();
+    gdk_display_flush (display);
     g_signal_stop_emission_by_name (icon_container, "middle_click");
```
sc0w: the same happens here, maybe is needed this before?

display = gtk_widget_get_display (GTK_WIDGET (icon_container));

raveit65: i agree, just updated the PR

libcaja-private/caja-autorun.c
raveit65: i agree, just updated the PR

raveit65:
PR is rebased.

lukefromdc:
Rebased and merged with 
https://github.com/mate-desktop/caja/commit/5ed053f9c332465a173c3ceeb8e3e16dfdbfc440
through
https://github.com/mate-desktop/caja/commit/ded490e1bc432210e6006e94273cc160a0af8e75

raveit65:
@lukefromdc 
A user send me this mail with concerns about first commit.
```
Hi again,

noticed in

http://git.mate-desktop.org/caja/commit/?h=gdk-deprecations&id=2a0e431f5c959c707b5e2bc271689a4deab85b97

attempt to eliminate a warning ...

libcaja-private: fix -Wlogical-not-parentheses build warning
fixes this warning:
caja-undostack-manager.c: In function 'retrieve_files_to_restore':
caja-undostack-manager.c:1967:38: warning: logical not is only applied to the left hand side of comparison [-Wlogical-not-parentheses]
if (!(g_hash_table_size (trashed)) > 0) { 
^
caja-undostack-manager.c:1967:7: note: add parentheses around left hand side expression to silence this warning
if (!(g_hash_table_size (trashed)) > 0) { 
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
( ) 

- if (!(g_hash_table_size (trashed)) > 0) {
+ if ((!(g_hash_table_size (trashed))) > 0) {

So I read the warning as
"LOGICAL NOT here is only applied to the left hand side of comparison,
if this is as you wanted, then add parentheses around left hand side expression
to silence this warning".

I think the original code was a bit missleading, I think it was meant to be like
if trashed has_table is empty then return null because nothing to handle ...

So we do not want that the LOGICAL NOT applies to the left hand side of
the comparison only, but to the whole comparison ...
so we have to correct the original comparison to silence the warning 

(!(A) > 0)  ==>  (!(A > 0))

- if (!(g_hash_table_size (trashed)) > 0) {
+ if (!(g_hash_table_size (trashed) > 0)) {

or because g_hash_table_size() returns unsigned integer, zero when has_table is empty,

- if (!(g_hash_table_size (trashed)) > 0) {
+ if (g_hash_table_size (trashed) == 0) {

or

- if (!(g_hash_table_size (trashed)) > 0) {
+ if (!g_hash_table_size (trashed)) {

Your code change works OK, and eliminates the warning.
But the original comparison logic was a little bit wrongly written,
causing a compiler warning, so we need to fix that,
and the compiler warning disappears suddenly ...

What do you think?

Regards,

Juha Leppänen, Kuopio, Finland
```
What do you think?

lukefromdc:
Not sure what that code is doing, but if you want to try it the other way I will test it. As it stands now the undo stack works

This concerns trashed files and the undo stack, and the behavior is exactly the same as before: if a file is trashed and the trash is emptied, the undo stack "jumps over" moving that item to trash and previous items can still be undone. This will work-unless it is operating on one or more trashed files that have been deleted. Not sure if that was the origjnal author's intended behavior or not, could relate to this.

lukefromdc:
I just did some experimenting with that trash undo stack code. Never saw the warning with or without the extra parenthesis, and if I moved them to cover the whole statement nothing changed in the trash/undo behavior. That behavior as a whole (skip over the missing files) differs from what Nemo does. Nemo stops at the deleted files and won't undo past that due to the missing files. This, however is apparently not the reason (or the fix for a potentially buggy bahavior).

Since I did not get the warning on removing the extra parenthesis, my compiler might be doing something differently, or just sending fewer warnings.

raveit65:
Those are the compile flags when i am building rpms with mock build-server application for fedora.
```
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I.. -DG_LOG_DOMAIN=\"Eel\" -I.. -I.. -I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include -I/usr/include/mate-desktop-2.0 -I/usr/include/gtk-3.0 -I/usr/include/pango-1.0 -I/usr/include/fribidi -I/usr/include/cairo -I/usr/include/pixman-1 -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/uuid -I/usr/include/harfbuzz -I/usr/include/gdk-pixbuf-2.0 -I/usr/include/gio-unix-2.0/ -I/usr/include/libdrm -I/usr/include/valgrind -I/usr/include/atk-1.0 -I/usr/include/at-spi2-atk/2.0 -I/usr/include/at-spi-2.0 -I/usr/include/dbus-1.0 -I/usr/lib64/dbus-1.0/include -I/usr/include/startup-notification-1.0 -I/usr/include/dconf -I/usr/include/gail-3.0 -I/usr/include/libxml2 -pthread -DG_DISABLE_DEPRECATED -DGDK_PIXBUF_DISABLE_DEPRECATED -DDATADIR=\"/usr/share\" -DSOURCE_DATADIR=\"../data\" -DMATELOCALEDIR=\"/usr/share/locale\" -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -mcet -fcf-protection -c eel-mate-extensions.c  -fPIC -DPIC -o .libs/eel-mate-extensions.o
```
This shows a lot of warnings in caja :/

raveit65:
Ok, thanks for testing.
If behaviour isn't different we should leave it as it is.
At least author of this mail can blame us here instead of sending me private mails.

PR #991, opened 2018-05-06
Gtk menu popup deprecations

raveit65:
Please test.
Some menus aren't executable with keyboard-navigation (shift+F10) in master, so this isn't a regression.
This affects editable-label, emblem-sidebar and fm-tree-view in sidebar.
Those menus are only executable via mouse event.
Does anyone know how to trigger the menu in caja-dnd.c?
And maybe should i use gtk_menu_popup_at_widget () for caja-navigation-window-pane too?
If using keyboard navigation it can be confusing that the menu pops up under the pointer at another place in window. :)

There is one deprecation left in eel-gtk-extensions.c for the moment, but here i killed keyboard-navigation.

### DIFF
--- a/src/caja-zoom-control.c
+++ b/src/caja-zoom-control.c
@@ -133,83 +133,18 @@ zoom_popup_menu_show (GdkEventButton *event, CajaZoomControl *zoom_control)
 }
 
 static void
-menu_position_under_widget (GtkMenu   *menu,
-                            gint      *x,
-                            gint      *y,
-                            gboolean  *push_in,
-                            gpointer   user_data)
-{
-    GtkWidget *widget;
-    GtkWidget *container;
-    GtkRequisition req;
-    GtkRequisition menu_req;
-    GdkRectangle monitor;
-    GdkMonitor *monitor_num;
-    GdkDisplay *display;
-    GtkAllocation allocation;
-
-    widget = GTK_WIDGET (user_data);
-    g_assert (GTK_IS_WIDGET (widget));
-
-    container = gtk_widget_get_ancestor (widget, GTK_TYPE_CONTAINER);
-    g_assert (container != NULL);
-
-    gtk_widget_get_preferred_size (GTK_WIDGET (menu), &menu_req, NULL);
-    gtk_widget_get_preferred_size (widget, &req, NULL);
-    gtk_widget_get_allocation (widget, &allocation);
-
-    display = gtk_widget_get_display (GTK_WIDGET (menu));
-    monitor_num = gdk_display_get_monitor_at_window (display, gtk_widget_get_window (widget));
-    if (monitor_num == NULL)
-        monitor_num = gdk_display_get_monitor (display, 0);
-    gdk_monitor_get_geometry (monitor_num, &monitor);
-
-    gdk_window_get_origin (gtk_widget_get_window (widget), x, y);
-    if (!gtk_widget_get_has_window (widget))
-    {
-        *x += allocation.x;
-        *y += allocation.y;
-    }
-
-    if (gtk_widget_get_direction (container) == GTK_TEXT_DIR_LTR)
-    {
-        *x += allocation.width - req.width;
-    }
-    else
-    {
-        *x += req.width - menu_req.width;
-    }
-
-    if ((*y + allocation.height + menu_req.height) <= monitor.y + monitor.height)
-    {
-        *y += allocation.height;
-    }
-    else if ((*y - menu_req.height) >= monitor.y)
-    {
-        *y -= menu_req.height;
-    }
-    else if (monitor.y + monitor.height - (*y + allocation.height) > *y)
-    {
-        *y += allocation.height;
-    }
-    else
-    {
-        *y -= menu_req.height;
-    }
-
-    *push_in = FALSE;
-}
-
-
-static void
 zoom_popup_menu (GtkWidget *widget, CajaZoomControl *zoom_control)
 {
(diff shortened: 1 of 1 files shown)


monsta:
Hmm... why was this squashed? I'm trying to find which part of UI causes log spam with new runtime warning - and this is impossible to bisect...

```
(caja:1603): Gdk-WARNING **: Event with type 19 not holding a GdkSeat. It is most likely synthesized outside Gdk/GTK+
```

monsta:
Menu in caja-dnd appears if you drag some file/folder and hold Alt, then release the mouse button.

raveit65:
> Menu in caja-dnd appears if you drag some file/folder and hold Alt, then release the mouse button.

Ok, that is working and using gtk_menu_popup_at_pointer is fine for this case

@monsta 
I did that as single commits and you will find them at https://github.com/mate-desktop/caja/pull/991#commits-pushed-394c489
If you use the whole commit as patch you can remove single parts in the patch.
Or use this single patches.

raveit65:
This menu is a bit weird because it is only selectable with the mouse pointer, keyboard selction via shift+F10 opens only the canvasitem menu.
Maybe related?

monsta:
Aha, Shift-F10 does the same as Menu button on keyboard (it's next to the right Win key). :slightly_smiling_face: 
But no, this behavior is there in older versions too, even in GTK+2 build of 1.16. The local copy/cut/paste menu is only accessible via right click. Maybe we should correct it for better a11y.

raveit65:
Indeed,
`(caja:1603): Gdk-WARNING **: Event with type 19 not holding a GdkSeat. It is most likely synthesized outside Gdk/GTK+`
This Warning is caused by https://github.com/mate-desktop/caja/pull/991/commits/394c489610390eab2accd5dc734de0e49400626c

Should i revert it?
Better to have a deprecation warning instead of spamming users xsession-errors log.
It happens every time when the popup in editable-label is triggered.

lukefromdc:
I can also confirm that if gtk_menu_popup_at_pointer is just removed, the menu does not appear-and neither do the warnings, confirming that code and probably the use of NULL as the culprit

lukefromdc:
Same warning with gtk_menu_popup_at_widget, so we need to either revert or find a way to get the event and pass it

lukefromdc:
I was able to pass an event (or apparently pass an event...) using the diff below as applied to eel-editable-label-c to gtk_menu_popup_at_pointer but it made absolutely NO difference, same warnings appeared.. Note that according to 
https://developer.gnome.org/gtk3/stable/GtkMenu.html#gtk-menu-popup-at-pointer
passing NULL simply causes gtk_menu_popup_at_pointer to use the current event as a default. In this case, that event is the one that triggers the warning, so I see no easy fix except to revert this

```
--- /home/luke/Desktop/Development/MATE_Development_Work/AA_Master/caja/eel/eel-editable-label.c	2018-05-16 18:44:02.036722321 -0400
+++ /home/luke/Desktop/Development/MATE_Development_Work/caja_1.21.1+git20180517/eel/eel-editable-label.c	2018-05-18 01:25:03.917234894 -0400
@@ -3044,6 +3044,7 @@ typedef struct
     EelEditableLabel *label;
     gint button;
     guint time;
+    GdkEventButton   *event;
 } PopupInfo;
 
 static void
@@ -3095,12 +3096,14 @@ popup_targets_received (GtkClipboard
         g_signal_emit (label,
                        signals[POPULATE_POPUP], 0,
                        label->popup_menu);
-
+        if (info->event = NULL){ 
+                return;
+             }
         if (info->button)
-            gtk_menu_popup_at_pointer (GTK_MENU (label->popup_menu), NULL);
+            gtk_menu_popup_at_pointer (GTK_MENU (label->popup_menu), (const GdkEvent*) info->event);
         else
         {
-            gtk_menu_popup_at_pointer (GTK_MENU (label->popup_menu), NULL);
+            gtk_menu_popup_at_pointer (GTK_MENU (label->popup_menu), (const GdkEvent*) info->event);
             gtk_menu_shell_select_first (GTK_MENU_SHELL (label->popup_menu), FALSE);
         }
     }
@@ -3125,11 +3128,13 @@ eel_editable_label_do_popup (EelEditable
     {
         info->button = event->button;
         info->time = event->time;
+        info->event = event;
     }
     else
     {
         info->button = 0;
         info->time = gtk_get_current_event_time ();
+        info->event = event;
     }
 
     gtk_clipboard_request_contents (gtk_widget_get_clipboard (GTK_WIDGET (label), GDK_SELECTION_CLIPBOARD),
```

monsta:
I also wonder if the `else` block here ever worked:

```C
        if (info->button)
            gtk_menu_popup_at_pointer (GTK_MENU (label->popup_menu), NULL);
        else
        {
            gtk_menu_popup_at_pointer (GTK_MENU (label->popup_menu), NULL);
            gtk_menu_shell_select_first (GTK_MENU_SHELL (label->popup_menu), FALSE);
        }
```

To get there, you need `info->button` to be 0, which is when `event` in `eel_editable_label_do_popup` is `NULL`, but I don't think that could happen. That callback is only activated on right click, so the `event` should be always filled.

raveit65:
I just tested it again.
```
(caja:32721): Gdk-WARNING **: 13:05:21.568: Event with type 19 not holding a GdkSeat. It is most likely synthesized outside Gdk/GTK+
```
Should i revert this part of the commit?
I am in worry that forgot this when i delete notification mails from my mailbox.

PR #992, opened 2018-05-08
replace deprecated gtk_menu_popup function part 2

raveit65:
zoom-contol:
In favor of constant using of gtk_menu_popup_at_widget for mouse and keyboard events.
Now the popup is under the zoom-button if using mouse.

eel-gtk-extensions: 
open all context menus under pointer, same like before.

### DIFF
--- a/eel/eel-gtk-extensions.c
+++ b/eel/eel-gtk-extensions.c
@@ -287,43 +287,12 @@ eel_gtk_window_set_initial_geometry_from_string (GtkWindow *window,
  * @event: The event that invoked this popup menu.
  **/
 void
-eel_pop_up_context_menu (GtkMenu	     *menu,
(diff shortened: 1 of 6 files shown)


lukefromdc:
The second commit in this introduced a new issue I did not notice right away:
https://github.com/mate-desktop/caja/issues/996
"Poor response from right click on pathbar since c4777b09e3e2b2184895c828c05cb9f1e290fd31"
Reverting this commit fixes the right-click menu response, so we've got a bit more to do here. I should have noticed this in testing, but without seeing all the places this was used I didn't see that it affected the right-click menus. I have not found any other problems from this, just this one.

lukefromdc:
https://github.com/mate-desktop/caja/pull/997
Fixes that issue and is a simple revision

PR #993, opened 2018-05-10
eel-editable-label: avoid deprecated gtk_style_context_set_background

sc0w:
I just deleted the related lines, it seems we don't need it

Do you see collateral effects with this PR?

### DIFF
--- a/eel/eel-editable-label.c
+++ b/eel/eel-editable-label.c
@@ -1230,17 +1230,6 @@ eel_editable_label_style_updated (GtkWidget *widget)
 
     /* We have to clear the layout, fonts etc. may have changed */
     eel_editable_label_recompute (label);
-
-    /* Set the background, foreground and cursor colors based on
-     * the new theme selected.
-     */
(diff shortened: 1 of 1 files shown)


lukefromdc:
I'm guessing with some older GTK version the background of the editable label didn't change with the theme unless it was included, but would have to find out where in the commit history of Caja (or maybe Nautilus) it came from. No need to bother with that if this works without a behavior change or warnings.

raveit65:
Isn't editable-label not set by caja.css?

lukefromdc:
Caja.css came in with https://github.com/mate-desktop/caja/commit/e8be955849664584f1fb697fd904ae9f85945a5e#diff-6cb8f9267fd44f3c78122a6551b19347
in 2016, so before it and in GTK2 we would have needed something else.

raveit65:
I mean we use a background-image and font color for desktop-view
https://github.com/mate-desktop/caja/blob/master/data/caja.css#L23
So at this point you can't change something with themes.

Issue #996, opened 2018-05-15
Poor response from right click on pathbar since c4777b09e3e2b2184895c828c05cb9f1e290fd31


lukefromdc:
#### Expected behaviour
First right-click on pathbar buttons should open right-click menu

#### Actual behaviour
First click is usually swallowed unless same pathbar button has had right-click menu open before
if Caja build includes commit
 https://github.com/mate-desktop/caja/commit/c4777b09e3e2b2184895c828c05cb9f1e290fd31

#### Steps to reproduce the behaviour
1: Build and install Caja from git master as of 5-14-2018 or otherwise including 
 https://github.com/mate-desktop/caja/commit/c4777b09e3e2b2184895c828c05cb9f1e290fd31

Open a directory in Caja, and try to right-click on any pathbar button. Usually you will have to right-click twice

#### MATE general version
1.20/1/21 mix, packages from git master as of 5-14-2018

#### Package version
Local builds from git master

#### Linux Distribution
Debian Unstable

#### Link to downstream report of your Distribution
None as these are local builds

PR #997, opened 2018-05-15
Pathbar button: fix poor response to right-click

lukefromdc:
Fix https://github.com/mate-desktop/caja/issues/996


lukefromdc:
Please check closely for any problems in other popup menus. This fixes the problem in the pathbar menus, the code was taken from elsewhere in the gtk_menu_popup deprecation fix commits

lukefromdc:
Rebased after latest change in master

monsta:
We could probably replace this function with `gtk_menu_popup_at_pointer` call (or `gtk_menu_popup_at_widget` if positioning is needed). Similar thing was done in https://github.com/mate-desktop/caja/commit/42e5c91bee1d9f7ae88d291187f6f792c0f36145 and it seems to work.

The only addition here is these two lines:
```
g_object_ref_sink (menu);
g_object_unref (menu);
```
No idea if the comment for this function is still correct or not. It was written in 2008...

raveit65:
I tried to use `gtk_menu_popup_at_widget` with origin PR first, this was really bad idea.
The result is that some menus opening under the sidebar or icon-view widgets, in other words outside of caja window.
Than i decided to use gtk_menu_popup_at_pointer again ;)

Btw. my main box is under maintenance for a while and i need to setup my notebook for testing PRs, building, etc first..

lukefromdc:
Thus, what I did in the PR was simply to pass the event, which is already passed to the function calling gtk_menu_popup_at_pointer and I have had no trouble with it. I have had it in daily use locally since I came up with it.

lukefromdc:
Rebased after https://github.com/mate-desktop/caja/pull/998

PR #1004, opened 2018-06-01
avoid deprecated 'gtk_widget_override_font'

sc0w:
I just removed related code, I don't see differences

### DIFF
--- a/eel/eel-gtk-extensions.c
+++ b/eel/eel-gtk-extensions.c
@@ -357,7 +357,11 @@ eel_gtk_label_make_bold (GtkLabel *label)
      * theme or user prefs, since the font desc only has the
      * weight flag turned on.
      */
-    gtk_widget_override_font (GTK_WIDGET (label), font_desc);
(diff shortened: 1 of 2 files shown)


lukefromdc:
With this PR:
info_sidebar_pr
With current master:
info_sidebar_master

raveit65:
I think we should follow gtk+ suggestions?
https://developer.gnome.org/gtk3/stable/GtkWidget.html#gtk-widget-override-font
```
gtk_widget_override_font has been deprecated since version 3.16 and should not be used in newly-written code.
This function is not useful in the context of CSS-based rendering. If you wish to change the font a widget uses to render its text you should use a custom CSS style, through an application-specific GtkStyleProvider and a CSS style class.
```

sc0w:
@lukefromdc @raveit65 please test/review again :)

lukefromdc:
Rebased on master as of now for clean merge

sc0w:
Regression: press F9 several times and see:

bug_broken_image

PR #1007, opened 2018-06-03
caja-progress-info.c: avoid deprecated 'gtk_status_icon_get_visible' and show notifications

sc0w:
avoid deprecated:

gtk_status_icon_new_from_icon_name
gtk_status_icon_set_tooltip_text
gtk_status_icon_get_visible
gtk_status_icon_set_visible

I just removed the related code, someone can see differences in the behavior?

**EDIT:** PR modified, see the new title

### DIFF
--- a/libcaja-private/caja-progress-info.c
+++ b/libcaja-private/caja-progress-info.c
@@ -594,28 +594,31 @@ static void
 update_status_icon_and_window (void)
 {
     char *tooltip;
-    gboolean toshow, window_shown;
+    gboolean toshow;
+    static gboolean window_shown = TRUE;
 
     tooltip = g_strdup_printf (ngettext ("%'d file operation active",
                                          "%'d file operations active",
                                          n_progress_ops),
                                n_progress_ops);
+
     gtk_status_icon_set_tooltip_text (status_icon, tooltip);
     g_free (tooltip);
 
     toshow = (n_progress_ops > 0);
-    window_shown = gtk_status_icon_get_visible (status_icon);
 
     if (!toshow && window_shown)
     {
         gtk_status_icon_set_visible (status_icon, FALSE);
         gtk_widget_hide (get_progress_window ());
+        window_shown = FALSE;
     }
     else if (toshow && !window_shown)
     {
         gtk_widget_show_all (get_progress_window ());
         gtk_status_icon_set_visible (status_icon, TRUE);
         gtk_window_present (GTK_WINDOW (get_progress_window ()));
+        window_shown = TRUE;
     }
 }
 
--- a/libcaja-private/caja-progress-info.c
+++ b/libcaja-private/caja-progress-info.c
@@ -595,7 +595,13 @@ update_status_icon_and_window (void)
 {
     char *tooltip;
     gboolean toshow;
-    static gboolean window_shown = TRUE;
+    GIcon *icon;
+    GNotification *notification;
+    static gboolean window_shown = FALSE;
+
+    notification = g_notification_new ("caja");
+    icon = g_themed_icon_new ("system-file-manager");
+    g_notification_set_icon (notification, icon);
 
     tooltip = g_strdup_printf (ngettext ("%'d file operation active",
                                          "%'d file operations active",
@@ -607,17 +613,25 @@ update_status_icon_and_window (void)
 
     toshow = (n_progress_ops > 0);
 
-    if (!toshow && window_shown)
+    if (!toshow)
     {
         gtk_status_icon_set_visible (status_icon, FALSE);
-        gtk_widget_hide (get_progress_window ());
-        window_shown = FALSE;
+
+        if (window_shown)
+        {
+            gtk_widget_hide (get_progress_window ());
+            g_notification_set_body (notification, _("Process completed"));
+            g_application_send_notification (g_application_get_default (), "caja", notification);
+            window_shown = FALSE;
+        }
     }
     else if (toshow && !window_shown)
     {
         gtk_widget_show_all (get_progress_window ());
         gtk_status_icon_set_visible (status_icon, TRUE);
         gtk_window_present (GTK_WINDOW (get_progress_window ()));
+        g_notification_set_body (notification, _("Working..."));
+        g_application_send_notification (g_application_get_default (), "caja", notification);
         window_shown = TRUE;
     }
 }


sc0w:
@lukefromdc yes, confirmed

someone knows the best replacement for gtkstatusicon?

libappindicator uses gtkstatusicon too, so, it is useless

we need to do something here to work with gtk 4

I added some changes in latest commit, it shows notifications in the begin and end of the work with caja.

what do you think? do you like latest changes?

raveit65:
Do we really need a StatusIcon?
You see it only if you copy a large file on a slow disk, eg. with a usb-2.0 port.
And it has only 2 functions:
1. Clicking on it hide/show the progress window.
2. It shows a tooltip

Gtk+ suggest to use App-Notifications for displaying notifications, see 2 page of gtk-widget-factory.

lukefromdc:
A notification takes a lot of desktop space, and times out. Some copy jobs can be quite long., as in the case of multi-GB camera card copy jobs. I can accept others deciding to remove this but would rather keep it.

lukefromdc:
If we start removing status-icons from programs, we lose much of the mate-panel tray unless we do the work to convert everything to indicators. There is no reason to do this other than a Wayland transition, in which case the Xembed issue comes up. GNOME deprecated status icons years ago,  many accused them of doing so simply because they did not use them in GNOME anymore. I don't know if they were looking towards Wayland or just trying to force app developers to "upgrade" to their new UI standards. Anyway, about to test the notifications for function and build.

lukefromdc:
The notifications work, but as I expected time out before a really long file transfer job can complete. In my experience that little tray icon can prevent data loss. It is a warning not to log out and not to do anything that might trigger a known bug and crash Caja.  If we were rid of all the crashes, just adding the text "Do not log out or close the file manager" to the notification would probably be enough to avoid data loss.

The other nice thing about the status icon is that by clicking it while on a different desktop than the transfer job you can bring up the progress dialog on the current desktop. As we now have interactive notifications, adding such an option and making the notification persistant would allow the same functionality as before, though it would not be as pretty. 

The other option is an indicator, though since clicking an indicator can only bring up a menu, moving/opening the progress dialog would be in response to clicking on that menu item, as is the case in mate-polkit for dropping priviliges. Ayatana indicators are cross-distro and are maintained. No requirement for Xembed with them so long as the tray is built in-process.

lukefromdc:
I will leave the decision on this to the rest of you, my vote is to keep the status icon for now  but I won't block the rest of the team doing something different. It does not rise to the "would revert locally in builds" level.

lukefromdc:
Rebased against current master for the convenience of everyone working on this

monsta:
Please leave the poor status icon alone. We're not migrating to GTK+4 tomorrow, right?

raveit65:
I don't want to remove it. I don't use it personally but i can imagine that it is useful for some use cases.
Maybe we should fork GtkStatusIcon code?
Than we are independent from worse SNI code.

vkareh:
@raveit65 - that's not that crazy of an idea. It's a pattern that we use throughout the desktop, and the code is fairly straightforward (most of what we use ends up being wrappers for around GtkWidget functions). I'd even remove the WIN32 and QUARTZ backends, which decreases code complexity (we don't really support Win/Mac, do we?).

Considering that the other option is moving towards a gnome-centric desktop, I'd rather fork it and keep to a traditional model. Which means having a proper tray bar with tray/status icons.

That said, are there any other ways of providing a tray icon that does not need GtkStatusIcon? I mean, in the end it's just a GtkWidget with other things around it. I think the main problem ends up being embedding it there, right?

vkareh:
I see the GTK team's reason for deprecating status icons. If you don't have a tray, your user will not get notifications. If we fork this thing, we should probably account for that, where we check whether the gtk_status_icon_is_embedded, and if not then we can show a notification instead. Based on that, I wonder if this fork should be part (or at least integrate tightly) with the mate-notification-daemon... or something along those lines.

raveit65:
> I'd even remove the WIN32 and QUARTZ backends, which decreases code complexity (we don't really support Win/Mac, do we?).

I don't think :)

And other applications needs to rename GtkStatusIcon Code if we do that.

raveit65:
Ok, this PR removes the applet. I am undecided , i will leave this decision to  you :)
But what i really like is the notification Code, this should really be added to master.
@sc0w ^^^^
Btw. i will add this to fedora if you don't want to test this because you don't like it ;)

vkareh:
@raveit65, @monsta, @lukefromdc - created a feature request for this: #1012 

This pull request works well and follows the current GTK guidelines. Since maintaining the status icon becomes costly once GtkStatusIcon is entirely removed, this should become a new feature for us to maintain independently. Hence #1012

lukefromdc:
What about converting from a status icon to a mate-panel applet? Main issue here would be making sure the applet can be removed and a user knows it has been added when adding it, as it would display no icon when Caja is not moving large files.  The code to add and remove it could simply be put into Caja itself to get around that, same thing could be done with the m-s-d applet.  

Given the number of places we use the tray though, I agree we might be better off to fork the code out of GTK, rename it, and put it into mate-desktop as the first symbols in "mate-gtk-extensions." We would do this with GtkStatusIcon, GtkPlug, and GtkSocket (and all related functions of them), and if we ever do a full Wayland port at that time would need to implement a "Wembed" that works like Xembed.  The tray would need to support GtkStatusIcons as well as MateStatusIcons so long as 3ed party apps still use them, but we would have independence from GNOME on this question.

The root of the deprecation is that since GNOME does not use tray icons, they don't want to be bothered with maintaining them or porting them to Wayland.  Taking this code over is the only way we will ever make it work in Wayland.

monsta:
An applet is always on the panel, while status icons may come and go if needed. This status icon is a good example of a real "status" icon: it only appears when Caja needs to tell "I'm busy" to the user.

monsta:
And if Wayland can't do something that X11 can, it's a problem in Wayland, which is actually still WIP.

sc0w:
I agree with merge this PR, the notifications are  better, and the user knows exactly when caja finish the job.

Anyway, I like to keep too the old behavior, but it is not easy, +1 for the fork.

The best choice to forget the bad gtk decisions is to fork the entire gtk 3.22 code, but we are a little team... (a crazy idea?)

vkareh:
It is a crazy idea! It would also mean that anything done with Gtk would eventually drift from Mtk in terms of compatibility.

A really cool project would be a tray applet that's able to hold on to notifications and keep adding icons to itself for each notification type (a dynamic list of app icons, and each of those icons have its own stack of notifications, drawer-style).

This means we can keep up with Gtk and still have a functional notifications tray applet.

Thoughts?

lukefromdc:
Forking all of GTK3 would both be too much work and contribute to fragementation. GNOME has committed to not removing deprecated widgets and symbols from within the same major version number, meaning that things like GtkPlug, GtkSocket, and GtkStatusIcon will never be removed from any GTK 3.22 version, and that if the new GTK 3.24 branch ever goes to a release they will still be present there too.

They are not guaranteed to keep working though if underlying code changes, we've had serious bugs before caused by unmantained deprecated widgets in GTK 3, notably the entire desktop redraw mess we had with the 3.21 development cycle in Caja.

I would recommend forking only things for which GTK offers no replacement that we actually need to use.  In this case, that means GtkStatusIcon, and the underlying GtkPlug and GtkSocket.  Rather than an entire MTK toolkit, we get "mate-gtk-extensions," which I would propose to add to mate-desktop so all MATE apps can use them. We would need these three for caja, mate-settings, daemon,  mate-media, and mate-power-manager at the least. 

 Hell, we've already got eel-glib-extensions.c,  eel-gdk-extensions.c , and eel-gtk-extensions.c in Caja's eel directory, but those are exclusive to eel and thus to Caja (and to Nemo).

As for a whole MTK, that could have become necessary if GNOME had not decided to branch GTK4 and create a stable GTK3. The GSK drawing rewrite would have broken every custom drawn widget. Because that would have trashed the then just about finished Firefox and Libreoffice GTK ports, GNOME thought better of it, so GTK3 is considered the "stable version" that the community is advised to use, and to avoid GTK4.

raveit65:
I wanna bring back the discussion to this applet.
Is this really a classic notification icon like nm-applet, which sends notifications to the notification-daemon?
If you say that only displaying the applet is a notification than it is.
But without the PR this applet does not send any notification about the status of a file-transfer to m-n-d.
Currently the applet does 3 jobs
1. "Hello i am there and a transfer job is running" --> The button of the progress-window in wnck-list-applet does the same.
2. Hide/show the progress-window --> The button of the progress-window in wnck-list-applet does the same.
3. A tool-tip , which only says that a undefined copy task is running, which is the same like (1).

Maybe i miss another important function, but personal i think that the PR does the 'real notification job' much better than current implementation.
But i agree to leave the applet there if a member of the core team like to use it, and we do not really have the gtk4 pressure for the moment.

I general i think that applets which do not have a notification function like volume-applet or xrandr-applet should be ported to a real panel-applet. 
Keep in in mind that we had a real volume panel-applet before the port to gstreamer-1.0.

@sc0w 
Can the second commit use as standalone without first commit?

sc0w:
done

I removed the function 'gtk_status_icon_get_visible' and I keep the tray icon

@raveit65 @lukefromdc @monsta @vkareh 

please test

raveit65:
Does what it promised.

sc0w:
@lukefromdc I don't need 'gtk_status_icon_get_visible', is more efficient to use one simply variable in my opinion

lukefromdc:
I agree it's cleaner code the way it was changed, merging

PR #1005, opened 2018-06-01
avoid deprecated 'gtk_window_set_wmclass'

sc0w:

PR #1011, opened 2018-06-05
caja-desktop-window.c: Fix: caja crashes unexpectedly

sc0w:
Fixes https://github.com/mate-desktop/caja/pull/1005#issuecomment-394805519

I don't know what means, but I keep the same comment here as before:
https://github.com/mate-desktop/caja/pull/1005/files#diff-5b74cc5e4b5c15d0430f0921ed104c3bL155


lukefromdc:
This stopped the startup crashes here. Note that I was only able to force a crash by restarting my entire session though, I was not able to invoke it at will by any other means

raveit65:
I can confirm that this commit fixes the crashes for me.
Thanks for quick reaction :)

PR #1021, opened 2018-06-21
eel-editable-label: show icons in right-click menu only if "menus-have-icons"  is set

lukefromdc:


lukefromdc:
@TomaszGasior , would you please test this as well?

sc0w:
the same I commented in https://github.com/mate-desktop/mate-panel/pull/820#issuecomment-399212825 :

To make sure we have spaces in the left, I suggest to do the changes here, with something like this:

```
if ((icon_name) && (g_settings_get_boolean (icon_settings, "menus-have-icons")))
        icon = gtk_image_new_from_icon_name (icon_name, GTK_ICON_SIZE_MENU);
else
        icon = gtk_image_new ();
```

and I am not sure if we will have problems with dbus with several clicks (remember https://github.com/mate-desktop/mate-terminal/issues/233)

lukefromdc:
I have no idea at all how to look for the possible dbus issue you mention, but I have been running this code since last night without problems. Also, I was not aware that keeping the space for the icon is the preferred behavior, I thought the opposite was, I will change the PR's to do that.

lukefromdc:
@sc0w , I just force-pushed to keep it one commit after implementing your suggestion to hold the space when the icons are not used, since you said this was the correct behavior. I will now apply this to the panel work

raveit65:
Works for me, i found one icon left which isn't switchable but i won't post that here.
Because the majority of users like the MATE style.

PR #1023, opened 2018-06-25
caja-location-bar: Fix: change background color based on activity state

sc0w:
and avoid deprecated 'gtk_widget_override_background_color'

The expected with this PR, exactly the same behavior like caja 1.8.2 (gtk2)

With this PR, see the video:

[out-65.ogv.zip](https://github.com/mate-desktop/caja/files/2134993/out-65.ogv.zip)

seems we have that issue since the gtk3 migration, I think, the color never changes, and this PR is the fix

### DIFF
--- a/src/caja-location-bar.c
+++ b/src/caja-location-bar.c
@@ -592,30 +592,63 @@ caja_location_bar_set_location (CajaLocationBar *bar,
     caja_location_bar_update_label (bar);
 }
 
+static void
+override_background_color (GtkWidget *widget,
+                           GdkRGBA   *rgba)
+{
+    gchar          *css;
+    GtkCssProvider *provider;
+
+    provider = gtk_css_provider_new ();
+
+    css = g_strdup_printf ("entry { background-color: %s;}",
+                           gdk_rgba_to_string (rgba));
+    gtk_css_provider_load_from_data (provider, css, -1, NULL);
+    g_free (css);
+
+    gtk_style_context_add_provider (gtk_widget_get_style_context (widget),
+                                    GTK_STYLE_PROVIDER (provider),
+                                    GTK_STYLE_PROVIDER_PRIORITY_APPLICATION);
+    g_object_unref (provider);
+}
+
 /* change background color based on activity state */
 void
-caja_location_bar_set_active(CajaLocationBar *location_bar, gboolean is_active)
+caja_location_bar_set_active (CajaLocationBar *location_bar, gboolean is_active)
 {
+    GtkStyleContext *style;
+    GdkRGBA color;
+    GdkRGBA *c;
+    static GdkRGBA bg_active;
+    static GdkRGBA bg_inactive;
+
+    style = gtk_widget_get_style_context (GTK_WIDGET (location_bar->details->entry));
+
     if (is_active)
-    {
-        /* reset style to default */
-        gtk_widget_override_background_color (GTK_WIDGET (location_bar->details->entry), GTK_STATE_FLAG_NORMAL, NULL);
-    }
+        gtk_style_context_get (style, GTK_STATE_FLAG_NORMAL,
+                               GTK_STYLE_PROPERTY_BACKGROUND_COLOR,
+                               &c, NULL);
     else
-    {
-        GtkStyleContext *style;
-        GdkRGBA color;
-        GdkRGBA *c;
-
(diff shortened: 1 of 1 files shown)


raveit65:
It seems to be working in a quick test.
Is there a chance to add a different style class for inactive location-bar-entry for better theming?
Normal style class (css selector) is `caja-location-entry`.

raveit65:
Hmm, i have some weird background of the location-entry in single mode of caja after a restart of caja (killall caja) with some themes.

src/caja-location-bar.c
```diff
@@ -592,30 +595,61 @@ caja_location_bar_set_location (CajaLocationBar *bar,
...
+static void
+override_background_color (GtkWidget *widget,
+                           GdkRGBA   *rgba)
+{
+    gchar          *css;
+    GtkCssProvider *provider;
+
+    provider = gtk_css_provider_new ();
+
+    css = g_strdup_printf ("entry { background-color: %s; background-image: none;}",
```
raveit65: Why background-image: none ?
The non-active entry should simply use 
entry:disabled {} from themes.

raveit65:
It is possible to use GTK-STYLE-PROPERTY-BACKGROUND-IMAGE too.
https://developer.gnome.org/gtk3/stable/GtkStyleContext.html#GTK-STYLE-PROPERTY-BACKGROUND-IMAGE:CAPS

sc0w:
@raveit65 @vkareh @lukefromdc 

With latest changes, this PR modifies only the `background-color`and it works in all themes, but the user only can see the changes in themes without `background-image` into the `GtkEntry`, for example, in GreenLaguna

One less deprecation, and the fix over the `background-image` can be done into another PR, by the moment, I don't know how to do that

raveit65:
I am fine with not breaking usage of background-image in themes.
And it works with GreenLaguna. But if i press F3 the first time than both locationbars are showing the **active** color.
After switching the panes it works like expected.

Why not setting up a style classes for active and inactive state?
Than we can setup the entries with the theme, eg.
.caja-location-entry {}  --> normal entry in single mode
.caja-location-entry.active {}  --> entry of active pane
.caja-location-entry.inactive {}  --> entry of inactive pane

sc0w:
@raveit65 

> I am fine with not breaking usage of background-image in themes.
> And it works with GreenLaguna. But if i press F3 the first time than both locationbars are showing the active color.
> After switching the panes it works like expected.

Yes, caja 1.8.2 (gtk2) have the same behavior like this PR, so, I think, that is the expected behavior in that function

> Why not setting up a style classes for active and inactive state?
> Than we can setup the entries with the theme, eg.

yes, I agree, but I don't know how to do that :)

vkareh:
Just wanted to mention that the latest update fixes the weird behavior for me. Some themes don't change the color of the text field, but the pencil icon next to it now gets properly grayed out when the pane is inactive.

raveit65:
Forgot my statement about missing style classes.
I think the real issue is that  caja-location-entry still ignore the `disabled` (old = insensitive) state.
An example for using with gtk3-inspector. Only `normal` and `backdrop` state is working.

```
entry {
	background-color: yellow;
	color: black;
}

entry:backdrop {
	background-color: red;
	color: blue;
}

entry:disabled {
	background-color: green;
	color: orange;
}
```

Any way, PR fixes a deprecation and it is a improvement for themes which don't use `backgound-image`.
Ready to go if you want.

sc0w:
@raveit65 

```
entry:disabled {
	background-color: green;
	color: orange;
}
```

It doesn't work because the widget isn't insensitive

It can be done insensitive in that function with `gtk_widget_set_sensitive`, but later you can't go from one pane to the entry of the other pane directly

@vkareh you approve this PR with latest changes?

raveit65:
> It doesn't work because the widget isn't insensitive
>It can be done insensitive in that function with gtk_widget_set_sensitive, but later you can't go from one pane to the entry of the other pane directly

But do i need to click in the location-bar to switch panes?
I click always in the window itself, and the 2 panes feature is something what i use all the time.

sc0w:
Personally, I prefer to keep the two items active

raveit65:
It's OK, using background-image isn't broken and i don't care if there is another color or not for in the non-active location-bar.

Btw, don't use Adwaita-dark theme.....

PR #1029, opened 2018-07-09
[desktop-bg] Fix memory leak during background change events

bitness:
The string returned by eel_bg_get_desktop_color() needs to be freed by
the caller, which wasn't happening before.

Commit b9cf3667 fixed the case where this function was being called
repeatedly, triggering this leak and causing it to rapidly fill system
memory.


bitness:
Thanks!

PR #1031, opened 2018-07-16
caja-desktop-directory: stop segfaults in directory_ready_callback

lukefromdc:
Prevent segfault on g_list_find in directory_ready_callback by catching the NULL case. This segfault can happen on rare occasions with GTK 3.23.0, I've had it several times in a week.

### DIFF
--- a/libcaja-private/caja-desktop-directory.c
+++ b/libcaja-private/caja-desktop-directory.c
@@ -162,6 +162,10 @@ directory_ready_callback (CajaDirectory *directory,
     g_assert (callback_data != NULL);
 
     merged_callback = callback_data;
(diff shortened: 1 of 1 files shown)


raveit65:
Maybe a fix for ?
http://ubuntu-bugs.narkive.com/mx8Jwahg/bug-1492495-new-usr-bin-caja-11-g-list-find-directory-ready-callback-ready-callback-call-call-ready-
https://bugs.launchpad.net/ubuntu/+source/caja/+bug/1443253

raveit65:
During the last weeks i noticed some weird black background when opening caja windows.
I said weird because i am using an nvme drive with my main box, so i was surprised to see that with a fast disk.
This seems to be gone with your PR :smiley:

PR #1042, opened 2018-08-10
Allow independent showing/hiding of backup files

Sargastic:
Caja can hide/show 'hidden or backup' files to the user. This is great, but it should be possible to show hidden or backup files independently. 

Existing functionnality has been split :

- The existing 'show hidden' functions have been altered to process/ignore only hidden files.
- A similar (copied/pasted/few edit) function group has been added to process/ignore backup files.

PR #1044, opened 2018-08-13
Independent management of show/hide for dot (hidden) and tilde (backup) files

Sargastic:
Following comments on PR https://github.com/mate-desktop/caja/pull/1042, this is the same stuff on branch 1.21.

Currently, the "show hidden files" parameter (global in Preferences, or per-directory in View) encompasses both hidden (dot) and backups (tilde) files. This PR splits the functionnality to separately show/hide hidden or backup files.

Note : I'm a very basic Caja user, no bells and whistles. In this use-case, the patches did not trigger any problem. I think there should not be problem for network directories and the like, but did not test this kind of usage.


src/file-manager/fm-directory-view.c
```diff
@@ -7266,6 +7280,44 @@ fm_directory_view_init_show_hidden_files (FMDirectoryView *view)
...
+{
+	CajaWindowShowBackupFilesMode mode;
+	gboolean show_backup_changed;
+	gboolean show_backup_default_setting;
+
+	/*
+	if (view->details->ignore_hidden_file_preferences) {
+		return;
+	}
+	*/
```
vkareh: Remove this commented code

Sargastic:
You may change persistence through Caja preferences. As for the shortcut, I agree but since I do not use Control-B to bookmarks, I remapped it. Any idea of a proper shortcut (Ctrl-k maybe ?)

src/file-manager/fm-tree-model.c
```diff
@@ -956,7 +957,9 @@ should_show_file (FMTreeModel *model, CajaFile *file)
 
     should = caja_file_should_show (file,
                                     model->details->show_hidden_files,
-                                    TRUE);
+				    /* TODO details->show_backup_files */
```
vkareh: This seems like a first attempt that got commented out. Please remove.

vkareh:
@lukefromdc - I agree with the shortcut issue. In testing it I kept wanting to use a shortcut for toggling it, since I'm used to `Ctrl+H`.

lukefromdc:
I did a quick look over the code, memory management seems to be OK.

src/file-manager/fm-tree-model.c
Sargastic: Done (for the three ponts)

vkareh:
@Sargastic - in line 455 of `libcaja-private/caja-file.c`, inside the `caja_file_clear_info()` function, you should probably initialize `is_backup` to `FALSE`:

lukefromdc:
We still have a build warning here:

```
caja-file.c: In function ‘caja_file_should_show’:
caja-file.c:3653:21: warning: implicit declaration of function ‘caja_file_is_backup_file’; did you mean ‘caja_file_is_hidden_file’? [-Wimplicit-function-declaration]
    (show_backup || !caja_file_is_backup_file (file)) &&
                     ^~~~~~~~~~~~~~~~~~~~~~~~
                     caja_file_is_hidden_file
```

but the build finishes. Maybe that function just needs to be moved up in the file.

vkareh:
Good catch on that implicit declaration @lukefromdc - just tested the latest one and it fixes the issue. Also tested the new Ctrl+K shortcut and works very well. From what I can find, Ctrl+K does not have a globally-defined binding, so using it locally for this purpose seems fine to me.

@raveit65, @sc0w - do you have opinions about adding this feature?

raveit65:
Ok, the show-hidden file feature has the same behaviour.
It looks like that using those settings from menu entry overrides settings from gsettings/prefererences-dialog temporarily until caja-browser window is restarted.

raveit65:
Thank you

PR #1050, opened 2018-09-05
Add 'Configurable' extension interface, enabling Configure Extension

20kdc:
This commit just creates a 'CajaConfigurable' interface for extensions to implement, so that things like open-terminal can have configuration UIs.

A particular use-case is open-terminal, which is rather configurable, but this configuration is all hidden, leading to the problem of hidden settings that can make the extension unusable in some cases.

This shouldn't affect any localizable strings, as it just makes existing UI call a function if available.

The reason I added a second signal handler to the extensions list for the separate button is because passing two buttons in various other ways seems simply more likely to cause memory leaks (that said, if there's a better solution, I'll modify the PR accordingly).

I tested this by adding the interface to an extension with a debug-print-stub (so not a full test. This said, it's meant to act somewhat like the About dialog in being an inner g\_dialog\_run, so this shouldn't be too much trouble), and having another extension installed to check the configuration-not-supported case.

IDK if I'll end up PRing an open-terminal configurator later, but this'll be an interface for whoever does.

### DIFF
--- a/libcaja-extension/Makefile.am
+++ b/libcaja-extension/Makefile.am
@@ -34,6 +34,7 @@ libcaja_extension_include_HEADERS = \
 	caja-property-page-provider.h \
 	caja-property-page.h \
 	caja-menu.h \
+	caja-configurable.h \
 	$(NULL)
 
 libcaja_extension_la_SOURCES = \
@@ -50,6 +51,7 @@ libcaja_extension_la_SOURCES = \
 	caja-property-page-provider.c \
 	caja-property-page.c \
 	caja-menu.c \
+	caja-configurable.c \
 	$(NULL)
 
 pkgconfigdir=$(libdir)/pkgconfig
--- a/libcaja-extension/caja-configurable.c
+++ b/libcaja-extension/caja-configurable.c
@@ -1,0 +1,88 @@
+/*
+ *  caja-configurable.c - Interface for configuration
+ *
+ *  Copyright (C) 2003 Novell, Inc.
+ *
+ *  This library is free software; you can redistribute it and/or
+ *  modify it under the terms of the GNU Library General Public
+ *  License as published by the Free Software Foundation; either
+ *  version 2 of the License, or (at your option) any later version.
+ *
+ *  This library is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *  Library General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Library General Public
+ *  License along with this library; if not, write to the Free
+ *  Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ *  Author: 20kdc <gamemanj@hotmail.co.uk>
+ *  Based on caja-menu-provider.c by Dave Camp <dave@ximian.com>
+ *
+ */
+
+#include <config.h>
+#include "caja-configurable.h"
+
+#include <glib-object.h>
+
+/**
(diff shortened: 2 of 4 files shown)


raveit65:
So at the moment there is no extension to test this new code?

20kdc:
Indeed. I was intending to get to it, but 1. I keep getting poked to do other stuff, and 2. It'd be a PR to a separate repository.

lukefromdc:
Memory management of this should be OK, though a second pair of eyes on that would be nice as I am inexperienced with checking memory allocation issues. About to try a test build to see if it builds and runs OK, even with nothing for a functional test

20kdc:
Yes. The configure button should only ever be not-greyed-out for configurable extensions.
For the sake of making this as small as possible to try and avoid any mistakes getting through in what is something of a chicken & egg problem for the caja maintainers, the Configurable interface just has a function that is more or less tied to the configure button.

20kdc:
Aha, found the testcase I was using.
Note: It's a testcase, and thus doesn't actually implement a useful dialog.
It's just useful to make sure there's no leaks/etc. in the relevant Caja-side code.

```
diff --git a/open-terminal/caja-open-terminal.c b/open-terminal/caja-open-terminal.c
index 449ace6..4ac6544 100644
--- a/open-terminal/caja-open-terminal.c
+++ b/open-terminal/caja-open-terminal.c
@@ -28,6 +28,7 @@
 #include "caja-open-terminal.h"
 
 #include <libcaja-extension/caja-menu-provider.h>
+#include <libcaja-extension/caja-configurable.h>
 
 #include <glib/gi18n-lib.h>
 #include <gtk/gtkicontheme.h>
@@ -525,6 +526,12 @@ caja_open_terminal_get_file_items (CajaMenuProvider *provider,
 	}
 }
 
+static void
+caja_open_terminal_run_config (CajaConfigurable *provider)
+{
+    g_message ("Hello? RunConfig");
+}
+
 static void
 caja_open_terminal_menu_provider_iface_init (CajaMenuProviderIface *iface)
 {
@@ -532,6 +539,12 @@ caja_open_terminal_menu_provider_iface_init (CajaMenuProviderIface *iface)
 	iface->get_file_items = caja_open_terminal_get_file_items;
 }
 
+static void
+caja_open_terminal_configurable_iface_init (CajaConfigurableIface *iface)
+{
+	iface->run_config = caja_open_terminal_run_config;
+}
+
 static void 
 caja_open_terminal_instance_init (CajaOpenTerminal *cvs)
 {
@@ -569,6 +582,12 @@ caja_open_terminal_register_type (GTypeModule *module)
 		NULL
 	};
 
+	static const GInterfaceInfo configurable_iface_info = {
+		(GInterfaceInitFunc) caja_open_terminal_configurable_iface_init,
+		NULL,
+		NULL
+	};
+
 	terminal_type = g_type_module_register_type (module,
 						     G_TYPE_OBJECT,
 						     "CajaOpenTerminal",
@@ -578,4 +597,9 @@ caja_open_terminal_register_type (GTypeModule *module)
 				     terminal_type,
 				     CAJA_TYPE_MENU_PROVIDER,
 				     &menu_provider_iface_info);
+
+	g_type_module_add_interface (module,
+				     terminal_type,
+				     CAJA_TYPE_CONFIGURABLE,
+				     &configurable_iface_info);
 }
```

raveit65:
> Indeed. I was intending to get to it, but 1. I keep getting poked to do other stuff, and 2. It'd be a PR to a separate repository.

A second PR for testing would be great.

20kdc:
Ok, that should help for testing.

lukefromdc:
I can now verify that this works with 
https://github.com/mate-desktop/caja-extensions/pull/43
and that the checkbox to select opening a terminal in /home if opened from the desktop successfully toggles this behavior. Looks good to go on this end.

Issue #1027, opened 2018-07-04
Open caja with several tabs opened


colibris79:
It is not properly a bug but something I want to know if it is possible to do. I would like to be able to open Caja with several tabs opened. For example, when  I open Caja I want it to open with my "Home", "Downloads" and "Documents" folders opened in the same window but in different tabs. Is there any way to do this? Via terminal I just manage to open the three folders but in different windows.
Thanks

#### MATE general version
1.18

#### Linux Distribution
Mint 18.3 Mate

monsta:
Implemented in 1.22 (https://github.com/mate-desktop/caja/pull/1053). Use `-t` switch.

PR #1053, opened 2018-09-14
Add option to open multiple uri in tabs at startup

klemsgithub:
Hi all, this is my first contribution to opensource world.
I've implemented a feature that i needed for a long time : to be able to open uri at startup in tabs instead of new windows.

Hope it will be usefull to someone else.

it matches to feature request #1027

### DIFF
--- a/src/caja-application.c
+++ b/src/caja-application.c
@@ -221,7 +221,10 @@ caja_empty_callback_to_ensure_read() {
 
 static void
 open_window (CajaApplication *application,
-         GFile *location, GdkScreen *screen, const char *geometry, gboolean browser_window)
+             GFile *location, 
+             GdkScreen *screen, 
+             const char *geometry, 
+             gboolean browser_window)
 {
     CajaApplication *self = CAJA_APPLICATION (application);
     CajaWindow *window;
@@ -239,9 +242,7 @@ open_window (CajaApplication *application,
     if (browser_window ||g_settings_get_boolean (caja_preferences, CAJA_PREFERENCES_ALWAYS_USE_BROWSER)) {
         window = caja_application_create_navigation_window (application,
                  screen);
-    }
-
-    else {
+    } else {
         window = caja_application_get_spatial_window (application,
                  NULL,
                  NULL,
@@ -254,8 +255,7 @@ open_window (CajaApplication *application,
 
     if (geometry != NULL && !gtk_widget_get_visible (GTK_WIDGET (window))) {
         /* never maximize windows opened from shell if a
-         * custom geometry has been requested.
-         */
+         * custom geometry has been requested. */
         gtk_window_unmaximize (GTK_WINDOW (window));
         eel_gtk_window_set_initial_geometry_from_string (GTK_WINDOW (window),
                                  geometry,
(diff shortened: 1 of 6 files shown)


raveit65:
How is the syntax for command line?
caja --tabs=uri1 --tabs=uri2 ......?
Or comma separated?

klemsgithub:
I just added --tabs or -t option and i use already implemented list of URI.
Without -t multiple uri open in new windows like before, with -t it open in tabs in a single window

lukefromdc:
This is on a private branch, not the main project repo so I can't push a PR but here's the fix for a memory leak:

In src/caja-application.c in caja_application_open, you added  

`       gchar** splitedOptions = g_strsplit (options, &splitter, 3);`

At line 373. This needs to be freed with g_strfreev() when you are done with it. I tested that and it worked.  Here's caja_application_open with
`g_strfreev(splitedOptions);`
added at line 384:

```
static void
caja_application_open (GApplication *app,
               GFile **files,
               gint n_files,
               const gchar *options)
{
    CajaApplication *self = CAJA_APPLICATION (app);
    gboolean browser_window = FALSE;
    gboolean open_in_tabs = FALSE;
    const gchar *geometry = NULL;
    const char splitter = '=';

    g_debug ("Open called on the GApplication instance; %d files", n_files);

    /*Check if local command line passed --browser, --geometry or --tabs*/
    if (strlen(options) > 0){
        gchar** splitedOptions = g_strsplit (options, &splitter, 3);
        sscanf(splitedOptions[0], "%d", &browser_window);
        if(strcmp(splitedOptions[1], "NULL")!=0){
            geometry = splitedOptions[1];
        }
        sscanf(splitedOptions[2], "%d", &open_in_tabs);

        /*Reset this or 3ed and later invocations will use same
	     *geometry even if the user has resized open window
         */
        self->priv->geometry = NULL;
        g_strfreev(splitedOptions);
    }

   open_windows (self, files,
              gdk_screen_get_default (),
              geometry,
              n_files,
              browser_window,
              open_in_tabs);
}


```

klemsgithub:
thank you, i've added the fix.
It's my first time using github, what is the next step of the process?
Should i merge it on the master on my forked repository?

lukefromdc:
No, it is only needed in your  open_in_tabs branch, which now includes it.

lukefromdc:
Merging this here will now include both your commits

klemsgithub:
ok, i just removed a useless g_print i had left in the code.

src/caja-application.c
```diff
@@ -267,26 +267,90 @@ open_window (CajaApplication *application,
     g_free (uri);
 }
 
+static void
+open_tabs (CajaApplication *application,
```
raveit65: indents

src/caja-application.c
```diff
@@ -267,26 +267,90 @@ open_window (CajaApplication *application,
...
+         GFile **locations, guint n_files, GdkScreen *screen, const char *geometry, gboolean browser_window)
+{
+    CajaApplication *self = CAJA_APPLICATION (application);
+    CajaWindow *window;
```
raveit65: indents

src/caja-application.c
```diff
@@ -267,26 +267,90 @@ open_window (CajaApplication *application,
...
-        open_window (application, NULL, screen, geometry, browser_window );
+        open_window (application, NULL, screen, geometry, browser_window);
     } else {
-        /* Open windows at each requested location. */ 
-        i = 0;
-        while (i < len ){ 
-              open_window (application, files[i], screen, geometry, browser_window);
-               i++ ;
-         }
+        if(open_in_tabs){
```
raveit65: same as before

src/caja-application.c
```diff
@@ -298,40 +362,41 @@ caja_application_open (GApplication *app,
...
-        if (g_str_match_string ("browser",
-                    options,
-                    FALSE) == TRUE){
-            browser_window = TRUE;
-            geometry = strchr(options, splitter);
-        }
-        else {
-        geometry = options;
+    /*Check if local command line passed --browser, --geometry or --tabs*/
+    if (strlen(options) > 0){
```
raveit65: same as before

src/caja-application.c
```diff
@@ -267,26 +267,90 @@ open_window (CajaApplication *application,
...
+        } 
+    }
+
+    if (geometry != NULL && !gtk_widget_get_visible (GTK_WIDGET (window))) {
+        /* never maximize windows opened from shell if a
+         * custom geometry has been requested.
+         */
+        gtk_window_unmaximize (GTK_WINDOW (window));
+        eel_gtk_window_set_initial_geometry_from_string (GTK_WINDOW (window),
+                                 geometry,
```
raveit65: indents

raveit65:
Are you familiar with the 'git rebase ' command for editing commits?

klemsgithub:
nop, but i can learn.

klemsgithub:
Just to be sure, the point is to have only one commit in the history that regroups the original code and all the fix?

klemsgithub:
humm... i used git rebase to regroup all my commit into one but when i put it on remote repository git have forced me to remerge with old commits...

raveit65:
True in this case, here most commits are a correction for the first comment.
In general you can squash/edit/rename commit with that command.
Start with `git rebase -i <last-good-commit-before-you-want-to-edit-git-history>`
which is
`git rebase -i 3619f72800e135973f3a2bf615fe7827ab06273d`
This will open a vi editor in terminal, which isself-explanatory
After closing the editor do
`git push -f` to update you remote repo with your local repo. Be careful, you can't undo this step.
PR will be automatic updated.

raveit65:
git push -f (force) is the secret ;)

klemsgithub:
thank you, i've learn something useful :)

raveit65:
Np, 'git rebase' is a very useful tool ;)
But you forgot some intends , you don't see them ?  :)
https://patch-diff.githubusercontent.com/raw/mate-desktop/caja/pull/1053.patch

klemsgithub:
only one, is it better?
i've some issues with indents as the original code was not uniformly indented so i can't decide what is for exemple the way to indent calls on severals lines (sometimes we have to align parameters with the first one, sometimes to the start of the function name... i'm lost) .

src/caja-application.c
klemsgithub: here it's the old code which was wrong idnented

raveit65:
Yeah, i know old code has a lot of indents issue, after a lot of ppl was working on that code after all that years of gnome and mate.
Personal i try to avoid indents errors when i add complete new code.
I old code blocks it's better to use old failure indents in the first commit and do another commit for fixing them.
In result you know better what the code change is.
Beside from that your PR works like a charm. :)
Is it possible to edit man page with the new option?
Of course this can be a single comment.

src/caja-application.c
```diff
@@ -267,71 +270,138 @@ open_window (CajaApplication *application,
...
+           gboolean browser_window)
+{
+    CajaApplication *self = CAJA_APPLICATION (application);
+    CajaWindow *window;
+    gchar *uri = NULL;
+
+    /* monitor the preference to use browser or spatial windows */
+    /* connect before trying to read or this preference won't be read by root or after change */
+    g_signal_connect_swapped(caja_preferences, "changed::"CAJA_PREFERENCES_ALWAYS_USE_BROWSER,
+                      G_CALLBACK (caja_empty_callback_to_ensure_read),
```
raveit65: indents

klemsgithub:
done :)

src/caja-application.c
raveit65: I meant this new line `if(open_in_tabs){`  --> `if (open_in_tabs) {`

src/caja-application.c
```diff
@@ -2143,30 +2216,18 @@ caja_application_local_command_line (GApplication *application,
...
+    /*Set up --geometry, --browser and --tabs options  */
     /*Invoke "Open" to create new windows */
-
-    if (browser_window == TRUE && self->priv->geometry == NULL){
-
-        if (len > 0)  {
-        g_application_open (application, files, len, "browser");
+    if (len > 0)  {
+        gchar* concatOptions = g_malloc0(64);
+        if(self->priv->geometry==NULL){
```
raveit65: sometimes you are using a space and sometimes not :)
`if(self->priv->geometry==NULL){` --> `if (self->priv->geometry==NULL) {`

src/caja-application.c
```diff
@@ -2143,30 +2216,18 @@ caja_application_local_command_line (GApplication *application,
...
-    else if (browser_window == TRUE && self->priv->geometry != NULL){
-        if (len > 0)  {
-            g_application_open (application, files, len, (g_strconcat("browser","=",
-                                self->priv->geometry, NULL)));
-        }
-    }
-
-    else {
+        g_application_open (application, files, len, concatOptions);
+        g_free(concatOptions);
```
raveit65: same here

src/caja-application.c
raveit65: and the lines below

src/caja-window.h
```diff
@@ -136,6 +136,8 @@ void             caja_window_disconnect_content_view (CajaWindow    *window,
 
 void             caja_window_go_to                (CajaWindow    *window,
         GFile             *location);
+void             caja_window_go_to_tab                (CajaWindow    *window,
```
raveit65: `(CajaWindow    *window,` should be under `(CajaWindow    *window,`, or not?

klemsgithub:
done

lukefromdc:
I finally managed to fix the conflicts from that last commit (which I had forgotten to pull before merge due to https://github.com/mate-desktop/caja/pull/1053#issuecomment-421781562), and commit this as three commits. Probably should have run git rebase -i and squashed them, but had already gotten into trouble by pushing the PR without the last round of changes that generated the conflict, then having to resolve the conflict manually for the first time. Hope I got them right, sorry about the mess.
Commits are
https://github.com/mate-desktop/caja/commit/fd1f45f308e418d1c1b44b968ce83f77e4f8a689
https://github.com/mate-desktop/caja/commit/04f7e80725c61ab511e2a4ead45eb31acc026e1a
https://github.com/mate-desktop/caja/commit/424cf4485515a51074c0811880560de360dad784
and as usual I built and tested (twice in this case due to the conflict issue) prior to push

raveit65:
Why not stopping merging something if you made mistakes?
Everything was fine in ths PR with one commit.
Now we have 3 commits, ......

raveit65:
I reverted the 3 single commits and was able to merge the PR local without any problem.
https://github.com/mate-desktop/caja/commit/6a31ce65e4213d6069e5eb515640233cd27c3937

lukefromdc:
The reason I did not stop was it was too late when I spotted the problem. I saw the comment asking if further review was needed and reacted automatically, only afterwards realizing I had merged only the first two commits(from the previous day's pull)  and missed the last update to the PR. 

Total screwup on my part, and I was not about to screw around with a force-push to master due to the risk someone pulling or cloning during the interval in between. I didn't want to risk compounding this error.  Thus I had to fix the last commit and push it separately, given the first two were already pushed.

sc0w:
Juha Leppänen mailed me that there are problems with this PR:

```
1)
caja-application.c
in open_tabs(...) function after /* open all locations */ comment

uri = g_file_get_uri (...) is executed possibly multiple times,
but only one g_free(uri) is executed before returning from function.
Memoryleak?

also at the same location inside the for-loop
variable i cannot be 0,
so only

caja_window_go_to_tab (window, locations[tab]);

is needed from the if-then-else statement.

2)
caja-application.c
in caja_application_local_command_line(...) function
in the new code after comment
/*Set up --geometry, --browser and --tabs options */

there is code like

if (len > 0) {
    code block A: options strings created here
                  and used as last argument in
                  g_application_open(...)
} else {
    if (len > 0) {
        code block B : default empty "" string
                       used as last argument in
                       g_application_open(...)                       
    }
}

If (len >0) is true, code block A executed, else-block never entered.

If (len >0) is false, then-block never entered, else-block is entered,
but code block B gets never executed.

So the whole else-block is redundant and can be removed.
Also looking at the earlier code before these patches,
if (len > 0) was false then no actual action was done either.
```

klemsgithub:
Hi, i completely agree with these remarks!
I've updated my branch with fixes.

lukefromdc:
I didn't spot the loop issue when looking over memory management. I wil test this now.

lukefromdc:
I cannot find a new commit to your branch to fix this, I will now attempt to deal with this myself and open another PR

lukefromdc:
Test #1: Moving the g_free statement works

klemsgithub:
that's weird : https://github.com/mate-desktop/caja/compare/master...klemsgithub:open_in_tabs?expand=1

klemsgithub:
ah ok, i've merged the commit in the already merged correction that's why

lukefromdc:
Anyway, I think we've got the same thing now except possibly for exact positions of characters, so I am about to open a PR with the three fixes.

raveit65:
> Juha Leppänen mailed me that there are problems with this PR:

Why this guy can't comment here at github?
And why he wrote personal emails to devs instead of posting here.
And why not before we merge this?
He did that one time before and spammed my mailbox with a personal mails about linux code problems.
... i don't like that.
Sorry, this guy needs to respect that we use github as working platform.

PR #1056, opened 2018-09-16
1.20

faustf:
is possible when open folder and click in file  have open new window joined , like double commander ???
thankz


raveit65:
This is not the way we work......

PR #1057, opened 2018-09-19
Remove dead code, fix a memory leak

lukefromdc:
Remove two dead code blocks and fix a memory leak left over from last commit

### DIFF
--- a/src/caja-application.c
+++ b/src/caja-application.c
@@ -301,17 +301,15 @@ open_tabs (CajaApplication *application,
     uri = g_file_get_uri (locations[0]);
     g_debug ("Opening new tab at uri %s\n", uri);
     caja_window_go_to (window, locations[0]);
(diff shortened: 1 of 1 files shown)


klemsgithub:
you can add a g_free (uri); just before the for to take into account the first occurence of g_file_get_uri (locations[0]).

lukefromdc:
Done and force-pushed

raveit65:
Please do not merge. I get an abrt alarm from fedora-bug-reporting-tool every time when i start caja as root.
I will give you more info later.

raveit65:
Forgot my last reply, the segfault with `/usr/libexec/xdg-desktop-portal` and starting caja as root, is gone after a session restart.
Btw. i could reproduce the segfault without this PR and the commit before.
So, it seems that everything is fine.

Issue #1019, opened 2018-06-16
Multi selection permissions issue


thwaller:
#### Expected behaviour
When I select multiple files, click properties and change the file permissions, I would expect that all selected files would change to what I have selected.

#### Actual behaviour
It does not actually change any file permissions when multiple files are selected.

#### Steps to reproduce the behaviour
select multiple files, click properties and change the file permissions, Click the close button. There is no other button to click like save or apply.

#### MATE general version
1.20.1

#### Package version
1.20.2-4ubuntu1

#### Linux Distribution
Ubuntu 18.04

#### Link to downstream report of your Distribution
?

NOTES: When I do not select multiple files (select only one file) the steps above work properly and as expected.

vkareh:
The first time I tried this, I thought I was able to reproduce, but every other time it seemed to work. Upon closer inspection, it seems as though there is an intermediate step in between an enabled and a disabled permission. A screenshot might help explain:

![screenshot at 2018-06-17 22-35-41](https://user-images.githubusercontent.com/259780/41515787-03acfc86-727f-11e8-9cf9-a3db0e9ce375.png)

Notice where the pointer is, in the _Others: Write_ permission. This group of files had it disabled, when I clicked the first time I got that symbol (not a checkmark, just a line). When clicking on it a second time, I got the proper checkbox, the _Text view_ at the bottom changed to `-rw-rw-rw-`, and the terminal showed the correct new permission.

So this works, but that intermediate step makes it really confusing. @thwaller - does this represent what happened on your side? Would you be able to verify whether this is what happened to you?

thwaller:
My initial response would have been no, that is not what happened to me. But after looking in detail and using your reply as a guide, I found that it is in fact the same issue. It only happens like this on multiple selections, or a folder selection. Single file selections are empty box or check mark only.

I mentioned my initial response because I did not even notice this at first. It is really hard to notice it if not paying close attention. You expect a check or no check behaviour, and by having the dash to check change while keeping the color change to green the same makes it almost unnoticeable. Would have been nice to change it to a yellow then a green or something....

[...]

Issue #954, opened 2018-03-26
caja seems to treat file permissions differently than the terminal does


gethob1:
I'm using Linux Mint 18 Sarah Mate 64-bit.
The problem is described in full here:
https://forums.linuxmint.com/viewtopic.php?f=47&t=266392&p=1448943

#### Expected behaviour
A file whose parent folder is immutable may be replaced by a copy of another file of the same name via 'cp -rp' command in the terminal, and there will be no permissions errors or issues.  When copying that same file to the same destination using caja, a permissions error wrongly occurs whether using ctrl+click / shift+click drag and drop or right-click context menu / window menu copy and paste.

#### Actual behaviour
The following error is given when copying with caja:
Error while copying [file]
There was an error copying the file into [directory]
Show more details
Error removing old file: Permission denied

#### Steps to reproduce the behaviour
1) create a test folder
2) create a file within the test folder
3) set ownership of each to your user and permissions to your liking, or to match mine exactly 'drwxr-xr-x' for the folder and '-rwxr-xr-x' for the file.
chown [user]:[user] [folder and file], chmod 755 [folder and file]
4) apply 'sudo chattr +i' command to the test folder but not the file within
5) cp -rp the file to another folder ( will be used to copy / overwrite  the original )
6) cp -rp the file back from terminal without root to note copy success
7) drag and drop in caja with cltr+click and shift+click, and try right-click copy and paste and / or window menu copy and paste-- to note error
8) apply 'sudo chattr -i' command to the test folder and repeat #5 - #7 to note the error going away

It seems caja doesn't allow overwriting a file into a folder that's immutable, while the terminal does, which seems to mean there is an inconsistency in how permissions are handled.

#### MATE general version
? It's whatever is bundled with Linux Mint 18 Sarah Mate 64-bit

#### Package version
1.14.0-1+sarah, per synpatic package manager

#### Linux Distribution
Linux Mint 18 Sarah Mate 64-bit

#### Link to downstream report of your Distribution
https://forums.linuxmint.com/viewtopic.php?f=47&t=266392&p=1448943#p1448943

Issue #421, opened 2015-06-07
Caja doesn't  remember password for remote shares


remifl:
Linux Mint 17.1 Rebecca MATE Edition x86

When I need to access files on Windows shares, Caja prompts me for a password and offers the following options:
- Forget password immediately.
- Remember password until you logout.
- Remember password forever.

"Remember password until you logout" is the default choice. I have to enter username and password twice : one to connect to the remote computer, a second to connect to a share

Workaround : select "Forget password immediately" then select " Remember password until you logout" and the information you entered is available until you log out.

Note : Nemo (on Linux Mint 17.1 Cinnamon) is not affected by this problem.
Thanks for your attention

NiceandGently:
Passwords are handle by gnome-keyring.
Are you shure that your session login keyring is unlocked?
You can check that with seahorse.

remifl:
gnome-keyring is running  but that is not the problem.
The problem is caja works with "Remember password until you logout" option as if i have selected "Forget password immediately".
If i browse a remote windows computer, i have to enter login information to access this computer and on every share.
The only way to recover a normal operation is to select "Forget password immediately" then select "Remember password until you logout". In this way, if i browse a remote Windows computer i have to enter login information once.

Note : Nemo (on Linux Mint 17.1 Cinnamon) is not affected by this problem neither nautilus on gnome v2.x.
Note 2 : the problem is the same on Rebecca Mate Edition x64.
best regards.

infirit:
As mentioned earlier, caja does not store passwords, It relies on gvfs to handle filesystems which handles the passwords if needed. gvfs uses libsecret to then store passwords in your (gnome) keyring.

Install seahore and verify the password is actually stored in the default keyring. Also verify the default keyring is unlocked.

remifl:
The problem is not to store the password but how to ask caja to do so.
From a user point of view, if I select an option called "Remember password until you logout", I expect to enter login and password once and navigate on the different shared folders of the remote computer.

It does not work and the dialog prompt me for login information when I select the computer and the first time i select a different shared folder.

The workaround is to click on "Forget password immediately" and immediatly after click on "Remember password until you logout" : after doing this, the password is stored and I can select every shared folders of a remote computer without enter again login and password.

![caja network](https://cloud.githubusercontent.com/assets/12637806/9155973/8b802b42-3eca-11e5-8d87-d12bbde22ae2.jpg)

NiceandGently:
[...]

Issue #748, opened 2017-03-22
Symlink loop crashes caja with a segfault


exostor:
Make a directory. Put a symlink pointing to another symlink pointing back to the original one. Open the directory in caja. It crashes.

$ ls -la
drwxr-xr-x 1223 exostor     exostor    49152 Mar 14 17:22 .
drwxr-xr-x 1223 exostor     exostor    49152 Mar 18 14:40 ..
lrwxrwxrwx    1 exostor exostor    10 Mar 18  14:40 index.html -> sida2.html
lrwxrwxrwx    1 exostor exostor    10 Mar 18  14:41 sida2.html -> index.html

Syslog:
Mar 18 14:49:44 lit-1 kernel: [  188.766170] caja[3356]: segfault at 7ffcff083ff8 ip 00007fc0e2f030c2 sp 00007ffcff084000 error 6 in libgobject-2.0.so.0.4002.0[7fc0e2ed5000+4f000]

raveit65:
Does this make any sense?
Why you  want to create such a madness?

exostor:
Not me, but I have users who do this by accident and then they don't know why caja keeps crashes "all the time"

lukefromdc:
This seems to be a filesystem-level problem,  and it would seem to me that trying to open any such symlink loop would either crash or wose, lock up in an infinite loop whatever tried to open it.  Presumably it is choosing the icon and filename of the symlink that is looped that is bringing down caja. A fix for Caja would be some way to sanity check this and treat it as a broken link, but not sure if that is normally a filemanager or a filesystem (kernel) task.

Problem at the moment is the requirement to either delete the entire directory or remove one of the links from the command line to fix it. Stopping creation of the bad symlinks from within caja is no defense because another program or even a bad script could create then.

exostor:
Nautilus also crashes.
Thunar, dolphin, konqueror and krusader can handle link loops.

So this is only a problem in Nautilus and Caja.

exostor:
@lukefromdc I'm not trying to open the link, I only open the directory.

lukefromdc:
Opening the directory no doubt starts an attempt to determine the correct icon for the link, which requires following the link to its target. This fails on the circular link just as though you had tried to open it.

PR #1058, opened 2018-09-21
Backports from Nautilus

rcaridade145:
I've been looking into Nautilus codebase and problems identified in Caja and i've identified two commits that were missing from Caja.
The first commit should help with permissions issues like #1019 and #954 while the second for issue #421.

### DIFF
--- a/libcaja-private/caja-file.c
+++ b/libcaja-private/caja-file.c
@@ -555,7 +555,7 @@ modify_link_hash_table (CajaFile *file,
 	GList **list_ptr;
 
 	/* Check if there is a symlink name. If none, we are OK. */
-	if (file->details->symlink_name == NULL) {
+	if (file->details->symlink_name == NULL || !caja_file_is_symbolic_link (file)) {
 		return;
 	}
 
--- a/src/caja-connect-server-dialog.c
+++ b/src/caja-connect-server-dialog.c
@@ -1228,6 +1228,12 @@ caja_connect_server_dialog_fill_details_async (CajaConnectServerDialog *self,
 							str);
 			set_flags ^= G_ASK_PASSWORD_NEED_PASSWORD;
 
+                        if (flags & G_ASK_PASSWORD_SAVING_SUPPORTED &&
+			    gtk_toggle_button_get_active (GTK_TOGGLE_BUTTON (self->details->remember_checkbox))) {
+				g_mount_operation_set_password_save (G_MOUNT_OPERATION (operation),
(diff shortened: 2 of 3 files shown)


rcaridade145:
@lukefromdc  can it be this bug - https://bugzilla.gnome.org/show_bug.cgi?id=773937 ?

lukefromdc:
That is something else, a feature request for a full-fledged administration backend. If the user had no permissions with Nautilus of that era, to error out was the correct behavior, particularily if the user had no permissions on grounds of not owning that file.  

Both with master and with this PR, right-clicking as a normal user on a file owned by root and selecting "delete"(delete bypasing trash must be turned on), does not error out (the behavior of Nautilus in the cited bug report), instead you get the normal confirmation dialog to delete the file, and on clicking "yes" the file is deleted even thouth it is owned by root and the user has no write permissions there. Different issue.

rcaridade145:
@lukefromdc still looking into that issue but backported another commit for issue #748

lukefromdc:
I just tested the new commit. Builds and runs fine, though I don't have anything that invokes the original bug to test it against. No new variables so no memory management changes from it. This is also true of the first two commits, so we should not have to worry about introducing new memory leaks with this.

rcaridade145:
@lukefromdc the only thing i can locate is this commit https://gitlab.gnome.org/GNOME/nautilus/commit/5db7a2952a8091cb11ec74e0fd87cd1814252da2

lukefromdc:
Fine by me. Better to break this up into several simple PR's than get things so complex all of the work gets hung up behind a problem with one piece and kills the whole thing.

raveit65:
https://github.com/mate-desktop/caja/issues/421
is unconfirmed since 3 years by several people?
So how this can be fixed by your PR?
And are you affected by this unconfirmed issue?
And why do you think this is a permission issue?

.. i only try to understand your PRs....

raveit65:
> As i'm not familiarized with both codebases my method of working goes through transversing the list of bugs on Mate and trying to match them with a commit from Gnome. Perhaps the titles are not the best and this is not the best approach. If the bug is not fixed I may look more into it. I'll refrain in the future from these backports

So you did test that the issues are gone?

raveit65:
I am tired to be the bad guy here. This should be review by someone other.

vkareh:
I'm closing this in favor of #1071 to preserve better log history. Thanks for finding these backports @rcaridade145

PR #1062, opened 2018-09-28
added rubberband select to list view mode

dsproch:
In "List View" mode, allows the user to perform rubber band selection by starting selection anywhere except the name column of a row.

### DIFF
--- a/src/file-manager/fm-list-view.c
+++ b/src/file-manager/fm-list-view.c
@@ -86,6 +86,7 @@ struct FMListViewDetails
     GtkTreePath *new_selection_path;   /* Path of the new selection after removing a file */
 
     GtkTreePath *hover_path;
+    gboolean hover_hit_test;
 
     guint drag_button;
     int drag_x;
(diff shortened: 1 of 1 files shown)


lukefromdc:
What should we do with this? It needs to be finished (cork the memory leaks as I suggested, and fix the build warning) but is a very nice feature addition. I don't want this to be forgotten

raveit65:
No idea why authors of PRs don't answer any more, or why they think that we can merge PRs with memleaks.
If you know how to fix the leaks open a new fixed PR and close this one.
@dsproch 
Are you alive?

lukefromdc:
OK, I will do that when I get some time, the memory leak fix is quick and easy to merge into a new PR but I also need to fix the build warning

lukefromdc:
On further testing I discovered that two of the three suspected memory leaks were NOT:
1: freeing path_copy causes a crash on clicking any item, so this must be a pointer to existing data 2: freeing cell_renderers causes a crash on clicking any item,, so this must be a pointer to existing data 

Variable l however (which is inside the loop)  could be freed without crashing anything I could find, so that probably was indeed a memory leak. I also found a fix for the build warning.

lukefromdc:
https://github.com/mate-desktop/caja/pull/1070
fixes the build warning and corks up one possible memory leak. The other two were false alarms

Issue #1060, opened 2018-09-27
Speed up Caja as Nemo has been sped up


lukefromdc:
#### Expected behaviour
Caja and Nemo would start up and open files about as fast as one another

#### Actual behaviour
Nemo 4.0 reported to start and open big directories much faster than before (and thus also much faster than current Caja). These optimizations listed at https://blog.linuxmint.com/?p=3644:

    Nemo startup time: 3.488751 seconds
    Folder load time: 2.644505 seconds
    Idle…Folder load time: 2.874406 seconds

    Generate tooltips for icon items on-demand, instead of when the view loads:
    Nemo startup time: 2.510636 seconds
    Folder load time: 1.596948 seconds
    Idle…Folder load time: 1.654450 seconds

    Optimize icon lookups by GIcon:
    Nemo startup time: 2.011817 seconds
    Folder load time: 1.158854 seconds
    Idle…Folder load time: 1.226699 seconds

    Decrease minimum update interval:
    Nemo startup time: 1.797009 seconds
    Folder load time: 0.937992 seconds
    Idle…Folder load time: 0.973764 seconds

#### Steps to reproduce the behaviour
1: Install Caja
2: Install 4.0 version of Nemo (not yet available in many distros but I'll take their word for it)
3: start a MATE session with "icons on desktop" turned off for a fair comparison
4: start up Caja and Nemo, time them both
5: Open the same large directory in both file managers

#### MATE general version
1.21 from git master

#### Package version
1.21.2 from git master

#### Linux Distribution
Debian Unstable

#### Link to downstream report of your Distribution
None, all my MATE packages are local builds

lukefromdc:
I will be looking at porting these over myself

lukefromdc:
These are the commits in question for Nemo:

TOOLTIP GENERATION:

https://github.com/linuxmint/nemo/commit/4f1af958b15aaf3c5e6221ae80f77c1b138a5c02
icon container: Generate tooltips for icon items on-demand, instead 


https://github.com/linuxmint/nemo/commit/28773d105d26c7648e23ad087c5f941a26f284e3
nemo-icon-view-container.c: Connect the tooltip signal per container 



MINIMUM UPDATE INTERVALS:

https://github.com/linuxmint/nemo/commit/61368e3fc33c0d662f45731d6bbb2a28fc5023ca
nemo-view.c: Decrease minimum update interval. 

ICON LOOKUP:

[...]

PR #1065, opened 2018-10-03
Backport two Nemo commits that substantially speed up opening and loading icon views

lukefromdc:
Backport these two Nemo commits to Caja:

https://github.com/linuxmint/nemo/commit/61368e3fc33c0d662f45731d6bbb2a28fc5023ca
nemo-view.c: Decrease minimum update interval. 

https://github.com/linuxmint/nemo/commit/98843e26b48cd3526cacfe90cfa4ba201d1f3aee
nemo-icon-info: Optimize icon lookups by GIcon.

Each of these significantly sped up Nemo in test benchmarks (shown in the commit comments and copied into commit comments in this PR), and my own testing in Caja bears out the speed improvements.

Fixes https://github.com/mate-desktop/caja/issues/1060

### DIFF
--- a/libcaja-private/caja-file.c
+++ b/libcaja-private/caja-file.c
@@ -4517,13 +4517,15 @@ caja_file_get_icon (CajaFile *file,
 		return icon;
 	}
 
-	if (flags & CAJA_FILE_ICON_FLAGS_FORCE_THUMBNAIL_SIZE) {
-		modified_size = size * scale;
-	} else {
-		modified_size = size * scale * cached_thumbnail_size / CAJA_ICON_SIZE_STANDARD;
-	}
 	if (flags & CAJA_FILE_ICON_FLAGS_USE_THUMBNAILS &&
 	    caja_file_should_show_thumbnail (file)) {
+
+		if (flags & CAJA_FILE_ICON_FLAGS_FORCE_THUMBNAIL_SIZE) {
+			modified_size = size * scale;
+			} else {
+			modified_size = size * scale * cached_thumbnail_size / CAJA_ICON_SIZE_STANDARD;
+		}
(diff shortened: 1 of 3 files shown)


lukefromdc:
Just tested icon rendering with HiPDI (window scaling=2), as some of the code from Nemo had to be modified to support our scaling work. Icons looked great, crisp and clean- no problems there

lukefromdc:
With my large directories this makes a big difference, enough here that I find it very noticable and still do after a day of use, and that's on a fast system.

monsta:
It turned out that commit 6c423bc27d2803f3259bcca9f518a1c22b2b5fd1 has a weird side effect. If an extension uses some non-existing icon in the context menu, `text-x-generic` icon will be shown instead. Previously no icon was shown at all.

For our extensions, I've worked around that in https://github.com/mate-desktop/caja-extensions/commit/9535aeeb72d9f0496df0188d9ffb92c1a3939745 and https://github.com/mate-desktop/engrampa/commit/770306b068d930b804cc1a62ce66fc5b87c8c8cd.

lukefromdc:
I spotted that in Engrampa but only noticed it recently,thus didn't realize it was a recent bug. Wondered why I got that icon

monsta:
Interesting that `text-x-generic` was already present in the code before - I guess that code path was never executed. Same for Nemo. This icon name is present in several places in its code, but according to @clefebvre, Nemo shows `image-missing` for non-existing icon names. Could be that `image-missing` is used as a fallback by one of these GTK+ functions.

monsta:
Ah wait... I've misread. Looks like there was no `text-x-generic` in `caja-icon-info.c` before...

Issue #1066, opened 2018-10-13
Caja caches folders from usb drives/disk


rcaridade145:
#### Expected behaviour
Two disks from Toshiba with the original name. Copy stuff from one to my pc. Safely remove it. Connect the other disk and i would get the folders from that disk.

#### Actual behaviour
Two disks from Toshiba with the original name. Copy stuff from one to my pc. Safely remove it. Connect the other disk and i get the folders from the first disk unless i close all the instances of Caja.

#### Linux Distribution
Ubuntu Mate 18.04.1

rcaridade145:
I have been looking into some commits on nautilus, that may fix this issue.

https://gitlab.gnome.org/GNOME/nautilus/commit/3891241ba760c59d284b7579dbd340651c8d4d29
https://gitlab.gnome.org/GNOME/nautilus/commit/3bca716182916da5832cdf61b54adcbf95e3a87f

Which were the fixes to these bugs:

https://bugzilla.gnome.org/show_bug.cgi?id=684226
https://bugzilla.gnome.org/show_bug.cgi?id=757555

lukefromdc:
I found only the first of the two Nautilus commits were directly applicable, as the "force-reload" function in Nautilus does not exist in Caja. None the less, I tested this with a pair of identically named camera cards formatted in the same camera, and no longer had the contents of the first card displayed after ejecting it, mounting the second card, and navigating to it again in Caja.

rcaridade145:
Really nice

lukefromdc:
Getting some issues with the PR, in that only a NEW window comes up with the correct contents. On the other hand, I just found that clicking the "reload" button in the toolbar will bring up the correct contents both in current master (and presumably thus releases) and with the PR when trying to show the new drive's contents in a window that was open when the old contents were showing

PR #1067, opened 2018-10-16
monitor: watch for removal of non-native mounts on GVolumeMonitor

lukefromdc:
Fix https://github.com/mate-desktop/caja/issues/1066 "Caja caches folders from usb drives/disk"

Nowadays, we rely on G_FILE_MONITOR_EVENT_UNMOUNTED to be emitted when a
mount disappears, since we already monitor the directory, in order to
switch the view to a different location.
Since non-native mounts very likely won't have file monitoring though,
we're not going to receive such events in that case, and fail to
redirect the view as a consequence.
This patch fixes it by listening to the mount-removed signal on
GVolumeMonitor when a monitor is requested for a non-native mount, and
emulating a file removal in that case.

Backport  Nautilus commit /gitlab.gnome.org/GNOME/nautilus/commit/3891241ba760c59d284b7579dbd340651c8d4d29

note that caja_get_mounted_mount_for_root had to be added, while the equivalent Nautilus function was from a prior commit

### DIFF
--- a/libcaja-private/caja-file-utilities.c
+++ b/libcaja-private/caja-file-utilities.c
@@ -756,6 +756,48 @@ caja_is_desktop_directory (GFile *dir)
     return g_file_equal (dir, desktop_dir);
 }
 
+GMount *
+caja_get_mounted_mount_for_root (GFile *location)
+{
+	GVolumeMonitor *volume_monitor;
+	GList *mounts;
+	GList *l;
+	GMount *mount;
+	GMount *result = NULL;
+	GFile *root = NULL;
+	GFile *default_location = NULL;
+
+	volume_monitor = g_volume_monitor_get ();
+	mounts = g_volume_monitor_get_mounts (volume_monitor);
+
+	for (l = mounts; l != NULL; l = l->next) {
+		mount = l->data;
+
+		if (g_mount_is_shadowed (mount)) {
+			continue;
+		}
+
+		root = g_mount_get_root (mount);
+		if (g_file_equal (location, root)) {
+			result = g_object_ref (mount);
+			break;
+		}
+
+		default_location = g_mount_get_default_location (mount);
+		if (!g_file_equal (default_location, root) &&
+		    g_file_equal (location, default_location)) {
+			result = g_object_ref (mount);
+			break;
+		}
+	}
+
+	g_clear_object (&root);
+	g_clear_object (&default_location);
+	g_list_free_full (mounts, g_object_unref);
+
+	return result;
+}
+
 /**
  * caja_get_pixmap_directory
  *
--- a/libcaja-private/caja-file-utilities.h
+++ b/libcaja-private/caja-file-utilities.h
@@ -47,6 +47,7 @@ gboolean caja_is_desktop_directory               (GFile *dir);
 gboolean caja_is_home_directory                  (GFile *dir);
 gboolean caja_is_home_directory_file             (GFile *dir,
         const char *filename);
+GMount * caja_get_mounted_mount_for_root         (GFile *location);
 gboolean caja_is_in_system_dir                   (GFile *location);
 char *   caja_get_pixmap_directory               (void);
 
--- a/libcaja-private/caja-monitor.c
+++ b/libcaja-private/caja-monitor.c
@@ -34,6 +34,9 @@
 struct CajaMonitor
 {
     GFileMonitor *monitor;
+    GVolumeMonitor *volume_monitor;
+    GMount *mount;
+    GFile *location;
 };
 
 gboolean
@@ -70,6 +73,27 @@ call_consume_changes_idle_cb (gpointer not_used)
     caja_file_changes_consume_changes (TRUE);
     call_consume_changes_idle_id = 0;
     return FALSE;
+}
+
+static void
+schedule_call_consume_changes (void)
+{
+    if (call_consume_changes_idle_id == 0) {
+	    call_consume_changes_idle_id =
+        g_idle_add (call_consume_changes_idle_cb, NULL);
+	}
+}
+
+static void
+mount_removed (GVolumeMonitor *volume_monitor,
+               GMount *mount,
+               gpointer user_data)
+{
+    CajaMonitor *monitor = user_data;
+    if (mount == monitor->mount) {
+        caja_file_changes_queue_file_removed (monitor->location);
+        schedule_call_consume_changes ();
+    }
 }
 
 static void
@@ -115,12 +139,7 @@ dir_changed (GFileMonitor* monitor,
 
     g_free (uri);
     g_free (to_uri);
-
-    if (call_consume_changes_idle_id == 0)
-    {
-        call_consume_changes_idle_id =
-            g_idle_add (call_consume_changes_idle_cb, NULL);
-    }
+    schedule_call_consume_changes ();
 }
 
 CajaMonitor *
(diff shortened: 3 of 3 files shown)


lukefromdc:
I tested this with a pair of identically named (and identical capacity  and brand though that may have less or no effect) camera cards formatted in  the same camera, and no longer had the contents of the first card  displayed after ejecting it, mounting the second card, and navigating to  it again in Caja.

To test, set up any pair of identically named USB devices and mount first one, then the other, viewing the contents in Caja from git master as of or before 10-15-2018 each time. Play with it as needed to invoke the bug where the contents of the first drive are still displayed when you have switched to the second drive (unmounting the first drive and then mounting the second) and navigated to it again in Caja. Now repeat with this PR, the contents shown should match the device on the second and subsequent mounts.

lukefromdc:
Nautilus's "check force reload" came in with a very complicated commit that was probably a bunch of stuff squashed together:
https://github.com/GNOME/nautilus/commit/5f777eba741727003fdc3f22153d2ca656abd96d#diff-2c08458c62793f59180267ce7a4b0f62
"slot: clean up" I was hoping to test it to see if it would help with the desktop failing to reload when something other than Caja adds a file to it, but it is too complex for tonight and would have to be applied along with 
gitlab.gnome.org/GNOME/nautilus/commit/3bca716182916da5832cdf61b54adcbf95e3a87f so as not to potentially cause a regression on the issue this PR fixes

raveit65:
I followed your steps-to-reproduce and i was able to reproduce the issue with 2 identical named usb-sticks and using caja from master.
But the issue still exists with PR for me.
I think the problem happens only when unmounting first usb-stick doesn't close the caja browser window, and the second usb-stick opens a new browser window.
I clicked on unmount-button in sidebar for unmount action.

lukefromdc:
My tests were always with a new window, as the old on on unmount either goes to "computer" or in many cases closes. OK, this means we've got this half-fixed and probably need the force-reload work for the rest.

lukefromdc:
Can you retest, always using a new Caja window for the second drive? You should still have the issue with master (as I did) and this should fix it so long as a new window is used. I had a LOT of trouble before this with not being able to get to the contents of a second identical drive without killing Caja at least. Sometimes I blamed the kernel and rebooted, not knowing Caja was the issue at all.

raveit65:
Now it becomes weird.
I tested it again without the PR more than half an hour and i wasn't able to reproduce the issue any more.
With the PR it happens sometimes but i couldn't find a clear reproducer.
I think the issue happens when an unmount signal doesn't close all caja windows and one caja window is still open.
I don't like such issues without a clear way to reproduce.
Any way. i don't see any other problems with the PR.

lukefromdc:
I have confirmation of your issue when the same window the first drive was displayed in was been kept open

lukefromdc:
OK, the exact behavioral change I've gotten with this PR is as follows:

With master as of now, all of Caja caches the contents of a mounted drive, so if another drive with an identical name is mounted, even opening a brand new window and navigating to that drive will show the content of the old drive.

With the PR, every window that was open when the old drive was mounted will show the old contents if navigated to the new drive, UNLESS the drive was first opened in a new window. Opening the drive in a new window now shows the contents of the new drive UNLESS an attempt was first made to open the drive in one of the existing windows.

I managed with considerable difficulty to apply relevant parts of 
https://github.com/GNOME/nautilus/commit/5f777eba741727003fdc3f22153d2ca656abd96d#diff-2c08458c62793f59180267ce7a4b0f62
followed by
https://gitlab.gnome.org/GNOME/nautilus/commit/3bca716182916da5832cdf61b54adcbf95e3a87f
which due to differences had to be applied to caja-window-manage-views.c rather than caja-window-slot.c

and got a buggy mess that segfaulted on moving up a directory tree but had no effect whatsoever on this behavior. Thus, to fix this right, we need to iterate through every open Caja window and run force-reload either at every mount or every unmount. That would be a nuisance when one or more windows is open to a very large directory, such as on Debian /var/lib/dpkg/info which has more than 10,000 files on my install and is quite slow to reload.

As the PR stands now, you have to close every window or restart Caja if you just attempted to show the contents of a new drive in a window that was open when an identically named other drive was mounted. With master as things stand now, you have to restart Caja unconditionally if you just replaced one drive with another, identically named drive.

We could document the new behavior in the help and manpage, then leave it so as not to force reloading other windows, we could set up running force-reload on all mounts or on all unmounts, or one more option might be this:  Whenever the "The folder contents could not be displayed" error dialog is called, first try running force-reload on that window once only, and show the dialog if that does not work. That is probably the best fix. Should we hold this PR for it, or merge this one and do another for an attempt at that?

lukefromdc:
There might be another, later Nautilus commit that does work for this: I just tested Nautilus and it did in fact correctly display the contents of both identically-named drives in the same window, one after the other.

lukefromdc:
One more thing: I just found that clicking the "reload" button in the toolbar will force the correct contents of the drive to appear, both in this PR and in Master.  What do we want to do?

raveit65:
I am fine with merging this PR as it is an improvement to old behavior.

raveit65:
Can we wait a week with merging? I like to run caja a while with this commit for testing.
@monsta @flexiondotorg @vkareh @sc0w 
Maybe your guys can test this too?

vkareh:
hmmm... one quirk that I noticed. Here's what I did, step by step:
1. Open caja in `computer:///`
2. Plug in USB drive. Icon shows up, but it looks unmounted.
3. Click on the USB drive icon. It gives an error that it's already mounted.
4. Refresh caja (F5). Icon now looks mounted and everything works.

Something similar happens on eject, but in reverse. Need to hit F5 for the icon to update after ejecting. This drive has an iso9660 filesystem (bootable media), so not entirely sure if this is considered native or not...

lukefromdc:
I've got no problem with waiting a week on this.

rcaridade145:
@vkareh Comparing caja code with nautilus monitor wise there are some events Nautilus uses that caja does not https://gitlab.gnome.org/GNOME/nautilus/blob/8acd51afd86b9c80daf7e3c284dee45ec300c965/src/nautilus-monitor.c

rcaridade145:
@vkareh Comparing caja code with nautilus monitor wise there are some events Nautilus uses that caja does not https://gitlab.gnome.org/GNOME/nautilus/blob/8acd51afd86b9c80daf7e3c284dee45ec300c965/src/nautilus-monitor.c

vkareh:
I haven't done a super exhaustive test using a bunch of different storage devices (apart from me regularly plugging my USB drive throughout the day for various reasons), but caja has been running smooth and stable for me all this time on my production machine. If anyone else wants to punish caja with some more mounts, feel free. Otherwise we can probably just merge this and address any new concerns (like more signal connections, if necessary) in a different pull request.

Thoughts?

raveit65:
Sadly, i found a regression with this commit.
Caja crashes every time i unmount a sftp mount to a VM with clicking on the eject button in sidebar under network.
I i found in xsession-errors log.
```
ERROR:caja-bookmark.c:533:caja_bookmark_connect_file: assertion failed: (!caja_file_is_gone (bookmark->details->file))
```
Steps to reproduce:
1. fire up a VM, or start with point 2 and connect to a real machine.
2. connect to VM via ssh in connect-to-sever-dialog (sftp).
3. Unmount mount point under network in sidebar with clicking on the eject button.
4. boom, caja crashes and all open caja windows were closed.

Without this commit unmounting a sftp mount works fine.
I don't think that the issue is limited to sftp (ssh) connections, but i can't test it with smb of ftp, because i don't use this.

lukefromdc:
I do not have any network shares whatsoever, so obviously this was not tested with one. I also do not have and have never used VM's and am not about to start digging into big downloads etc for that.  

MTP mounts don't crash on unmount with what I have now, and neither do SSHFS mounts, but that's all I can test.  I could look at the code for this tomorrow when I get some time, but I can't test the remote shares.

I am OK with reverting this, and maybe sticking a notation in the help that identically-named devices require clicking "reload." I simply cannot work on remote share stuff without someone else testing it.

lukefromdc:
One more point: If Nautilus ever had this problem they may have a commit that fixed it

raveit65:
You can connect to  your loopback network device (127.0.0.1) via ssh in connect to server-dialog for testing.
```
[root@mother rave]# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
```
Means you connect to your own account.
bildschirmfoto zu 2018-10-30 19-04-50
Of course you need to enable ssh server as root.
`systemctl enable sshd.service`

raveit65:
Searching for this error message for nautilus gave me some results, but no solution which fixes the issue for caja.

lukefromdc:
https://github.com/mate-desktop/caja/pull/1074
fixes the crash on unmounting at least the loopback device. Code is simple though NOT my favorite way to go about this. None the less it does work and does not throw errors in terminal

raveit65:
https://github.com/mate-desktop/caja/pull/1075 fixes the regression but pulls in another regression.
If you don't choose a dir , eg. /home, for the bookmark of the remote share, and you are using `/` dir, than the bookmark is broken.
Executing the bookmar again gives you an error message popop.
Something like ` dir isn't a folder, can`t open it`

I will try out https://github.com/mate-desktop/caja/pull/1074

raveit65:
https://github.com/mate-desktop/caja/pull/1074 does have the same problem.
I found out that a remote share bookmark with `/` dir works again after a restart of caja, but only one time.

PR #1070, opened 2018-10-21
add rubberband select to list view mode

lukefromdc:
Based on https://github.com/mate-desktop/caja/pull/1062 but with a build warning fixed and a possible memory leak freed

### DIFF
--- a/src/file-manager/fm-list-view.c
+++ b/src/file-manager/fm-list-view.c
@@ -86,6 +86,7 @@ struct FMListViewDetails
     GtkTreePath *new_selection_path;   /* Path of the new selection after removing a file */
 
     GtkTreePath *hover_path;
+    gboolean hover_hit_test;
 
     guint drag_button;
     int drag_x;
@@ -157,6 +158,11 @@ static void   fm_list_view_rename_callback                 (CajaFile      *file,
         GError            *error,
         gpointer           callback_data);
 
+
+static gboolean cell_event_hit_test (GtkTreeView *tree_view,
+                                     GtkTreePath *path,
+                                     GtkTreeViewColumn *column,
+                                     GdkEvent *event);
 
 G_DEFINE_TYPE_WITH_CODE (FMListView, fm_list_view, FM_TYPE_DIRECTORY_VIEW,
                          G_IMPLEMENT_INTERFACE (CAJA_TYPE_VIEW,
@@ -528,6 +534,10 @@ motion_notify_callback (GtkWidget *widget,
                         gpointer callback_data)
 {
     FMListView *view;
+    GtkTreeViewColumn *column;
+    GtkTreeModel *model;
+    GtkTreeIter hover_iter;
+    gboolean hit_test;
 
     view = FM_LIST_VIEW (callback_data);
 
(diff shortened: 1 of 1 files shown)


lukefromdc:
Note that there is one behavioral change, in that drag and drop is possible ONLY from the icon and not the entire tree row with this change

lukefromdc:
I've installed a "dev-all" build of Caja with both this and https://github.com/mate-desktop/caja/pull/1067 applied so I can watch this in terms of behavior

src/file-manager/fm-list-view.c
```diff
@@ -652,20 +683,137 @@ do_popup_menu (GtkWidget *widget, FMListView *view, GdkEventButton *event)
     }
 }
 
+
```
raveit65: Do we need those extra lines?

src/file-manager/fm-list-view.c
```diff
@@ -652,20 +683,137 @@ do_popup_menu (GtkWidget *widget, FMListView *view, GdkEventButton *event)
...
+    /* determine column/row area occupied by cell renderers */
+    for (l = cell_renderers; l != NULL; l = l->next) {
+        gtk_tree_view_get_cell_area (tree_view, path, column, &rect);
+        gtk_tree_view_column_cell_get_position (column, GTK_CELL_RENDERER (l->data),
+                                &x_offset, &width);
+        rect.x += x_offset;
+        rect.width = width;
+
+        gtk_cell_renderer_get_size (GTK_CELL_RENDERER (l->data),
```
raveit65: this gives me a deprecation warning.
```
fm-list-view.c: In function 'cell_event_hit_test':
fm-list-view.c:749:9: warning: 'gtk_cell_renderer_get_size' is deprecated: Use 'gtk_cell_renderer_get_preferred_size' instead [-Wdeprecated-declarations]
         gtk_cell_renderer_get_size (GTK_CELL_RENDERER (l->data),
         ^~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/gtk-3.0/gtk/gtktreeviewcolumn.h:25,
                 from /usr/include/gtk-3.0/gtk/gtktreeview.h:27,
                 from /usr/include/gtk-3.0/gtk/gtkcombobox.h:27,
                 from /usr/include/gtk-3.0/gtk/gtkappchooserbutton.h:29,
                 from /usr/include/gtk-3.0/gtk/gtk.h:42,
                 from fm-directory-view.h:31,
                 from fm-list-view.h:30,
                 from fm-list-view.c:29:
/usr/include/gtk-3.0/gtk/gtkcellrenderer.h:225:18: note: declared here
 void             gtk_cell_renderer_get_size       (GtkCellRenderer      *cell,
                  ^~~~~~~~~~~~~~~~~~~~~~~~~~
```

raveit65:
Hmm, open a file or folder in list-view with double-click doesn't work any more with that commit.
Can you confirm?

lukefromdc:
You have to double-click the icon itself (first column) to open the file. Between that and the deprecation warning (too complex for me to fix right now and not my code), we might want to close this. The behavior change on double-clicking files is likely to confuse people, and is apparently necessary for this to work, as the rubberbanding code had to exclude the first column,

raveit65:
Ok, that is working.
Personal i don't use the list-view, but is that the expected behaviour for using the list-view?
If you you don't know that you will think it is a bug, same like me.
So, let's close this.

raveit65:
Ok, that is working.
Personal i don't use the list-view, but is that the expected behaviour for using the list-view?
If you you don't know that you will think it is a bug, same like me.
So, let's close this.

raveit65:
Ok, that is working.
Personal i don't use the list-view, but is that the expected behaviour for using the list-view?
If you you don't know that you will think it is a bug, same like me.
So, let's close this.

PR #1071, opened 2018-10-25
Backport permission fixes

vkareh:
This cleans up commits from #1058 so that they preserve the log history

May fix the following issus (according to the original PR, yet to be tested)
* #1019 
* #954 
* #421

### DIFF
--- a/libcaja-private/caja-file.c
+++ b/libcaja-private/caja-file.c
@@ -555,7 +555,7 @@ modify_link_hash_table (CajaFile *file,
 	GList **list_ptr;
 
 	/* Check if there is a symlink name. If none, we are OK. */
-	if (file->details->symlink_name == NULL) {
+	if (file->details->symlink_name == NULL || !caja_file_is_symbolic_link (file)) {
 		return;
 	}
 
--- a/src/caja-connect-server-dialog.c
+++ b/src/caja-connect-server-dialog.c
@@ -1227,6 +1227,12 @@ caja_connect_server_dialog_fill_details_async (CajaConnectServerDialog *self,
 			g_mount_operation_set_password (G_MOUNT_OPERATION (operation),
 							str);
 			set_flags ^= G_ASK_PASSWORD_NEED_PASSWORD;
+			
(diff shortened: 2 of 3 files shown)


lukefromdc:
Needs testing with a remote share filesystem

rcaridade145:
Using the method @raveit65 described on #1075 i was able to have the behaviour below:

1 - Start sshd
2 - Connect to localhost and specify you want to save the password and create a bookmark
3 - Click the bookmark and you're asked for the password

raveit65:
about https://github.com/mate-desktop/caja/issues/421
I was never able to reproduce the issue reported with samba shares, because i don't have a server which use samba shares. I prefer nfs or sftp remote connections.
And with those protocols i can't reproduce it.
So what does https://github.com/mate-desktop/caja/pull/1071/commits/af22056fbc264f965aa9dc24bbc0d6b51ad4f4eb fix if nobody can reproduce it?
@rcaridade145 @monsta 
Maybe you can reproduce https://github.com/mate-desktop/caja/issues/421

Keep in mind that nautilus is complete different to caja after all that years with gnome3, and not everything which fixes an issue in nautilus helps for caja.
In another already merged PR we have 2 regressions from 1 fix from nautilus.
So, i would prefer that an issue should be clear reproduced by a team member before we add blindly a fix which can cause other regressions.

raveit65:
Yes, if you're able to test this with a samba server.
If i read bug description of https://github.com/mate-desktop/caja/issues/421#issue-85874293
```
"Remember password until you logout" is the default choice. I have to enter username and password twice : one to connect to the remote computer, a second to connect to a share
```
Entering the password 2 times, for login and the share, seems smb specific for me.

monsta:
Sigh, the only samba I can access at work has anonymous access. If the anonymous access is selected in the dialog, then the radio buttons for choosing how to remember the password are greyed out.

Though I always get the dialog when I try to connect to that server... yeah, the anonymous access option is selected by default, but I have to click the Connect button to actually connect. Well, I guess this doesn't help much with #421.

PR #1072, opened 2018-10-29
Fix case-insensitive in running_in_mate

zhangxianwei8:
The environment variable XDG_SESSION_DESKTOP and DESKTOP_SESSION in
running_in_mate are compared using g_strcmp0. Actually the environment
variable in MATE DE is "mate"(lower case). Using g_ascii_strcasecmp
instead of g_strcmp0 to fix it.

The XDG_SESSION_DESKTOP and DESKTOP_SESSION are set here:
https://github.com/CanonicalLtd/lightdm/blob/7241fd81c0dfcbe354f2c9bc4ace04088de5d860/src/seat.c#L979

export:
[root@localhost]# export |grep XDG_SESSION_DESKTOP
declare -x XDG_SESSION_DESKTOP="mate"

[root@localhost]# export |grep DESKTOP_SESSION
declare -x DESKTOP_SESSION="mate"

### DIFF
--- a/libegg/eggsmclient.c
+++ b/libegg/eggsmclient.c
@@ -57,8 +57,8 @@ static gboolean
 running_in_mate (void)
 {
     return (g_strcmp0 (g_getenv ("XDG_CURRENT_DESKTOP"), "MATE") == 0)
-        || (g_strcmp0 (g_getenv ("XDG_SESSION_DESKTOP"), "MATE") == 0)
-        || (g_strcmp0 (g_getenv ("DESKTOP_SESSION"), "MATE") == 0);
+        || (g_strcmp0 (g_getenv ("XDG_SESSION_DESKTOP"), "mate") == 0)
+        || (g_strcmp0 (g_getenv ("DESKTOP_SESSION"), "mate") == 0);
 }
 
 static void
--- a/src/caja-application.c
+++ b/src/caja-application.c
@@ -2041,8 +2041,8 @@ static gboolean
 running_in_mate (void)
 {
     return (g_strcmp0 (g_getenv ("XDG_CURRENT_DESKTOP"), "MATE") == 0)
-        || (g_strcmp0 (g_getenv ("XDG_SESSION_DESKTOP"), "MATE") == 0)
-        || (g_strcmp0 (g_getenv ("DESKTOP_SESSION"), "MATE") == 0);
+        || (g_strcmp0 (g_getenv ("XDG_SESSION_DESKTOP"), "mate") == 0)
+        || (g_strcmp0 (g_getenv ("DESKTOP_SESSION"), "mate") == 0);
 }
 
 static gboolean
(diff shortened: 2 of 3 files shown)


zhangxianwei8:
Could you unset XDG_CURRENT_DESKTOP and set XDG_SESSION_DESKTOP and DESKTOP_SESSION to “mate” to test？

lukefromdc:
I am working on my main machine (no VM's available) and really do not know how to mess with that. I would not want to get into a situation I did not know how to get back out of. Anyway, I had no problems on new sessions.

zhangxianwei8:
This is not a bug, just optimizing code:)

raveit65:
```
[rave@mother ~]$ env | grep XDG_SESSION_DESKTOP
XDG_SESSION_DESKTOP=mate
[rave@mother ~]$ env | grep DESKTOP_SESSION
DESKTOP_SESSION=mate
```
Why should we change that ?
Btw. Where is it written that it is wrong?

raveit65:
@monsta @vkareh @XRevan86 
thoughts?

vkareh:
@zhangxianwei8 - why not change to lower case instead:
```diff
-        || (g_strcmp0 (g_getenv ("XDG_SESSION_DESKTOP"), "MATE") == 0)
-        || (g_strcmp0 (g_getenv ("DESKTOP_SESSION"), "MATE") == 0);
+        || (g_strcmp0 (g_getenv ("XDG_SESSION_DESKTOP"), "mate" == 0)
+        || (g_strcmp0 (g_getenv ("DESKTOP_SESSION"), "mate" == 0);
```

raveit65:
> I want to know if there is a specification for reference?

At freedesktop?

raveit65:
I hope you will find detail infos at freedesktop site.
If not than their isn't a specification about.

zhangxianwei8:
I find that XDG_SESSION_DESKTOP is defined here:
https://www.freedesktop.org/software/systemd/man/pam_systemd.html
A single, short identifier string for the desktop environment. This may be used to indicate the session desktop used, where this applies and if this information is available. For example: "GNOME", or "KDE". It is recommended to use the same identifiers and capitalization as for $XDG_CURRENT_DESKTOP, as defined by the Desktop Entry Specification.

XDG_CURRENT_DESKTOP is defined here:
https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s06.html
$XDG_CURRENT_DESKTOP should have been set by the login manager, according to the value of the _DesktopNames_ found in the session file. The entry in the session file has multiple values separated in the usual way: with a semicolon.

zhangxianwei8:
If nobody has different suggest, I will change the "MATE" to "mate" and keep using g_strcmp0. @raveit65 @vkareh

zhangxianwei8:
Updated.

raveit65:
@vkareh @monsta @lukefromdc 
I think we can consider this as bugfix and cherry-pick to 1.20 branch ?

lukefromdc:
Fine by me, the final merged code is working fine here on master

raveit65:
done.
2676eaf3ed92ef33e2a39aece779465c427a59db

PR #1074, opened 2018-10-31
Fix crash on umounting remote network shares

lukefromdc:
fix regression from https://github.com/mate-desktop/caja/commit/8fff655552209fd181f9608d02d0f72660b10768 that caused unmounting network shares

### DIFF
--- a/libcaja-private/caja-bookmark.c
+++ b/libcaja-private/caja-bookmark.c
@@ -530,8 +530,11 @@ caja_bookmark_connect_file (CajaBookmark *bookmark)
     if (!caja_bookmark_uri_known_not_to_exist (bookmark))
     {
         bookmark->details->file = caja_file_get (bookmark->details->location);
-        g_assert (!caja_file_is_gone (bookmark->details->file));
+        /*Catch the case of removing a remote mount triggering this code block*/
+        if(caja_file_is_gone (bookmark->details->file))
+            return;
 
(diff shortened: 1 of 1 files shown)


lukefromdc:
Not my favorite way to do this, but it does keep Caja running and does not throw errors while running from the terminal. . Unmounting network shares after backporting Nautilus commit /gitlab.gnome.org/GNOME/nautilus/commit/3891241ba760c59d284b7579dbd340651c8d4d29 was causing an assert in bookmark code to trigger, even though the mount is shown in "Network" on the sidebar. In some way this must be a bookmark or that code would not trigger, so catch the case of the file being gone by returning.  At that point the assert can no longer be reached unless the condition contained in it is TRUE anyway, so comment it out.

If anyone knows a better way to do this I am open to it. Nautilus can no longer mount network shares but can unmount them from the sidebar without crashing, even though both the offending commit (not helped by a later commit in the case of Caja) and the assert with nothing to catch this issue are both present.

rcaridade145:
Looking at the nautilus commit log it seems that this commit https://gitlab.gnome.org/GNOME/nautilus/commit/9429c94f354460e7bcc2a9063b98f3a904e6a358 may fix the issue.

raveit65:
@rcaridade145 
Thanks for finding that. nautilus commit fixes the issue for me too.
I will open an alternative PR.

PR #1061, opened 2018-09-28
Backport verify_destination from Nautilus - Symlink and freespace fixes

rcaridade145:
Should help with the symlink issues like #315 , #947 , #748 . This version also verifies the amount of freespace available before copy.

PR #1075, opened 2018-10-31
Alternative fix for crash on umounting remote network shares

raveit65:
This is an alternative for https://github.com/mate-desktop/caja/pull/1074
First commit fixes the regression causes by 8fff655 .
Second commit fixes a similar crash in nautilus if mounting the unmounted drive again, with same error message.
```
ERROR:caja-bookmark.c:533:caja_bookmark_connect_file: assertion failed: (!caja_file_is_gone (bookmark->details->file))
````
See here for details.
https://bugs.launchpad.net/nautilus/+bug/1202159
https://bugzilla.gnome.org/show_bug.cgi?id=708282#c1
https://bugzilla.redhat.com/show_bug.cgi?id=1254161
Not sure if we really need the second one and we can remove them from PR for safety if you don't agree.

Please test.

### DIFF
--- a/libcaja-private/caja-bookmark.c
+++ b/libcaja-private/caja-bookmark.c
@@ -530,7 +530,14 @@ caja_bookmark_connect_file (CajaBookmark *bookmark)
     if (!caja_bookmark_uri_known_not_to_exist (bookmark))
     {
         bookmark->details->file = caja_file_get (bookmark->details->location);
-        g_assert (!caja_file_is_gone (bookmark->details->file));
+        /* Similar to what we do when a file has been trashed or deleted
+         * withouth noticing in bookmark_changed, just try to fall graciously.
+         */
+        if (caja_file_is_gone (bookmark->details->file))
+          {
+                caja_bookmark_disconnect_file (bookmark);
+                return;
+          }
 
         g_signal_connect_object (bookmark->details->file, "changed",
                                  G_CALLBACK (bookmark_file_changed_callback), bookmark, 0);
--- a/src/caja-window-menus.c
+++ b/src/caja-window-menus.c
@@ -1092,6 +1092,7 @@ add_extension_menu_items (CajaWindow *window,
             caja_menu_item_list_free (children);
             g_free (subdir);
         }
+        g_object_unref (action);
     }
 }
 

### CODE REVIEW

**Comment by raveit65** [comment]
This pulls in another regression.
If you don't choose a dir , eg. `/home`, for the bookmark of the remote share, and you are using `/` dir, than the bookmark is broken.
Executing the bookmar again gives you an error message popop.
Something like `dir isn't a folder, cant open it`

**Comment by rcaridade145** [comment]
@raveit65 Sorry but it seems this commit was reverted on a commit on 3.14 development https://gitlab.gnome.org/GNOME/nautilus/commit/07719c951f9d4d077a990590e845b22a521f5ab8 .

From your description can you be hitting this also https://gitlab.gnome.org/GNOME/nautilus/commit/6d0a7e059207000307acaf96ec2c44196266f115 ?

**Comment by raveit65** [comment]
```
Revert "nautilus-bookmark: don't crash if file is gone"

This reverts commit 9429c94f.
This was a blind fix, and it's not correct. Correct one is only
need on 3.14 with commit a0cbf728
```
The as correct mentioned commit is already use in this PR and doesn't fix the problem
`window-menus: unref extension created action`

Any chance that you try out your suggestions in your local installation before asking other people to do that?

rcaridade145:
@raveit65 Sure. I just mentioned it because the fix it seems  Nautilus is just ignoring c326374 and maintaining 7c1f3fe. 

> From your description can you be hitting this also https://gitlab.gnome.org/GNOME/nautilus/commit/6d0a7e059207000307acaf96ec2c44196266f115 ?

Just put up this question to understand if this was the bug you were hitting. Didn't understand very well the dir issue.So my question.

My apologies.

raveit65:
I mean this problem.
screenshot at 2018-10-31 19-17-51
The bookmark `f29` is a remote share which points to the root dir of the remote server.
`sftp://rave@192.168.101.133/` vs `sftp://rave@192.168.101.133/home`
Last one, which is working, is the /home dir of the remote server.
Take a look at connect-to-server-dialog
bildschirmfoto zu 2018-10-30 19-04-50
Here you can select a dir of the remote connection.

But for me it doesn't make any sense to choose a dir in most cases.
If i work on a remote server i need always the root dir.

And you don't need  a server fo testing this , simply connect via ssh protocol to your own box with 127.0.0.1=localhost , the address of your loopback network device.

rcaridade145:
I have not been able to reproduce this.

raveit65:
The problem exists if you mount the share again after unmounting the share.

lukefromdc:
https://github.com/mate-desktop/caja/pull/1074 behaved as bad as or worse than this one when trying to open bookmarks made on mounting a remote filesystem, often with even the FIRST attempt failing.

I also found that if a filepath beyond / was specified when the filesystem was mounted, the bookmark stayed valid and clicking it mounted the filesystem in question. There is a GNOME commit that relates to subdirectories that might help with this.

https://github.com/GNOME/nautilus/commit/a23ed4ad0173ca7a51e26282b85d06ce4b25912a#diff-e99b6a5a724144533ffe864ffec81782 
"monitor: improve unmount detect heuristic"

I have the ssh key staying unlocked for the session duration by default, and under normal conditions always NOT being used and the server turned off.

lukefromdc:
https://github.com/GNOME/nautilus/commit/a23ed4ad0173ca7a51e26282b85d06ce4b25912a was of no effect

lukefromdc:
I've also confirmed that reverting the first commit (as Nautilus did) brings back the crash. OK, what to do about the bookmark?  It is initially valid with this, but once unmounted cannot be used again, yet if Caja is restarted the file is reconnected in such a way that clicking the bookmark mounts the share.

It looks like the problem is that Caja can only accept mounting the share ONCE at this point and still properly connect to the bookmark. On the first connection, the sidebar selected row is set to the newly created bookmark. If this bookmark is removed and the process repeated, the bookmark comes up with a plain file icon, the selected row does NOT navigate to it in sidebar-and the Caja window does not change to display the contents of the newly mounted volume until the mount in "network" or the file on the desktop (if mounts displayed there) is clicked.

Possibly something in 
https://github.com/mate-desktop/caja/commit/8fff655552209fd181f9608d02d0f72660b10768
is not all the way removing the previous reference to a remote filesystem? If we have to we can still revert that whole mess and add a notation in the help to reload when changing between any set of identically named external volumes

lukefromdc:
I found a much simpler fix, by just commenting out the part of the original GNOME commit that treated remote filesystems differently.
https://github.com/mate-desktop/caja/pull/1076
Stops the crash, keeps the assert, and does not cause any problems with bookmarks in my testing with the loopback fs or my laptop

rcaridade145:
So this fixes also what @raveit65 was getting about "location not being a folder"? If not on #1061 i backported the latest "verify_destination" from Nautilus to include symlink fixes but also changes the trigger of that error.

raveit65:
Closing in favour of https://github.com/mate-desktop/caja/pull/1076

PR #1076, opened 2018-11-02
Simpler fix for crash on umounting remote network shares, keeps any bookmarks from them valid

lukefromdc:
Don't treat remote filesystems differently when unmounting, so that bookmarks can remount them if they are available and so caja does not crash.

An alternative to 
https://github.com/mate-desktop/caja/pull/1075
and to
https://github.com/mate-desktop/caja/pull/1074

### DIFF
--- a/libcaja-private/caja-monitor.c
+++ b/libcaja-private/caja-monitor.c
@@ -154,13 +154,13 @@ caja_monitor_directory (GFile *location)
     if (dir_monitor != NULL) {
         ret->monitor = dir_monitor;
     }
-
+    /*This caused a crash on umounting remote shares
     else if (!g_file_is_native (location)) {
         ret->mount = caja_get_mounted_mount_for_root (location);
         ret->location = g_object_ref (location);
         ret->volume_monitor = g_volume_monitor_get ();
     }
-
+    */
     if (ret->monitor != NULL) {
         g_signal_connect (ret->monitor, "changed",
                   G_CALLBACK (dir_changed), ret);


lukefromdc:
No crashes resulted from repeatedly mounting and unmouting the same loopback "share," so the second commit of https://github.com/mate-desktop/caja/pull/1075 should not be necessary here. If someone hits the crash it was meant to fix we could add it.

With this version, bookmarks created by a remote filesystem that can still be mounted always work, mounting the filesystem when clicked on. In all cases where the fs could be reached, the window from which the connect to server dialog was invoked loaded the contents of the directory connected to, whether the root of the filesystem or a subdirectory. Bookmarks also worked the same either way, and if a bookmark was created on mounting the share, the sidebar selected row was set to the bookmark.

I also tested this with an SSH connection into my laptop, which worked fine. I then tested to see what happened when the server could not be reached, by disabling networking on the laptop. Caja kept running, the curser spun until the timout was reached, invoking this dialog
```

could not display sftp"//(IP address). 
Error:timed out when logging in 
Please Select another viewer and try again
```

and on a second try a shorter timeout gave
 "because the host could not be found" as the second line in the dialog
I assume that is correct for an unreachable server? On re-enabling networking the router gave the laptop a different IP address, so the bookmark could not be tested on reestablising contact.

I assume this is the desired behavior? If so, it looks like the whole mess came from the part of the orignal GNOME commit that treated a remote FS differently, so I just commented that part out for now.

raveit65:
Yes, this is the expected behaviour if you can't reach another box because the box isn't available in the network or a firewall blocks the connection.

In my router there is an option which gives a notebook or any outher box always the same address from dhcp server, btw. this is an option in general dhcp server config.

Issue #1046, opened 2018-08-15
segfault 4  from usb drive unmount


TuxVinyards:
**Unmounting a usb drive causes whole of Caja to crash**.  
(Caja 1.20.2  with Ubuntu Mate 18.04)

Using  `cat /var/log/kern.log | grep caja`  yielded  constant  `segfault 4 ` references

Following this link:  [https://stackoverflow.com/a/26638899](url) I traced the fault with `addr2line` to this:
"caja_path_bar_get_button_from_button_list_entry"

My main HDD is set as 3 ext4 partitions which all mount on boot and should remain open in Caja, usually in tabs, when unmounting the usb.  Is this Ubuntu or Caja?  Or something else?

Hope this trace helps.  Bests.

raveit65:
template ignored....

lukefromdc:
That function is called only from `real_set_active` in caja-navigation-window-pane.c and this looks like another race condition. My guess is that, something is calling "real_set_active on the same window with a new directory showing before the buttons actually exist. I sometimes get the window showing the drive's directory closing rather than switching back to the "computer" directory but never a segfault. 

Wonder if this is something distro-specific? Thus the template and the ask for a downstream bug report if this is  a package provided by the distro.

TuxVinyards:
@raveit65 will format a templated report for you. I felt the report made more sense without but presume you would like this for auto-parsing ....

@lukefromdc  thanks for comments.  In the meantime, I have done a bit more research:

1. This doesn't happen with PcManFM. The default Lubuntu file manager will install quite happily onto Ubuntu Mate and both USB HDD's and sticks will mount and unmount without problems.

2. I have noticed that if using a USB stick, that Caja remains up if selecting 'eject'. It only fails if selecting 'safely remove drive'. The latter is of course the only option with external HDD.

3. I would caution anyone trying to replicate this error not to have important data on the drive as it does seem to cause damage to the external drive file system. (Gnome Disks does repair this though).

4. Synaptic lists pcmanfm as using GVFS for removable media, which I think is the same as Caja.

5. Of note, I have had this problem for quite a while, certainly since U.Mate 18.04 but not sure about before that. I originally thought it was the drive, not the software ....  I have also carried out re-installs of Caja, gvfs and other components too.

I now think this problem centres on the 'safely remove drive' button function of Caja and probably an upstream fix will be needed.

Hope someone familiar with Caja's inner workings can fix this.

TuxVinyards:
[...]

PR #1084, opened 2018-12-01
Fix segfault on stopping USB hard drives

lukefromdc:
Fix https://github.com/mate-desktop/caja/issues/1046
drive_stop_cb should never have used g_drive_poll_for_media_finish which is for
ejectable media in a fixed drive (e.g CD drive) and not for stopping a removable drive

See
https://developer.gnome.org/gio/stable/GDrive.html
https://developer.gnome.org/gio/stable/GDrive.html#g-drive-stop-finish

### DIFF
--- a/src/caja-places-sidebar.c
+++ b/src/caja-places-sidebar.c
@@ -2548,7 +2548,7 @@ drive_stop_cb (GObject *source_object,
     g_object_unref (window);
 
     error = NULL;
-    if (!g_drive_poll_for_media_finish (G_DRIVE (source_object), res, &error))
+    if (!g_drive_stop_finish(G_DRIVE (source_object), res, &error))
     {
         if (error->code != G_IO_ERROR_FAILED_HANDLED)
         {
--- a/src/caja-places-sidebar.c
+++ b/src/caja-places-sidebar.c
@@ -1899,7 +1899,7 @@ drive_start_from_bookmark_cb (GObject      *source_object,
     char *name;
 
     error = NULL;
-    if (!g_drive_poll_for_media_finish (G_DRIVE (source_object), res, &error))
+    if (!g_drive_start_finish (G_DRIVE (source_object), res, &error))
     {
         if (error->code != G_IO_ERROR_FAILED_HANDLED)
         {
@@ -2488,7 +2488,7 @@ drive_start_cb (GObject      *source_object,
     char *name;
 
     error = NULL;
-    if (!g_drive_poll_for_media_finish (G_DRIVE (source_object), res, &error))
+    if (!g_drive_start_finish (G_DRIVE (source_object), res, &error))
     {
         if (error->code != G_IO_ERROR_FAILED_HANDLED)
         {


lukefromdc:
OK, I have not been able to find any way to start a drive from the sidebar or a bookmark, as both drives and bookmarks for drives are removed from the sidebar when the drive is stopped. I updated the code for both these blocks using just the change of monitoring function and not moving the error dialog code to a separate callback as the Nemo devs did(more complex), but I'm thinking those two blocks may now be dead code unless we can find a way to start a stopped drive from the sidebar..

If no such method now exists, we have three options: drop the second commit to this as untested/untestable, keep the second commit as needed if starting drives from the sidebar is ever re-enabled (by showing bookmarks of stopped drives or stopped drives again), or simply drop both start from sidebar code blocks and the code calling them as unused code.

raveit65:
I can't reproduce the issue , so i am not able to test this.

raveit65:
I tested it again with f9 + glib2-2.58.1-1.fc29.x86_64 + gtk3-3.24.1-1.fc29.x86_64,
but i don't get any crash of caja if unmounting an usb-stick. My steps to reproduce:
1. plug in a stick (usb-2.0/3.0, doesn't matter)
2. unmount the stick from sidebar with eject icon or context menu
3. i see a notification (save to unplug stick)
4. unplug the stick from system.

Do you do something different?
Are you sure this is a caja problem?

raveit65:
> OK, I have not been able to find any way to start a drive from the sidebar or a bookmark, as both drives and bookmarks for drives are removed from the sidebar when the drive is stopped.

I don't think that is needed for  a removable drive like a usb-stick.
You can unplug/plugin the stick to see them again in sidebar.

lukefromdc:
USB sticks don't need to be powered up on insertion by these functions either, as they start automatically. The only use case I could see for them was if a stopped spinning drive was still shown in the sidebar, but as long as I can recall stopped drives were removed from sidebar, and on reinsertion started automatically without Caja's involvement. Setting a bookmark doesn't work, as the bookmark gets hidden when the drive is stopped for removal

raveit65:
With `spinning drive` you mean a partition of a hard-disc?
I have some partitions mounted via /etc/fstab with the `noauto` option.
Means they are in sidebar but not mounted.

lukefromdc:
I mean whole hard drives that can be powered off, not partitions of drives. Even an entire drive that is connected to the machine at boot or later with everything as "noauto" would still be powered on. If the drive is in "standby" and spun down it is still powered on at the board level and will spin up when accessed.

In the past, a bookmark for a broken symlink (or maybe also a bookmarked stopped disk) would remain in sidebar with a little red dot. Fixes backported from Nautilus for other issues ( Was it https://github.com/mate-desktop/caja/commit/8fff655552209fd181f9608d02d0f72660b10768  for https://github.com/mate-desktop/caja/issues/1066 maybe?) removed bookmarks for no longer present devices showing, and such bookmarks might have been the one case for restarting an entire hard drive that had been bookmarked and then stopped

One time the "stop" option does not work (the case I used for testing the error dialog) is a USB hard drive with multiple partitions, one mounted root from the command line and another mounted from caja, then unmounted, then trying to stop the drive. The error dialog then pops up as intended saying the drive cannot be stopped because that partition is in use (or in my case is an unlocked LUKS volume that can't be locked).

raveit65:
You can change power state of a drive with `hdparm`
```
[root@mother rave]# hdparm -C /dev/sda
/dev/sda:
 drive state is:  active/idle
[root@mother rave]# hdparm -y /dev/sda
/dev/sda:
 issuing standby command
[root@mother rave]# hdparm -C /dev/sda
/dev/sda:
 drive state is:  standby
[root@mother rave]# hdparm -Y /dev/sda
/dev/sda:
 issuing sleep command
```
But my drives are always visible in sidebar because i mount them via fstab.
And changing power state doesn't have any influence.

lukefromdc:
I can also power down a disk from gnome-disks, but it still disappears from the sidebar. The second commit here affects ONLY restarting a disk that has been fully powered down for removal, not a disk in sleep or spun down but still powered on mode. BTW, powering off also causes the drive to disappear from gnome-disks.

The HDparm commands I could find for power management can put a disk into its lowest "sleep" state so a reset command is needed to wake it, but cannot power it off in the same way the "stop" command from the right-click menu from the sidebar or a "Power Off" command from gnome-disks does. In the former case, the drive stays in sidebar and even stays mounted, just being slow to access as it wakes up. In the latter case, the drive is treated as though it had been already unplugged both in Caja and in gnome-disks. Will now test Nautilus

lukefromdc:
In Nautilus the drive is also completely removed from the sidebar when it is powered off from "safely remove" or gnome-disks. A bookmark set in Nautilus will continue to appear, but attempting to restart the drive by clicking on that bookmark brings up a "something went wrong" error dialog.  Thus, these changes should be usable as-is, and if we ever do get bookmarks to restart a fully powered off drive working again we won't get segfaults from using the wrong monitoring function in the underlying code here,

lukefromdc:
I just confirmed that bookmarks now show up only for MOUNTED volumes. "Stop" also unmounts as well as powering down, and same for doing this from gnome-disks etc

lukefromdc:
I just further confirmed by looking at the /dev directory that at least with current kernels, stopping a hard drive ("stop" or "safely remove") actually removes the /dev/sd(something) handle for it, so it is all the way gone at that point.

raveit65:
I guess we can consider this as bugfix and cherry-pick to 1.20 branch?

lukefromdc:
I was thinkng the memory leak fixes should also be cherrypicked to 1.20 and then these two commits, and in the same order?

raveit65:
Yes, please go ahead.

lukefromdc:
The two memory leak fix commits below could not be cherrypicked due to conflicts:
caja-bookmark:fix memory leaks
eel-gtk-extensions:fix memory leaks
I cherrypicked the other two memory leak fixes and then these two commits

Issue #1099, opened 2018-12-24
weird item in bookmarks if the folder was removed


sc0w:
#### Expected behaviour

no weird item if it was removed

#### Actual behaviour

weird item

#### Steps to reproduce the behaviour

1. make new folder
2. go to the new folder
3. go to previous position
4. remove the new folder
4. click here and see:

![2018-12-22_13-24](https://user-images.githubusercontent.com/7734191/50396917-24b23800-076d-11e9-8ad9-7381e34173a7.png)

#### MATE general version

1.20.3

#### Package version

1.20.2 and git master

#### Linux Distribution

debian buster/testing

PR #1100, opened 2018-12-24
Fix: weird item in bookmarks if the folder was removed

sc0w:
Fixes https://github.com/mate-desktop/caja/issues/1099


sc0w:
@vkareh I can't see that, I just see a little point istead:

sc0w:
@vkareh ok, confirmed in green submarine theme, in traditional ok I see only the little point I mentioned

sc0w:
I don't know how to avoid the void rectangle, so, I just put an error message instead with latest changes, and now it is gray out



@vkareh @lukefromdc please test

lukefromdc:
This works fine, and gives the "folder-removed" error message and icon as the menuitem and it is insensitive so as not to take the user to an error dialog.

Issue #1101, opened 2018-12-30
regression: weird look in icon view


sc0w:
#### Expected behaviour

Normal look

#### Actual behaviour

![2018-12-30_01-52](https://user-images.githubusercontent.com/7734191/50543329-aecb2800-0bd5-11e9-94d2-058b04efbacc.png)

#### Steps to reproduce the behaviour

_desktop folder_: 258 items in _compact view_

_home (scow) folder_: 513 items in _icon view_

I go using the pane to _desktop_ - _home_ - _desktop_ - _home_ several and quikly times, and I see in home the look I showed before

culprit commit: https://github.com/mate-desktop/caja/commit/021ff2ac062f1cfd52726c78e9a3956405001c6e

#### MATE general version

1.20.3

#### Package version

1.21.3 and git master

#### Linux Distribution

debian buster/testing

#### Link to downstream report of your Distribution

version not released yet

lukefromdc:
I was unable to duplicate this, but don't have more than 100 items (directories plus normal files) in either directory. Using two other directories I still could not duplicate this.. I could not find a similar report against Nemo (maybe nobody has it it yet?)  Still, I won't oppose reverting a commit that causes a problem, we can revisit this if Nemo fixes it right

sc0w:
little video with the issue:

[Untitled Project.ogg.zip](https://github.com/mate-desktop/caja/files/2716917/Untitled.Project.ogg.zip)

lukefromdc:
Oh yeah-that's an ugly mess. Can't duplicate it here but if anyone has this the offending commit has to go. I'm thinking some kind of race condition and we are getting an attempt to display the directory before it is ready

raveit65:
In the video all folder icons are blurred. Is this the issue?

sc0w:
@raveit65 

no, I blurred it intentionally (to hide names of personal folders)

the final result is the screenshot in my first message

sc0w:
[compact or list view], to [icon view], weird look in [icon view]

from [icon view] to [icon view], no problems

raveit65:
Yes, i can confirm the problem, but only if child folder use compact view and the sidebar is used.
And i see only one item in broken icon-view folder (about 3700 items).
For some reasons the problem looks different here.
See my video https://www.dropbox.com/s/6jmmow6klbl9kkm/caja-icon-view.ogv.tar.xz?dl=0
I tested the same procedure with nemo, but i can't reproduce the issue with it.

[...]

PR #1102, opened 2018-12-30
Fix weird look in icon view

sc0w:
This reverts commit 021ff2ac062f1cfd52726c78e9a3956405001c6e.

Fixes https://github.com/mate-desktop/caja/issues/1101


sc0w:
6c423bc `Caja-icon-info: Optimize icon lookups by GIcon.`

in my tests, that commit isn't related to the issue

lukefromdc:
That's what I was hoping: being able to keep that any difference in responsiveness will be less obvious and the larger the directory the greater the importance of it. Glad that one is OK.

PR #1103, opened 2018-12-30
[libcaja-private] Re-check file MIME type before picking an application.

Max-E:
An example of when the MIME type might change: a file is initially created
with 0 bytes of content, but more data is added later. Empty files are always
detected as plain text, but the file might not be empty anymore when the user
opens it.

I encounter this daily when I start downloading a youtube video using youtube-dl, and I want to double-click on the .part file to start watching it before it's done downloading.

I developed this against MATE 1.20 since that's what I'm running on my computer. Didn't feel comfortable submitting a PR against a version of MATE that I didn't personally test.

### DIFF
--- a/libcaja-private/caja-file.c
+++ b/libcaja-private/caja-file.c
@@ -2555,6 +2555,24 @@ caja_file_update_info (CajaFile *file,
 	return update_info_internal (file, info, FALSE);
 }
 
+void
+caja_file_refresh_info (CajaFile *file)
+{
+	GFile *gfile;
+	GFileInfo *new_info;
+
+	gfile = caja_file_get_location (file);
+	new_info = g_file_query_info (gfile, CAJA_FILE_DEFAULT_ATTRIBUTES,
+	                              G_FILE_QUERY_INFO_NONE, NULL, NULL);
+	if (new_info != NULL) {
+		if (caja_file_update_info (file, new_info)) {
+			caja_file_changed (file);
+		}
+		g_object_unref (new_info);
+	}
+	g_object_unref (gfile);
+}
+
 static gboolean
 update_name_internal (CajaFile *file,
 		      const char *name,
--- a/src/file-manager/fm-directory-view.c
+++ b/src/file-manager/fm-directory-view.c
@@ -8813,6 +8813,13 @@ real_update_menus (FMDirectoryView *view)
 
 		file = CAJA_FILE (l->data);
 
+		/* Double-check if the files' MIME types have changed before we
+		   commit to a choice of applications for them. This can happen
+		   if, for instance, a file was originally created with 0 bytes
+		   and then content was added to it later-- it will change from
+		   plaintext to something else. */
+		caja_file_refresh_info (file);
+
 		if (!caja_mime_file_opens_in_external_app (file)) {
 			show_app = FALSE;
 		}
--- a/libcaja-private/caja-file.h
+++ b/libcaja-private/caja-file.h
@@ -144,6 +144,12 @@ void                    caja_file_monitor_add                       (CajaFile   
 void                    caja_file_monitor_remove                    (CajaFile                   *file,
         gconstpointer                   client);
 
+/* Synchronously refreshes file info from disk.
+ * This can call caja_file_changed, so don't call this function from any
+ * of the callbacks for that event.
+ */
+void                     caja_file_refresh_info                     (CajaFile                   *file);
+
 /* Waiting for data that's read asynchronously.
  * This interface currently works only for metadata, but could be expanded
  * to other attributes as well.
--- a/libcaja-private/caja-mime-actions.c
+++ b/libcaja-private/caja-mime-actions.c
@@ -997,6 +997,12 @@ make_activation_parameters (GList *uris,
         uri = l->data;
         file = caja_file_get_by_uri (uri);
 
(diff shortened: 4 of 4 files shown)


lukefromdc:
Build completes so this should just be a matter of rearranging things or adding a declaration at the top of the file. We need an easy way to test this: I have seen the behavior you describe with video downloads, but that is bandwidth-intensive and I don't have a landline

lukefromdc:
OK, the test program not only works but works well enough I will keep it around for other uses if I can avoid losing it.  On testing Caja from master the issue is reliably invoked with a video file that has the headers in front for playablity of a partial download. As the copy job ran, the right-click menu first still specified Pluma to open the file, but simply clicking on it opened it in MPV, my default video player. Afterwards MPV also showed up as the default application to open it. The icon remained a "plain text" icon until the copy job was done, then it changed to a video thumbnail.

OK, we've got the build warning, and if possible it would be nice to get the icon/thumbnail generation and right-click menu to update as well, though just fixing the build warning would be enough

Max-E:
I can take a stab at it, but only if I can do it without delaying the application launch until after the thumbnail is generated.

I noticed there's a test/ directory in the caja tree, should I just stick this program there?

lukefromdc:
We don't need to add the slowcopy program to Caja's test source code as it would add nearly a MB to every download of it. I agree that the thumbnail generation must not delay application launch, with video there is usually a "pending" thumbnail which can be a transparent dotted line, a clock, or something else depending on the icon theme and that is what should display while a partial video file is downloading.  If you are not successful with that, deal with the build warning and if possible the right-click menu update.

raveit65:
Can please do a PR in master branch?

src/file-manager/fm-directory-view.c
```diff
@@ -8743,7 +8743,14 @@ real_update_menus (FMDirectoryView *view)
 	for (l = selection; l != NULL; l = l->next) {
 		CajaFile *file;
 
-		file = CAJA_FILE (selection->data);
+		file = CAJA_FILE (l->data);
```
Max-E: I threw this change in because the code looked wrong. I think this was a bug, but it was being made harmless by the behavior of caja_mime_get_default_application_for_files-- see the commit message for details.

Max-E:
I think the warning should be fixed now, and I got context-menus updating as well. As far as I can tell, the thumbnail already _does_ turn into a ticking clock when the update happens, but the thumbnailer must not get run until after the file is done being actively modified.

Max-E:
Retargeted to master and retested.

libcaja-private/caja-file.c
```diff
@@ -2555,6 +2555,25 @@ caja_file_update_info (CajaFile *file,
 	return update_info_internal (file, info, FALSE);
 }
 
+void
+caja_file_refresh_info (CajaFile *file)
+{
+	GFile *gfile;
+	GError *error;
```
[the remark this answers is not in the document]
cwendling: It *must* be initialized to `NULL`, and yes, it should be freed.

If keeping `error`:
```suggestion
	GError *error = NULL;
```

libcaja-private/caja-file.c
```diff
@@ -2555,6 +2555,25 @@ caja_file_update_info (CajaFile *file,
...
+
+	gfile = caja_file_get_location (file);
+	new_info = g_file_query_info (gfile, CAJA_FILE_DEFAULT_ATTRIBUTES,
+	                              G_FILE_QUERY_INFO_NONE, NULL, &error);
+	if (new_info != NULL) {
+		if (caja_file_update_info (file, new_info)) {
+			caja_file_changed (file);
+		}
+		g_object_unref (new_info);
+	}
```
cwendling: If keeping `error`:
```suggestion
	} else {
		g_clear_error (&error);
	}
```

libcaja-private/caja-file.c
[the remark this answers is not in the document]
cwendling: However, here `error` seems unused, so `NULL` should probably simply be passed to `g_file_query_info()` below.

libcaja-private/caja-file.c
```diff
@@ -2555,6 +2555,25 @@ caja_file_update_info (CajaFile *file,
...
+void
+caja_file_refresh_info (CajaFile *file)
+{
+	GFile *gfile;
+	GError *error;
+	GFileInfo *new_info;
+
+	gfile = caja_file_get_location (file);
+	new_info = g_file_query_info (gfile, CAJA_FILE_DEFAULT_ATTRIBUTES,
+	                              G_FILE_QUERY_INFO_NONE, NULL, &error);
```
cwendling: If *not* keeping `error`:
```suggestion
	                              G_FILE_QUERY_INFO_NONE, NULL, NULL);
```

raveit65:
Ok, in general it works with coping a video file with the test program. I can open the unfinished copied file with a video-player.

lukefromdc:
The build warning is gone, in fact that file now builds clean giving only
`  CC       caja-mime-actions.lo` , and the finished build works. It takes a few seconds for the right-click menu to update, but it does update and simply clicking on the partially copied (headers in front mp4)  file opens in in mpv(my default video player) right from the start. Desktop thumbnail changes at the end but that's less of an issue and should not block this.

PR #1108, opened 2019-01-13
configure.ac: Add check for PKG_CONFIG being set

sc0w:
Add check for PKG_CONFIG being set.

Author: @sunweaver 

not sure how to test this PR, I just open this with the debian patch:

https://salsa.debian.org/debian-mate-team/caja/blob/debian/1.20.3-1/debian/patches/1001_use-correct-macros-to-cross-build.patch

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -54,7 +54,11 @@ AC_PROG_LN_S
 AC_PROG_MAKE_SET
 AM_DISABLE_STATIC
 AM_PROG_LIBTOOL
-AC_PATH_PROG(PKG_CONFIG, pkg-config, no)
+PKG_PROG_PKG_CONFIG
+dnl Give error and exit if we dont have pkgconfig
+if test "x$PKG_CONFIG" = "x"; then
+  AC_MSG_ERROR([you need to have pkgconfig installed!])
+fi
 
 AC_CHECK_LIB(m, floor)
 


configure.ac
```diff
@@ -54,7 +54,11 @@ AC_PROG_LN_S
 AC_PROG_MAKE_SET
 AM_DISABLE_STATIC
 AM_PROG_LIBTOOL
-AC_PATH_PROG(PKG_CONFIG, pkg-config, no)
+PKG_PROG_PKG_CONFIG
+dnl Give error and exit if we dont have pkgconfig
+if test "x$HAVE_PKGCONFIG" = "xno"; then
```
cwendling: is this guaranteed to be set?  I don't see any mention of it in the manual page, and reading the macro's source code suggests there are at least some cases where it won't be set.  Actually, nothing references it there.

A better check seems to be whether `$PKG_CONFIG` is non-empty, which although not explicitly documented is the behavior of the macro.

```suggestion
if test "x$PKG_CONFIG" = "x"; then
```

configure.ac
```diff
@@ -54,7 +54,11 @@ AC_PROG_LN_S
...
+  AC_MSG_ERROR([you need to have pkgconfig installed !])
```
cwendling: Cosmetics:
```suggestion
  AC_MSG_ERROR([you need to have pkgconfig installed!])
```

raveit65:
What is the advantage of using `AS_IF` instead of `if`?

configure.ac
sunweaver: cwendling is correct, you need to check if $PKG_CONFIG is set (to the full path of the build host's pkg-config executable. If it is unset, pkg-config is missing and builds should abort.

configure.ac
```diff
@@ -148,7 +152,7 @@ AC_ARG_ENABLE(xmp,
     AC_HELP_STRING([--disable-xmp],
                    [build without xmp support]))
 msg_xmp=no
-if test "x$enable_xmp" != "xno"; then
+AS_IF([test "x$enable_xmp" != "xno"],[
```
sunweaver: This AS_IF change is not necessarily required by this PR. It founds its way into the patch I ship with the Debian package. But it is off-topic here. Sorry for the confusion.

configure.ac
```diff
@@ -192,10 +196,10 @@ AC_ARG_ENABLE(packagekit,
     AC_HELP_STRING([--disable-packagekit],
         [build without PackageKit support]))
 msg_packagekit=no
-if test "x$enable_packagekit" != "xno"; then
+AS_IF([test "x$enable_packagekit" != "xno"],[
```
sunweaver: Same here (AS_IF off-topic)

configure.ac
```diff
@@ -213,7 +217,7 @@ else
 fi
 ])
 AC_MSG_CHECKING(for more warnings, including -Werror)
-if test "$GCC" = "yes" -a "$set_more_warnings" != "no"; then
+AS_IF([test "$GCC" = "yes" -a "$set_more_warnings" != "no"],[
```
sunweaver: Same here (AS_IF off-topic)

configure.ac
```diff
@@ -316,9 +320,9 @@ AM_CONDITIONAL(ENABLE_UPDATE_MIMEDB, test x$enable_update_mimedb = xyes)
 
 AC_ARG_ENABLE(icon-update, AC_HELP_STRING([--disable-icon-update],
                                         [Disable icon cache update]))
-if (test "$enable_icon_update" != no); then
+AS_IF([test "$enable_icon_update" != no],[
```
sunweaver: Same here (AS_IF off-topic)

sc0w:
@cwendling @sunweaver I did your requested changes

PR #1110, opened 2019-01-16
caja-application: Check for smclient before trying to load session

lukefromdc:
Fix crash on starting with env DESKTOP_AUTOSTART_ID=ANY_CHARS caja

### DIFF
--- a/src/caja-application.c
+++ b/src/caja-application.c
@@ -1741,10 +1741,10 @@ caja_application_load_session (CajaApplication *application)
     GKeyFile *state_file;
     char *data;
     caja_application_smclient_initialize (application);
-   if (!egg_sm_client_is_resumed (application->smclient))
+    if (!egg_sm_client_is_resumed (application->smclient))
     {
         return;
-  } 
+    }
 
     state_file = egg_sm_client_get_state_file (application->smclient);
     if (!state_file)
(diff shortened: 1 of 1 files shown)


yetist:
The warning disappeared when running the second caja instance, but the file manager was not opened.

yetist:
> I am still wondering what `DESKTOP_AUTOSTART_ID` means?

If an application runs automatically, then mate-session adds the `DESKTOP_AUTOSTART_ID` variable to the running environment of the application, so its subprocesses inherit the variable.

For example, if `xterm` and `guake` run automatically, they pass this variable to the child process. If then run Caja in them, these warnings will appear and Caja cannot be opened.

But for the mate desktop, mate-terminal deletes the variable at runtime and does not pass it to the child process, so caja will open, and gnome-terminal does the same. So, I think for now, avoid the warnings is enough.

How to test?
```
$ cp /usr/share/applications/xterm.desktop /usr/share/applications/mate-terminal.desktop ~/.config/autostart/
```
Then logout and login, will see xterm and mate-terminal is running, then run the follow in them:
```
$ strings /proc/$PPID/environ |grep DESKTOP_AUTOSTART_ID
$ echo $DESKTOP_AUTOSTART_ID
$ caja
```
In xterm, will see `DESKTOP_AUTOSTART_ID` is set , and caja show the warnings, and not open.
In mate-terminal, `DESKTOP_AUTOSTART_ID` is not set, and caja will open a window.

PR #1117, opened 2019-01-24
Support querying files by modification time and size

yetist:
Thanks for Leslie Zhai's patch, UI like this:

Please review and test it, thanks.

### DIFF
--- a/src/caja-query-editor.c
+++ b/src/caja-query-editor.c
@@ -35,6 +35,17 @@
 #include <gdk/gdkkeysyms.h>
 #include <gtk/gtk.h>
 
+enum
+{
+        DURATION_INVALID,
+        DURATION_ONE_HOUR,
+        DURATION_ONE_DAY,
+        DURATION_ONE_WEEK,
+        DURATION_ONE_MONTH,
+        DURATION_SIX_MONTHS,
+        DURATION_ONE_YEAR,
+};
+
 typedef enum
 {
     CAJA_QUERY_EDITOR_ROW_LOCATION,
@@ -1054,7 +1065,7 @@ static GtkWidget *modtime_row_create_widgets(CajaQueryEditorRow *row)
 
     gtk_combo_box_set_active(GTK_COMBO_BOX(combo), 0);
 
-    duration_store = gtk_list_store_new(2, G_TYPE_LONG, G_TYPE_STRING);
+    duration_store = gtk_list_store_new(2, G_TYPE_INT, G_TYPE_STRING);
     duration_combo = gtk_combo_box_new_with_model(GTK_TREE_MODEL(duration_store));
     g_object_unref(duration_store);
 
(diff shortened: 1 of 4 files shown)


xiangzhai:
@lukefromdc I am Leslie Zhai :)

> Some test searches seemed to hang

Perhaps it needs to optimize the algorithm, for example O(N), about search function especially for *huge* filesystem testcase.

Regards,
Leslie Zhai

lukefromdc:
Caja can be set to list filesizes either by K/M/GiB or by K/M/GB in preferences, so the labels ideally would be following user preferences but that would require reading the gsettings value and then selecting which label text to use. 

File size comes from 
`gint64 file_size = g_file_info_get_size (info);`
I do not know which units that returns file size in.

src/caja-query-editor.c
```diff
@@ -999,6 +1025,231 @@ type_add_rows_from_query (CajaQueryEditor    *editor,
...
+    gtk_list_store_append(duration_store, &iter);
+    gtk_list_store_set(duration_store, &iter, 0, 86400, 1, _("1 Day"), -1);
+    gtk_list_store_append(duration_store, &iter);
+    gtk_list_store_set(duration_store, &iter, 0, 604800, 1, _("1 Week"), -1);
+    gtk_list_store_append(duration_store, &iter);
+    gtk_list_store_set(duration_store, &iter, 0, 2419200, 1, _("1 Month"), -1);
+    gtk_list_store_append(duration_store, &iter);
+    gtk_list_store_set(duration_store, &iter, 0, 14515200, 1, _("6 Months"), -1);
+    gtk_list_store_append(duration_store, &iter);
+    gtk_list_store_set(duration_store, &iter, 0, 29030400, 1, _("1 Year"), -1);
```
vkareh: These values seem wrong. Here's what I noticed:
🗸 3600: 1 hour
🗸 86400: 1 day, 24 hours
🗸 604800: 7 days, 168 hours
🗴 2419200: 28 days, arguably correct?
🗴 14515200: 168 days, definitely wrong (should be around 15778800, 31557600 / 2, or about 183 days)
🗴 29030400: 336 days, definitely wrong (should be around 31557600, 365.25 days)

src/caja-query-editor.c
```diff
@@ -999,6 +1025,231 @@ type_add_rows_from_query (CajaQueryEditor    *editor,
...
+    gtk_list_store_append(size_store, &iter);
+    gtk_list_store_set(size_store, &iter, 0, 104857600, 1, _("100 MB"), -1);
+    gtk_list_store_append(size_store, &iter);
+    gtk_list_store_set(size_store, &iter, 0, 524288000, 1, _("500 MB"), -1);
+    gtk_list_store_append(size_store, &iter);
+    gtk_list_store_set(size_store, &iter, 0, 1073741824, 1, _("1 GB"), -1);
+    gtk_list_store_append(size_store, &iter);
+    gtk_list_store_set(size_store, &iter, 0, 2147483648, 1, _("2 GB"), -1);
+    gtk_list_store_append(size_store, &iter);
+    gtk_list_store_set(size_store, &iter, 0, 4294967296, 1, _("4 GB"), -1);
```
vkareh: Agree with the comments above. These units (and their value) should reflect the users preference

lukefromdc:
The filesize labels now follow user preferences. One minor quibble is this: on toggling "show file sizes with IEC units" in File Management Preferences>Display the labels do not update until the search results window has been toggled away from "size" and back  or the search closed and an new search started (which returns the search subwindow to "location" . 

Since it would probably not be common for a user to toggle this preference in the middle of a search I'm not sure if its worth the extra work (and code at runtime) to connect to the "changed" signal for that gsettings preference to update the labels or not. Still this would work, and would catch the case of a user doing a filesize search and deciding at that point to reset the filesize units preference for Caja to use a particular size cutoff.

raveit65:
I got build warnings.
```
caja-query.c: In function 'caja_query_to_xml':
caja-query.c:387:12: warning: unused variable 'size' [-Wunused-variable]
     gint64 size;
            ^~~~
caja-query.c:386:12: warning: unused variable 'timestamp' [-Wunused-variable]
     gint64 timestamp;
            ^~~~~~~~~
```
```
caja-query-editor.c: In function 'modtime_row_add_to_query':
caja-query-editor.c:1162:17: warning: 'datetime' may be used uninitialized in this function [-Wmaybe-uninitialized]
     timestamp = g_date_time_to_unix (datetime);
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```
Not sure if the following is related to PR
```
caja-search-engine-simple.c: In function 'visit_directory':
caja-search-engine-simple.c:350:17: warning: unused variable 'tag' [-Wunused-variable]
     const char *tag, *mime_type, *display_name;
```

yetist:
Okey, I'll check it.

yetist:
@raveit65 
New PR here: #1127

PR #1120, opened 2019-01-26
Avoid deprecated g_type_class_add_private under libcaja-private

yetist:
Changed CajaDesktopLink, CajaDirectoryAsync and CajaDirectory

### DIFF
--- a/libcaja-private/caja-directory.c
+++ b/libcaja-private/caja-directory.c
@@ -54,8 +54,7 @@ static guint signals[LAST_SIGNAL] = { 0 };
 static GHashTable *directories;
 
 static void               caja_directory_finalize         (GObject                *object);
-static void               caja_directory_init             (gpointer                object,
-        gpointer                klass);
+static void               caja_directory_init             (CajaDirectory *directory);
 static void               caja_directory_class_init (CajaDirectoryClass *klass);
 static CajaDirectory *caja_directory_new              (GFile                  *location);
 static char *             real_get_name_for_self_as_new_file  (CajaDirectory      *directory);
@@ -64,9 +63,7 @@ static gboolean		  real_is_editable                    (CajaDirectory      *dire
 static void               set_directory_location              (CajaDirectory      *directory,
         GFile                  *location);
 
-EEL_CLASS_BOILERPLATE (CajaDirectory,
-                       caja_directory,
-                       G_TYPE_OBJECT)
+G_DEFINE_TYPE_WITH_PRIVATE (CajaDirectory, caja_directory, G_TYPE_OBJECT)
 
 static void
 caja_directory_class_init (CajaDirectoryClass *klass)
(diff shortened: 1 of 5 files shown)


raveit65:

libcaja-private/caja-directory.c
```diff
@@ -522,9 +514,12 @@ caja_directory_are_all_files_seen (CajaDirectory *directory)
...
 
-    return EEL_CALL_METHOD_WITH_RETURN_VALUE
-           (CAJA_DIRECTORY_CLASS, directory,
-            are_all_files_seen, (directory));
+    if (CAJA_DIRECTORY_GET_CLASS(directory)->are_all_files_seen == NULL)
+    {
+        return FALSE;
+    } else {
+        return CAJA_DIRECTORY_GET_CLASS(directory)->are_all_files_seen (directory);
+    }
```
cwendling: brace style

libcaja-private/caja-directory.c
```diff
@@ -1454,29 +1457,34 @@ caja_directory_file_monitor_remove (CajaDirectory *directory,
...
 
-    return EEL_CALL_METHOD_WITH_RETURN_VALUE
-           (CAJA_DIRECTORY_CLASS, directory,
-            is_not_empty, (directory));
+    if (CAJA_DIRECTORY_GET_CLASS(directory)->is_not_empty == NULL)
+    {
+        return FALSE;
+    } else {
+        return CAJA_DIRECTORY_GET_CLASS(directory)->is_not_empty (directory);
+    }
```
cwendling: same, and below too

libcaja-private/caja-directory.h
```diff
@@ -53,18 +56,12 @@
 #define CAJA_DIRECTORY_GET_CLASS(obj) \
   (G_TYPE_INSTANCE_GET_CLASS ((obj), CAJA_TYPE_DIRECTORY, CajaDirectoryClass))
 
-/* CajaFile is defined both here and in caja-file.h. */
-#ifndef CAJA_FILE_DEFINED
-#define CAJA_FILE_DEFINED
-typedef struct CajaFile CajaFile;
-#endif
```
cwendling: was there a reason for this?  otherwise, it's better indeed, but make sure *caja-file.h* doesn't have conditional dependencies on this header

yetist: This is good, but I think it would be better to use another PR to modify this. So, I will revert it in this PR.

libcaja-private/caja-directory.c
yetist: Can you give me an example? What kind of style will be better, I really don't know, thank you.

cwendling: Look at the code around; apart from a few inconsistencies, it's putting braces on their own lines, e.g.

```c
    if (CAJA_DIRECTORY_GET_CLASS(directory)->are_all_files_seen == NULL)
    {
        return FALSE;
    }
    else
    {
        return CAJA_DIRECTORY_GET_CLASS(directory)->are_all_files_seen (directory);
    }
```

PR #1124, opened 2019-01-26
Avoid deprecated g_type_class_add_private under src

yetist:
Changed CajaInformationPanel, CajaLocationBar and CajaNavigationAction

### DIFF
--- a/src/caja-navigation-action.c
+++ b/src/caja-navigation-action.c
@@ -39,11 +39,7 @@
 static void caja_navigation_action_init       (CajaNavigationAction *action);
 static void caja_navigation_action_class_init (CajaNavigationActionClass *class);
 
-static GObjectClass *parent_class = NULL;
-
-#define CAJA_NAVIGATION_ACTION_GET_PRIVATE(object)(G_TYPE_INSTANCE_GET_PRIVATE ((object), CAJA_TYPE_NAVIGATION_ACTION, CajaNavigationActionPrivate))
-
-struct CajaNavigationActionPrivate
+struct _CajaNavigationActionPrivate
 {
     CajaNavigationWindow *window;
     CajaNavigationDirection direction;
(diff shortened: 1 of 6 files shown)


raveit65:

lukefromdc:
At least the code change for this one is not as complex as some of the others though it does remove `caja_navigation_action_get_type` I see that is exposed in caja-navigation-action.h but that a string search for "caja_navigation_action_get_type" in mate-search-tool does not find that string anywhere in the code except in caja-navigation-action.c, implying that that is unused code unless used by extensions (it's not found in our own caja-extensions package).

Thus removing that block and references to it should cause no problems unless a third party extension can use it, and the rest of the code changes in this particular PR are not so complex as some of the others.

yetist:
The _get_type function is generated by the G_DEFINE_TYPE macro.

lukefromdc:
OK, I can see that in the code now but I really don't know enough about these macros to catch that on my own.  Most of the rest of these PR's (other MATE apps) have been boilerplate that seemed to always work. Caja is quite complex, heck the original Nautilus was Easel's sole released product. The libunique->GtkApplication transition was also brutally complex, mostly guided by what Nautilus had done but not entirely because the 2.32>3.0 transition for Nautilus also dropped some features I had to add back. That was months of work spanning from MATE 1.15 to 1.17.

Nautilus still has this in seven files,  and seven more either converted to or used the G_DEFINE_TYPE_WITH_PRIVATE macro from the start of some other change:

nautilus_files_3 30

yetist:
@lukefromdc 

The nautilus files in the left window you pasted can see the compile warning in the build log:
https://gitlab.gnome.org/GNOME/nautilus/-/jobs/199162

src/caja-navigation-action.c
```diff
@@ -58,33 +54,7 @@ enum
     PROP_WINDOW
 };
 
-GType
-caja_navigation_action_get_type (void)
```
vkareh: The prototype for this function is still in the header file, and it's used in a type macro? what does this function do anyway?

vkareh: okay I should have read the earlier comments, but still there is a prototype for this function. If the function is generated by the macro, why do we need a prototype for it then?

lukefromdc:
Should I go ahead and merge this or is further review needed?

yetist:
I think there is no problem about `caja_navigation_action_get_type`:

1. The prototype of this function in the header file is required.
Maybe other files will call `caja_navigation_action_get_type` by `CAJA_TYPE_NAVIGATION_ACTION`, and I got one here: [src/caja-navigation-window-menus.c#L943](https://github.com/mate-desktop/caja/blob/master/src/caja-navigation-window-menus.c#L943)


2. `caja_navigation_action_get_type` is generated by the `G_DEFINE_TYPE_WITH_PRIVATE` macro in the body file.
This can be confirmed by the following command:
```
$ cat test.c 
#include <gtk/gtk.h>

G_DEFINE_TYPE_WITH_PRIVATE (CajaNavigationAction, caja_navigation_action, GTK_TYPE_ACTION)
$ gcc -E test.c `pkg-config --cflags gtk+-3.0` | tail | indent -
```

yetist:
Rebase on the master done.

PR #1125, opened 2019-01-26
Avoid deprecated g_type_class_add_private under src

yetist:
Changed CajaSidePane, CajaSpatialWindow, CajaViewAsAction and CajaXContentBar


raveit65:

PR #1133, opened 2019-02-06
Add --disable-self-check for configure

yetist:
Fixed make distcheck, see #1129

PR #1134, opened 2019-02-06
update travis to support archlinux and enable distcheck

yetist:
After merge #1133 , this can works fine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -118,3 +118,4 @@ variables:
 
 after_scripts:
   - make distcheck
+  - 'if [ $? -ne 0 ];then RED="\033[0;31m"; ENDC="\033[0m"; printf "${RED}!!! ERROR: Run make distcheck failed.${ENDC}\n"; fi'
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,6 @@ language: bash
 services:
   - docker
 
-# The branch just for test.
-#branches:
-#  only:
-#  - travis
-
 before_install:
   - curl -L -o docker-build https://github.com/mate-desktop/mate-dev-scripts/raw/master/travis/docker-build
   - chmod +x docker-build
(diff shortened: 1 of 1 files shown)


yetist:
`make distcheck` failed on debian.

Debian seems to check very strict, these files need to be added to POTFILES.in to fix it:
```
The following files contain translations and are currently not in use. Please
consider adding these to the POTFILES.in file, located in the po/ directory.
sub/data/caja-autorun-software.desktop.in
sub/data/caja-browser.desktop.in
sub/data/caja-computer.desktop.in
sub/data/caja-file-management-properties.desktop.in
sub/data/caja-folder-handler.desktop.in
sub/data/caja-home.desktop.in
sub/data/caja.desktop.in
If some of these files are left out on purpose then please add them to
POTFILES.skip instead of POTFILES.in. A file 'missing' containing this list
of left out files has been written in the current directory.
Please report to http://www.mate-desktop.org
```

raveit65:
They are as *.in.in files in POTFILES.in

```
data/caja-autorun-software.desktop.in.in
data/caja-browser.desktop.in.in
data/caja-computer.desktop.in.in
data/caja.desktop.in.in
data/caja-file-management-properties.desktop.in.in
data/caja-folder-handler.desktop.in.in
data/caja-home.desktop.in.in
```
If i recall correctly than adding them again as .in files willl´ break build.

yetist:
This line is in *.desktop.in.in files, is it useful?
```
X-MATE-Bugzilla-Version=@VERSION@
```
If it doesn't work, I think we can fix the distcheck fail on debian.

raveit65:
I think this is an old remnant from gnome2.
Still wondering if gnome use that today after they switched to gitlab and dropped bugzilla.
But i am not sure if Mate needs that.
@monsta 
Any idea?

raveit65:
It is already removed from nautilus https://gitlab.gnome.org/GNOME/nautilus/commit/3868b4cc871734fe9e9fa3d050dfd70b4b6d6932
We can remove it too.......from all repos :)

raveit65:
https://bugzilla.gnome.org/show_bug.cgi?id=776796

monsta:
Ok, that explains what it did originally at least. And our lines look like another case of excessive gnome -> mate renaming back in 2011... So it can be removed, I'm sure we never had any server or tool that parsed `X-MATE-Bugzilla` stuff :slightly_smiling_face:

PR #1139, opened 2019-02-06
Fix distcheck

yetist:
Fix `make distcheck` fail on debian.


caja.pot
```diff
@@ -6,9 +6,9 @@
 #, fuzzy
 msgid ""
 msgstr ""
-"Project-Id-Version: MATE Desktop Environment\n"
```
raveit65: PACKAGE VERSION is wrong, this should always be 
`"Project-Id-Version: MATE Desktop Environment\n"`

Project-Id-Version at Transifex is MATE Desktop Environment, so this need to be correct by hand after running ./makepot.
I will change this here.

yetist: OK, Thank you.

PR #1142, opened 2019-02-13
Update Documenters

rbuj:
Update "Documented by" in About Caja dialog


raveit65:
Maybe it's time to add all current team members to Contributors list?

lukefromdc:
We probably should, those files have gotten very out of date over the years as more people have worked on the code.

rbuj:
I added the new authors #1143

PR #1143, opened 2019-02-14
Update authors

rbuj:
Merged the authors with the new ones:

git log --pretty="%an <%ae>" --since "2012-01-01" -- . "*.h" "*.c" | sort | uniq


raveit65:
@rbuj 
I think it's time to do that for all repos after all the years of MATE desktop,
if you have some free resources to do this.

rbuj:
@raveit65 
Ok, I think that we can use a resource file named app-name.about like mate-terminal
https://github.com/mate-desktop/mate-terminal/blob/7ab20eb6f84a057d0b255cd2bd35c9bf9834fc45/src/terminal-window.c#L4348

raveit65:
Good idea, such a resource file is much easier to update.

PR #1151, opened 2019-02-20
Finished and updated the italian translation

pittix:
All lines are translated and some were fixed for clarity.
I didn't make mistakes, I think, but another look could be useful.


raveit65:
Please use transifex to translate it.po and become a member of your language team.
https://www.transifex.com/mate/MATE/dashboard/

PR #1152, opened 2019-02-20
Fix attempt to free released memory

rbuj:


sc0w:
This PR fixes the Clang Analyzer warning:

PR #1159, opened 2019-02-23
caja-information-panel: Fix use of memory after it is freed

sc0w:

```
caja-information-panel.c:954:35: warning: Use of memory after it is freed
                    temp_button = gtk_button_new_with_label (button_name);
                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/caja-information-panel.c
+++ b/src/caja-information-panel.c
@@ -945,6 +945,9 @@ add_buttons_from_metadata (CajaInformationPanel *information_panel, const char *
             *temp_str = '\0';
             if (!g_ascii_strcasecmp (current_term, "button"))
             {
+                if (button_name)
+                    g_free (button_name);
+
                 button_name = g_strdup (temp_str + 1);
             }
             else if (!g_ascii_strcasecmp (current_term, "script"))
@@ -958,7 +961,6 @@ add_buttons_from_metadata (CajaInformationPanel *information_panel, const char *
                                         0);
                     information_panel->details->has_buttons = TRUE;
                     command_string = g_strdup (temp_str + 1);
-                    g_free (button_name);
 
                     g_signal_connect_data (temp_button,
                                            "clicked",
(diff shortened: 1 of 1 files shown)


src/caja-information-panel.c
```diff
@@ -960,6 +960,8 @@ add_buttons_from_metadata (CajaInformationPanel *information_panel, const char *
                     command_string = g_strdup (temp_str + 1);
                     g_free (button_name);
 
+                    button_name = NULL;
```
rbuj: Just add `g_free (button_name);` after `for (index = 0; (term = terms[index]) != NULL; index++)`, before `g_strfreev (terms);`
https://github.com/mate-desktop/caja/blob/e6421559dd1c4431502551d665c90d44f08e40ce/src/caja-information-panel.c#L980

sc0w: confirmed, the warning in Clang with your suggested change is gone

other question, we have into the `loop for` the line:

`button_name = g_strdup (temp_str + 1);`

and `g_strdup` needs to be freed, so, we need to add before this line `if (button_name) g_free (button_name);` ?

If `g_strdup` is called several times without freed we have memory leak

rbuj: I'm agree, free **button_name** before new memory allocation, and after the loop end.

You can move `g_free (button_name);` from line
https://github.com/mate-desktop/caja/blob/e6421559dd1c4431502551d665c90d44f08e40ce/src/caja-information-panel.c#L961
to line
https://github.com/mate-desktop/caja/blob/e6421559dd1c4431502551d665c90d44f08e40ce/src/caja-information-panel.c#L948
And add `g_free (button_name);` before the `g_strfreev (terms);`

For loop: button iteration followed by script iterarion -> proced to create button
For loop: button iteration followed by button iterarion -> can't create button without the script

sc0w: @rbuj  I did changes, please review again

PR #1160, opened 2019-02-23
fm-tree-model: Fix use of memory after it is freed

sc0w:

```
fm-tree-model.c:784:9: warning: Use of memory after it is freed
        destroy_node_without_reporting (model, parent->first_child);
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/file-manager/fm-tree-model.c
+++ b/src/file-manager/fm-tree-model.c
@@ -779,9 +779,13 @@ stop_monitoring_directory (FMTreeModel *model, TreeNode *node)
 static void
 destroy_children_without_reporting (FMTreeModel *model, TreeNode *parent)
 {
-    while (parent->first_child != NULL)
+    TreeNode *current_child = parent->first_child;
+    TreeNode *next_child;
+    while (current_child != NULL)
     {
-        destroy_node_without_reporting (model, parent->first_child);
+        next_child = current_child->next;
+        destroy_node_without_reporting (model, current_child);
+        current_child = next_child;
     }
 }
 


src/file-manager/fm-tree-model.c
```diff
@@ -782,6 +782,7 @@ destroy_children_without_reporting (FMTreeModel *model, TreeNode *parent)
     while (parent->first_child != NULL)
     {
         destroy_node_without_reporting (model, parent->first_child);
+        parent->first_child = NULL;
```
rbuj: At first sight, this breaks [tree transversal](https://en.wikipedia.org/wiki/Tree_traversal).

sc0w: thoughts are welcome :)

rbuj: Set parent->first_child to sibling node before the line:
https://github.com/mate-desktop/caja/blob/e0c34cb3f8fe32e584c5a8f3108de65790ab2c56/src/file-manager/fm-tree-model.c#L795
```
(node->parent)->first_child = node->next;
```

rbuj: And update sibling pointers, for example:
```
if (node->prev != NULL) { 
    (node->prev)->next = node->next;
}
if (node->next != NULL) { 
    (node->next)->prev = node->prev;
}
```

sc0w: It seems your proposed changes doesn't avoid the clang warning, I tried with these changes:
```patch
index 500b905..e93993e 100644
--- a/src/file-manager/fm-tree-model.c
+++ b/src/file-manager/fm-tree-model.c
@@ -791,6 +791,13 @@ destroy_node_without_reporting (FMTreeModel *model, TreeNode *node)
     abandon_node_ref_count (model, node);
     stop_monitoring_directory (model, node);
     node->inserted = FALSE;
+    (node->parent)->first_child = node->next;
+    if (node->prev != NULL) { 
+        (node->prev)->next = node->next;
+    }
+    if (node->next != NULL) { 
+        (node->next)->prev = node->prev;
+    }
     destroy_children_without_reporting (model, node);
     g_hash_table_remove (node->root->file_to_node_map, node->file);
     tree_node_destroy (model, node);
```

rbuj: This line uses always the same arguments, which should be updated.
https://github.com/mate-desktop/caja/blob/e0c34cb3f8fe32e584c5a8f3108de65790ab2c56/src/file-manager/fm-tree-model.c#L784

Node delelion is performed at:
https://github.com/mate-desktop/caja/blob/da3b63420d2b34afe97e72e272078aa13622af8e/src/file-manager/fm-tree-model.c#L212

rbuj: it's not else if...
https://github.com/mate-desktop/caja/blob/da3b63420d2b34afe97e72e272078aa13622af8e/src/file-manager/fm-tree-model.c#L200

shoudl be `if (prev != NULL)`

rbuj:
```patch
--- a/src/file-manager/fm-tree-model.c
+++ b/src/file-manager/fm-tree-model.c
@@ -779,9 +779,13 @@ stop_monitoring_directory (FMTreeModel *model, TreeNode *node)
 static void
 destroy_children_without_reporting (FMTreeModel *model, TreeNode *parent)
 {
-    while (parent->first_child != NULL)
+    TreeNode *current_child = parent->first_child;
+    TreeNode *next_child;
+    while (current_child != NULL)
     {
-        destroy_node_without_reporting (model, parent->first_child);
+        next_child = current_child->next;
+        destroy_node_without_reporting (model, current_child);
+        current_child = next_child;
     }
 }
```

sc0w: @rbuj 

well done, thanks!

the warning in clang is gone!

done!

lukefromdc:
Note that in testing these I am getting 
`gtk_container_foreach: assertion 'GTK_IS_CONTAINER (container)' failed`
warnings in .xsession-errors but those are also present with master and can be disregarded for the purpose of testing these PR's. These warnings appear on closing a caja navigation window

PR #1162, opened 2019-02-25
caja-file: Fix use of memory after it is freed

sc0w:

```
caja-file.c:8326:3: warning: Use of memory after it is freed
                caja_file_call_when_ready (file,
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/libcaja-private/caja-file.c
+++ b/libcaja-private/caja-file.c
@@ -8299,10 +8299,10 @@ file_list_file_ready_callback (CajaFile *file,
 
 void
 caja_file_list_call_when_ready (GList *file_list,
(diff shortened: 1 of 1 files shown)


sc0w:
how this PR works:

`caja_file_call_when_ready` frees `data`

so, in the next line I set value to `data` to use in the loop and to avoid the use of freed memory

and after the loop, I free `data` with `file_list_ready_data_free (data);`

that's all

@vkareh I don't know if this can be done in another way, and I don't know how to test the function, but if caja works better with this PR for you, I think is ready to go :)

PR #1166, opened 2019-03-03
eel-canvas: Fix warning: Undefined or garbage value returned to caller

sc0w:

```
eel-canvas.c:3426:9: warning: Undefined or garbage value returned to caller
        return item;
        ^~~~~~~~~~~
```


vkareh:
> CLANG authors who are presumably better programmers than we are

Whaaaaa? :grin:

PR #1172, opened 2019-03-05
eel: replace eel_ref_str with GRefString

rbuj:
GNOME/nautilus@b4d200f

https://developer.gnome.org/glib/stable/glib-Reference-counted-strings.html


vkareh:
@lukefromdc - which distros are shipping MATE 1.22 currently? I agree that we shouldn't merge until we can confirm that at least the major distros that our team uses are at glib 2.58.

I use Fedora 29 and Ubuntu 18.10 - those are both using 2.58.3 and 2.58.1-2, respectively.

vkareh:
@lukefromdc - also, since 1.22 is already out, and this is not a bug fix, there's really no urgent need to backport this to 1.22. We can also see when Nautilus releases the next point version that includes their glib 2.58.1 commit and use that as our guide regarding when it would be safe to release this. (Assuming that the GNOME team does their due diligence in not releasing breaking changes...)

raveit65:
I just update Fedora30 (next) with mate-1.22 and  glib2-2.60.0-2.fc30.x86_64 is used here.
Upgrading f29 is on my list, but maybe i wait until we do a point release.
Not sure about f28 because it becomes EOL usually after 1 month of f30 release.

raveit65:
Looks like we better use it only for master because of situation in OpenSuse.
@XRevan86  ???

vkareh:
If openSUSE will still have glib2 2.54 when the next Leap version comes, they should patch it out rather than hold MATE back.
But yes, since this doesn't fix a bug nor adds additional features, there's no need to use it for 1.22

sc0w:
1.22 is stable branch

I vote to push this PR only to master

raveit65:
My vote is to push it only to master,
but only when @XRevan86 promised to give us a hint when OpenSuse is updated to glib2-2.60. :wink: 
Than we can push it to 1.22 too, IHMO.

PR #1173, opened 2019-03-07
caja-file-operations: add comment with false positive warning

sc0w:
avoid false positive Clang static analyzer warning:

```
caja-file-operations.c:958:11: warning: Cast a region whose size is not a multiple of the destination type size
        common = g_malloc0 (job_size);
                 ^~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/libcaja-private/caja-file-operations.c
+++ b/libcaja-private/caja-file-operations.c
@@ -955,6 +955,14 @@ init_common (gsize job_size,
 	CommonJob *common;
 	GdkScreen *screen;
 
+	/* expected warning with Clang static analyzer:                                *
+	 * "Cast a region whose size is not a multiple of the destination type size"   *
+	 *                                                                             *
+	 * It is expected for job_size to be larger than sizeof(CommonJob) no matter   *
+	 * what Clang analyzer reports: we're allocating the whole structure for a job *
+	 * (e.g. a TrashJob), but only initializing the common part of it (CommonJob)  *
+	 * which is a subset of all "real" job structures, structures that all start   *
+	 * with a CommonJob, and that thus can be used as such.                        */
 	common = g_malloc0 (job_size);
 
 	if (parent_window) {


libcaja-private/caja-file-operations.c
```diff
@@ -955,7 +955,7 @@ init_common (gsize job_size,
 	CommonJob *common;
 	GdkScreen *screen;
 
-	common = g_malloc0 (job_size);
+	common = g_malloc0 (job_size * sizeof (CommonJob));
```
cwendling: Don't do that, you'll be allocating an unreasonable amount of memory.

The warning is telling you that you're allocating something not a multiple of `sizeof(CommonJob)`, which is true, but also which is *what you want*.  `CommonJob` is the "base class" for all jobs, and this allocation allocates this, plus whatever job-specific extra there is.  Such scheme allows to consider every jobs being `CommonJob`s, as they have the same memory layout for the first `sizeof(CommonJob)` bytes, but after this they have specific extra memory.

This is a typical "check you're doing it right" type of warning, not a "there *definitely* is a *bug* in your program" one.
I'm not sure what you can do if you just wanna hide this warning that is a *false positive*, maybe cast to `void*` or something.  But don't just allocate a lot more memory just to silence this: it means multiplying by about 92 on amd64, which means that for example the smallest job, `EmptyTrashJob` will go from using about 120 bytes to about 11040 bytes (!!). Similarly, a bigger job like `CreateJob` which takes about 180 bytes now would then take 16560 (16+kb) bytes.

lukefromdc: That is entirely unacceptable. Better to simply add a comment to the existing code referring to the clang warning and stating that this is the intended behavior, to allocate only the amount of RAM actually needed. Bloating up software's RAM consumption, when a whole OS does this, contributes to the "page's law" mess, in which new computers with new software installed run no faster (or in this case have no more extra RAM), than old ones with older, lighter software.

lukefromdc:
Didn't meant to re-request review by  @cwendling , that was an accident in dismissing my own review

libcaja-private/caja-file-operations.c
sc0w: @cwendling thanks for the info, I like to silence the warning, I tried, but I don't know how to do that

rbuj:
I don't know whether it's a silly question, in such case, I'm sorry...
In general, not only for caja, to avoid typeof, offset, type enum, casting...,  can we use `_Generic` (C11) or only `GVariant` (glib)?

vkareh:
Looking at the original code, it seems correct. At this point anything we do to silence the warning would only accomplish silencing the warning. There are things we can do for that:
```c
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-W%whatever-flag-causes-this-warning%"
common = g_malloc0 (job_size);
#pragma clang diagnostic pop
```

But that's ugly, and a "proper fix" would really be rewritting the macro so that we have a different macro for each Job sub-type - definitely not a good idea.

I suggest leaving as is and putting in a comment right before the `g_malloc0` line to let the devs know that it causes a warning...

sc0w:
@lukefromdc @cwendling @rbuj @vkareh 

I did new changes, please take a look

sc0w:
@yetist 

I rebased the PR against master, and https://caja.mate-desktop.dev was updated sucesfully :)

lukefromdc:
I don't care for that particular approach, as we are just telling clang analyzer to look at one piece of code while building another. Better to just put a comment on the code and live with the warning than play this sort of game

cwendling:
> I don't care for that particular approach, as we are just telling clang analyzer to look at one piece of code while (EDIT: we are normally ) building another. Better to just put a comment on the code and live with the warning than play this sort of game

Agreed, I don't think it's a good idea.  And as it's analyzing something different than what we're using in normal circumstances, it'll give a different set of warnings (fewer or more, but it won't reflect the "real" code state).

@sc0w Did you try something like this?
```c
gpointer ptr = g_malloc0 (job_size);
common = ptr;
// or
common = (gpointer) g_malloc0 (job_size);
```
Or maybe clang is not dumb enough to get fooled by that?

Or if we really care, take the allocation out of `init_common()`, maybe something like this:
```c
#define op_job_new(__type, parent_window, should_start, can_pause) \
	((__type *)(init_common (g_malloc0 (sizeof(__type)), parent_window, should_start, can_pause)))

static gpointer
init_common (gpointer job,
	     GtkWindow *parent_window,
	     gboolean should_start, gboolean can_pause)
{
	CommonJob *common = job;
// ...
```

But IMO it's also OK to just add a comment mentioning this warning is to be ignored. BTW, don't clang analyzer have warning suppression files for this kind of things?

sc0w:
@cwendling 

> gpointer ptr = g_malloc0 (job_size);
> common = ptr;
> // or
> common = (gpointer) g_malloc0 (job_size);

it doesn't work, Clang is very intelligent xD

I added the comment.

cwendling:
Yeah maybe stating what this is supposed to allocate and why it makes sense might be a good thing.  But current state is already better than before.

sc0w:
If someone suggests other text with more information, please free to put here, my english is not good xD

rbuj:
http://maemo.org/api_refs/4.0/gobject/howto-gobject.html

lukefromdc:
I don't understand the warning well enough to write exactly what we are doing to avoid allocating too much RAM. We can go with this as it stands if we need to.

libcaja-private/caja-file-operations.c
```diff
@@ -955,6 +955,8 @@ init_common (gsize job_size,
 	CommonJob *common;
 	GdkScreen *screen;
 
+	/* expected warning with Clang static analyzer:
+	   Cast a region whose size is not a multiple of the destination type size */
```
cwendling: Quick take, if you like it or wanna build on it:

*"It is expected for `job_size` to be larger than `sizeof(CommonJob)` no matter what Clang analyzer reports: we're allocating the whole structure for a job (e.g. a `TrashJob`), but only initializing the common part of it (`CommonJob`) which is a subset of all "real" job structures, structures that all start with a `CommonJob`, and that thus can be used as such."*

sc0w: ok, done, thanks!

sc0w:
@rbuj what do you mean? do you think it is possible to avoid the warning using GObject?

I have no idea how to do that, please feel free to open new PR :)

cwendling:
> @rbuj what do you mean? do you think it is possible to avoid the warning using GObject?

You could because it'd be GObject that'd be doing the allocation that Clang doesn't like.  But in this situation I don't think it's necessarily a great idea, given it doesn't seem to require any of the GObject advanced features and the only thing it wants is share some code by having a common base.

lukefromdc:
Latest change to the comment is a much more complete explanation. Thanks

PR #1178, opened 2019-03-09
eel-string: Fix Clang static analyzer warnings

sc0w:
Fixes Clang static analyzer warnings:

```
eel-string.c:325:9: warning: Null pointer passed as an argument to a 'nonnull' parameter
        memcpy (result_position, replacement, replacement_length);
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

```
eel-string.c:319:13: warning: Memory copy function overflows destination buffer
            memcpy (result_position, p, remaining_length);
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/eel/eel-string.c
+++ b/eel/eel-string.c
@@ -292,6 +292,10 @@ eel_str_replace_substring (const char *string,
     {
         return NULL;
     }
+    else if (replacement == NULL)
+    {
+        return (char *) string;
+    }
 
     substring_length = strlen (substring);
     replacement_length = eel_strlen (replacement);
--- a/eel/eel-string.c
+++ b/eel/eel-string.c
@@ -308,7 +308,8 @@ eel_str_replace_substring (const char *string,
         {
             break;
         }
-        result_length += replacement_length - substring_length;
+        if (replacement_length > substring_length)
+            result_length += replacement_length - substring_length;
     }
 
     result = g_malloc (result_length + 1);
@@ -329,7 +330,7 @@ eel_str_replace_substring (const char *string,
         memcpy (result_position, replacement, replacement_length);
         result_position += replacement_length;
     }
-    g_assert (result_position - result == result_length);
+
     result_position[0] = '\0';
 
     return result;


sc0w:
the function `eel_str_replace_substring` is called in `caja-file`: https://github.com/mate-desktop/caja/blob/7008505e2350814b72aac711f47185003f902310/libcaja-private/caja-file.c#L5553

and according to here: https://github.com/mate-desktop/caja/blob/7008505e2350814b72aac711f47185003f902310/eel/eel-string.c#L1041

seems if `replacement` is `NULL`, it must returns `string`:

```
char *
eel_str_replace_substring (const char *string,
                           const char *substring,
                           const char *replacement)
```

lukefromdc:
I don't really understand this code segment or what it does, so only made a functional test which seemed to work. Does NOT mean the code is actually correct

sc0w:
I added one more commit, to fix one more warning with clang analyzer

raveit65:
You broke `make check` , Please fix it.

```
[rave@mother caja]$ git bisect good
312394dfa42c07a3afa5089c6e3e5b9a1c1001ae is the first bad commit
commit 312394dfa42c07a3afa5089c6e3e5b9a1c1001ae
Author: Pablo Barciela <scow@riseup.net>
Date:   Sat Mar 9 21:45:50 2019 +0100

    eel-string: avoid 'memcpy' with NULL argument
    
    eel-string.c:325:9: warning: Null pointer passed as an argument to a 'nonnull' parameter
            memcpy (result_position, replacement, replacement_length);
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

:040000 040000 3fc2e18f96a1597302eebd3a197cddd1aeb3bfd2 8f37bde4e2b93de64c97dbb8ccdbcd87022afb94 M	eel
```

eel/eel-string.c
```diff
@@ -292,6 +292,10 @@ eel_str_replace_substring (const char *string,
     {
         return NULL;
     }
+    else if (replacement == NULL)
+    {
+        return (char *) string;
```
cwendling: This is *not* good: normally this function returns a newly allocated chunk of memory, and here you just return one of the (constant) arguments.
You should instead return a copy of `string`.

However, the check itself seems wrong: if a `NULL` `replacement` is a valid input for this function, it should be considered the same as a zero-length string, and thus remove occurrences of `substring` from `string`, not merely return an unchanged `string`.

eel/eel-string.c
```diff
@@ -304,7 +308,8 @@ eel_str_replace_substring (const char *string,
         {
             break;
         }
-        result_length += replacement_length - substring_length;
+        if (replacement_length > substring_length)
+            result_length += replacement_length - substring_length;
```
cwendling: why only expand?  If the replacement is shorter than the the substring it means the result will be shorter so there's no need to widen the allocation.

eel/eel-string.c
```diff
@@ -325,7 +330,7 @@ eel_str_replace_substring (const char *string,
         memcpy (result_position, replacement, replacement_length);
```
cwendling: The nonnull warning here is misleading and slightly wrong: actually it is perfectly valid to pass `NULL` as source parameter (`replacement` here), but *only if* the length to copy is 0 (`replacement_length` here).  Which here is the case AFAICT.

This is probably a glitch of the way the analyzer finds this issue, thinking `dest` ought to be non-null.  If really needed, this could be worked around like `replacement == NULL ? "" : replacement`.

cwendling: I can't comment on new line 324 (old 319), so I do it here:

I don't get this warning at all, the only thing I could possibly imagine would be that the analyzer has a bug and fails to understand line 329 and things `result_position` gets assigned the value of `substring_position` or `p` somehow.  Otherwise it doesn't make sense as `result_position` is an offset in `result`, which is a newly allocated chunk of memory that cannot overlap any part of `string` (which `p` and `substring_position` are offsets of)

eel/eel-string.c
sc0w: it was changed with https://github.com/mate-desktop/caja/commit/3ed089797f798ea47c40d6126d3c002efe81015f

Issue #1183, opened 2019-03-23
Incomplete file replacement dialog since commit 1dc8857824c287347a921003b695849dce812782 "caja-file: Fix use of memory after it is freed"


lukefromdc:
#### Expected behaviour
On dragging a file to a directory already containing a file by the same name, file replacement dialog looks like this:

![File_Replace_Dialog_FULL_3-22-2019](https://user-images.githubusercontent.com/14036792/54859003-6321a980-4d00-11e9-91ca-c1cef0f1a945.png)

#### Actual behaviour
Partial file replacement dialog shown as below:
![File_Replace_Dialog_partial_3-22-2019](https://user-images.githubusercontent.com/14036792/54859048-cf9ca880-4d00-11e9-95f3-1deac90b0fce.png)

#### Steps to reproduce the behaviour
Create a new directory and copy any file into it. Now try to copy there again by drag and drop (or presumably any other method)

#### MATE general version
1.22

#### Package version
Caja 1.22.0+ git master 3-22-2019

#### Linux Distribution
Debian Unstable last updated 3-22-2019

#### Link to downstream report of your Distribution
None as this is a local build

PR #1184, opened 2019-03-23
caja-file: Fix incomplete file replacement dialog

sc0w:
Fixes https://github.com/mate-desktop/caja/issues/1183

### DIFF
--- a/libcaja-private/caja-file.c
+++ b/libcaja-private/caja-file.c
@@ -8322,16 +8322,12 @@ caja_file_list_call_when_ready (GList *file_list,
 		/* Need to do this here, as the list can be modified by this call */
 		l = l->next;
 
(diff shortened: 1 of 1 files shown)


lukefromdc:
In testing the removal of the 
`file_list_ready_data_free (data);`
line I was able to confirm that we cannot free that variable here with this or Caja crashes, whether it is inside or outside the` if (file)` conditional block. Thus memory management should be OK, and this data is still in use.

sc0w:
@lukefromdc yes, the same here, I have tested again with this PR and adding in the same place the line `file_list_ready_data_free (data);` and craja crashes every time on click over a folder, so `data` doesn't need to be freed

in my investigations, `caja_file_call_when_ready` frees `data` if `file` is NULL

so, I think this  PR is safe to merge

PR #1189, opened 2019-03-25
caja-progress-info: make sure argument is initialized

sc0w:

```
caja-progress-info.c:626:17: warning: 1st function call argument is an uninitialized value
                g_notification_set_body (notification, _("Process completed"));
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/libcaja-private/caja-progress-info.c
+++ b/libcaja-private/caja-progress-info.c
@@ -595,9 +595,10 @@ update_status_icon_and_window (void)
     gboolean toshow;
     GIcon *icon;
     GNotification *notification;
+    gboolean show_notifications = g_settings_get_boolean (caja_preferences, CAJA_PREFERENCES_SHOW_NOTIFICATIONS);
     static gboolean window_shown = FALSE;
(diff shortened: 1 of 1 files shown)


sc0w:
@lukefromdc you can analyze the report:




First clang assumes the gsettings key is `false`, and later `true`, for me this is a false positive, or while the function is working the user can change the key?

well, anyway, the code is more readable with this PR

lukefromdc:
Oh yeah: that is a theoretical possibility: user is changing gsettings values while a big file copy job is running. That is more than theoretical, as users do other things while long jobs run. Thus, reading gsettings once only prevents the function from being "turned on" too late to be initialized. Gsettings is read once and the result stored in a boolean, forcing use of the same value thoughout the running of this block of code for any one pass through it. Thus it is not a false posituve, because I did not consider a race condition with a change to gsettings.

PR #1190, opened 2019-03-28
caja-open-with-dialog: make sure there is error before showing it

sc0w:

```
caja-open-with-dialog.c:252:100: warning: Access to field 'message' results in a dereference of a null pointer (loaded from variable 'error')
        message = g_strdup_printf (_("Could not add application to the application database: %s"), error->message);
                                                                                                   ^~~~~~~~~~~~~~
```

### DIFF
--- a/libcaja-private/caja-open-with-dialog.c
+++ b/libcaja-private/caja-open-with-dialog.c
@@ -249,12 +249,15 @@ add_or_find_application (CajaOpenWithDialog *dialog)
 
     if (app == NULL)
     {
-        message = g_strdup_printf (_("Could not add application to the application database: %s"), error->message);
+        message = g_strdup_printf (_("Could not add application to the application database: %s"), error ? error->message : _("Unknown error"));
         eel_show_error_dialog (_("Could not add application"),
                                message,
                                GTK_WINDOW (dialog));
         g_free (message);
-        g_error_free (error);
+
+        if (error)
+            g_error_free (error);
+
         return NULL;
     }
 


libcaja-private/caja-open-with-dialog.c
```diff
@@ -247,7 +247,7 @@ add_or_find_application (CajaOpenWithDialog *dialog)
         }
     }
 
-    if (app == NULL)
+    if ((app == NULL) && (error))
```
cwendling: that's cute, but normally `app` can only be `NULL` when `error` is set, so it's kind of pointless.

Also, this prevents the `return` inside this `if` it `error` was *really* not set, which leads to more bugs below, like using a `NULL` `app` as parameter to [`g_app_info_set_as_default_for_type()`](https://developer.gnome.org/gio/stable/GAppInfo.html#g-app-info-set-as-default-for-type) and friends.

If you really want to guard accessing `error`, I'd rather do that in the if's body, like `error ? error->message: _("unknown error")`.

sc0w:
I did new changes, @cwendling please review again

libcaja-private/caja-open-with-dialog.c
```diff
@@ -249,12 +249,18 @@ add_or_find_application (CajaOpenWithDialog *dialog)
...
+            g_error_free (error);
+        }
+        else
+            message = _("unknown error");
+
         eel_show_error_dialog (_("Could not add application"),
                                message,
                                GTK_WINDOW (dialog));
         g_free (message);
-        g_error_free (error);
```
cwendling: You could use [`g_clear_error()`](https://developer.gnome.org/glib/stable/glib-Error-Reporting.html#g-clear-error) which deals with a `NULL` error:
```diff
-        g_error_free (error);
+        g_clear_error (&error);
```

libcaja-private/caja-open-with-dialog.c
```diff
@@ -249,12 +249,18 @@ add_or_find_application (CajaOpenWithDialog *dialog)
...
     if (app == NULL)
     {
-        message = g_strdup_printf (_("Could not add application to the application database: %s"), error->message);
+        if (error)
+        {
+            message = g_strdup_printf (_("Could not add application to the application database: %s"), error->message);
+            g_error_free (error);
+        }
+        else
+            message = _("unknown error");
```
cwendling: This might not be clear enough as an error message.  I'd rather have the same one as in the normal case, just with the specific detail using this "unknown error".
Something like this:
```diff
-        message = g_strdup_printf (_("Could not add application to the application database: %s"), error->message);
+        message = g_strdup_printf (_("Could not add application to the application database: %s"), error ? error->message : _("unknown error"));
```
If you don't like ternary, you could easily use an extra variable:
```diff
-        message = g_strdup_printf (_("Could not add application to the application database: %s"), error->message);
+        const gchar *detail;
+
+        if (error)
+            detail = error->message;
+        else
+            detail = _("unknown error");
+        message = g_strdup_printf (_("Could not add application to the application database: %s"), detail);
```

cwendling:
> Builds and runs fine,though I still don't know how to invoke that error dialog(see above)

Not sure, maybe try and *Open With…* and then enter a garbage command line, maybe this will show up.  But for the case this PR adds (the case where `error` is `NULL`), there is no way to test short of altering the code to create an error, because it should never happen given that there isn't a bug in GIO or Caja on that regard.

libcaja-private/caja-open-with-dialog.c
sc0w: done

PR #1196, opened 2019-03-31
eel-string: Use 'va_copy' instead of 'G_VA_COPY'

sc0w:
Fixes `cppcheck` warnings:

```
[eel-string.c:681]: (error) va_list 'va' used before va_start() was called.
[eel-string.c:682]: (error) va_list 'va' used before va_start() was called.
[eel-string.c:683]: (error) va_list 'va' used before va_start() was called.
[eel-string.c:684]: (error) va_list 'va' used before va_start() was called.
[eel-string.c:697]: (error) va_list 'va' used before va_start() was called.
[eel-string.c:698]: (error) va_list 'va' used before va_start() was called.
[eel-string.c:699]: (error) va_list 'va' used before va_start() was called.
[eel-string.c:700]: (error) va_list 'va' used before va_start() was called.
[eel-string.c:709]: (error) va_list 'va' used before va_start() was called.
[eel-string.c:710]: (error) va_list 'va' used before va_start() was called.
[eel-string.c:714]: (error) va_list 'va' used before va_start() was called.
[eel-string.c:720]: (error) va_list 'va' used before va_start() was called.
[eel-string.c:722]: (error) va_list 'va' used before va_start() was called.
```

### DIFF
--- a/eel/eel-string.c
+++ b/eel/eel-string.c
@@ -678,7 +678,7 @@ eel_strdup_vprintf_with_custom (EelPrintfHandler *custom,
         {
             char *val;
 
-            G_VA_COPY(va, va_orig);
+            va_copy (va, va_orig);
             skip_to_arg (&va, args, custom, conversions[i].precision_pos);
             val = g_strdup_vprintf ("%d", va);
             va_end (va);
@@ -694,7 +694,7 @@ eel_strdup_vprintf_with_custom (EelPrintfHandler *custom,
         {
             char *val;
 
-            G_VA_COPY(va, va_orig);
+            va_copy (va, va_orig);
             skip_to_arg (&va, args, custom, conversions[i].width_pos);
             val = g_strdup_vprintf ("%d", va);
             va_end (va);
@@ -706,7 +706,7 @@ eel_strdup_vprintf_with_custom (EelPrintfHandler *custom,
             g_free (val);
         }
 
-        G_VA_COPY(va, va_orig);
+        va_copy (va, va_orig);
         skip_to_arg (&va, args, custom, conversions[i].arg_pos);
         type = args[conversions[i].arg_pos];
         if (type < 0)


sc0w:
I have already tested with `va_start` before `va_copy` and the build fails.

The documentation says:

> va_copy()
> 
> The va_copy() macro copies the (previously initialized) variable argument list src to dest. The behavior is as if va_start() were applied to dest with the same last argument, followed by the same number of va_arg() invocations that was used to reach the current state of src. 

the argument is already initialized before the function `eel_strdup_vprintf_with_custom` in some points:

```
    va_start (va, format);
    new = eel_strdup_vprintf_with_custom (handlers, format, va);
    va_end (va);
```

Seems like cppcheck doesn't know what exactly does `G_VA_COPY` , but it knows the function `va_copy`.

I mean, false positive warning.

cwendling:
couldn't cppcheck be taught about [`G_VA_COPY`](https://developer.gnome.org/glib/stable/glib-Miscellaneous-Macros.html#G-VA-COPY:CAPS)?  Maybe with something similar to `-DG_VA_COPY=va_copy` on cppcheck's command line?

The goal of `G_VA_COPY()` is portability alone, because `va_copy()` is not part of C89.  If we're happy with depending on systems with a working `va_copy()` (which a C99-complient one should be, yet fully C99-compliant systems are surprisingly rare), this change is OK, but I'm not sure what it actually means in term of portability (e.g. which systems would be impacted).

sc0w:
@cwendling this change works, are you agree?
```
#ifdef G_VA_COPY
            G_VA_COPY (va, va_orig);
#else
            va_copy (va, va_orig);
#endif
```

cwendling:
hum it should, but what's the point?  it's using G_VA_COPY if it's defined, but if it ain't current build would fail, wouldn't it?

Maybe you meant the other way around, if `va_copy` is defined use that, otherwise fallback on `G_VA_COPY`?

Either way, i'm not too fond of adding a bunch of ifdefs to hide false positives, but that's my take.

sc0w:
> Maybe you meant the other way around, if va_copy is defined use that, otherwise fallback on G_VA_COPY?

It doesn't work

cwendling:
> It doesn't work

you mean it doesn't make cppcheck happy, or it breaks the build?
if it doesn't make cppcehck happy it'd be because it doesn't see `va_copy` defined, which is possible but sad.
The problem with your suggestion is that it adds dead code: `G_VA_COPY` will always be defined if you include `glib.h`, so the `#else` is effectively dead code.

sc0w:
yes, anyway, C89 isn't deprecated?

cwendling:
Well, depends what you mean by that.  Yes, there's newer versions of the C standard, actually 3: C99, C11 and C18.  This however doesn't mean that in 2019 you can rely on every single feature of C18, not even C99.

I would however think that `va_copy` is a fairly safe feature to rely on, but I'm not sure.

sc0w:
@mate-desktop/core-team we must give support to the old C89 ?

I think support old standards is a wall in the development

PR #1199, opened 2019-04-02
Fix invocation with --geometry from the command line

lukefromdc:
Due to changes somewhere else windows called with --geometry must now be opened before splitedOptions is freed or the geometry passed to later functions gets corrrupted

Note that setting a width smaller than 654 px will generate GTK warnings (negative content width and similar) but will still mostly work.

### DIFF
--- a/src/caja-application.c
+++ b/src/caja-application.c
@@ -374,18 +374,25 @@ caja_application_open (GApplication *app,
         }
         sscanf (splitedOptions[2], "%d", &open_in_tabs);
 
+        open_windows (self, files,
+                      gdk_screen_get_default (),
+                      geometry,
+                      n_files,
+                      browser_window,
+                      open_in_tabs);
+
         /* Reset this or 3ed and later invocations will use same
          * geometry even if the user has resized open window */
         self->priv->geometry = NULL;
         g_strfreev (splitedOptions);
     }
-
-    open_windows (self, files,
-                  gdk_screen_get_default (),
-                  geometry,
-                  n_files,
-                  browser_window,
-                  open_in_tabs);
+    else
+        open_windows (self, files,
+                      gdk_screen_get_default (),
+                      geometry,
+                      n_files,
+                      browser_window,
+                      open_in_tabs);
 }
 
 void


lukefromdc:
To test: run caja --geometry=960x960 (or any other chosen geometry) from master and then from this PR

lukefromdc:
Just rebased for latest changes in master

sc0w:
I have tested this, and I see no difference apparently, wich is the issue in git master? here I see the same

lukefromdc:
My results were that caja --geometry=700x700 (or anything else) did not work in master, the geometry being ignored. I put in g_message statements to print the value of the geometry string, and it was getting corrupted after the 
`g_strfreev (splitedOptions);`
line, at least on Debian Unstable.

Appparently this is distro-specific, depending on exact version of glib or maybe some other underlying library, since it works for you and did not work for me in master. I have glib 2.60.0 built from the tarball and GTK 3.24.7(which should not change this) built from the git branch at that point. I first noticed this when putting Caja through its paces while checking another PR, and double-checked with master at that time. Problem also existed in older versions back (I think) into the 1.20 series when I tested them. 

 If there are any cases where it does not work it should be fixed, and in this case the fix is just to open the windows prior to freeing the string and resetting  self->priv->geometry to NULL. Since your underlying system is not corrupting the geometry string, just check for regressions.

sc0w:
ah, yes, confirmed the issue, caja always show the same size

src/caja-application.c
```diff
@@ -374,13 +374,20 @@ caja_application_open (GApplication *app,
...
         /* Reset this or 3ed and later invocations will use same
          * geometry even if the user has resized open window */
         self->priv->geometry = NULL;
         g_strfreev (splitedOptions);
     }
-
-    open_windows (self, files,
+    else
+        open_windows (self, files,
                   gdk_screen_get_default (),
```
sc0w: indentation

src/caja-application.c
```diff
@@ -374,13 +374,20 @@ caja_application_open (GApplication *app,
         }
         sscanf (splitedOptions[2], "%d", &open_in_tabs);
 
+        open_windows (self, files,
+                  gdk_screen_get_default (),
```
sc0w: indentation

sc0w:
seems this is the culprit commit: https://github.com/mate-desktop/caja/commit/04f7e80725c61ab511e2a4ead45eb31acc026e1a , reverted in https://github.com/mate-desktop/caja/commit/0d7a79dc7d4e1cef1a69005f58d0c29872c3d55c , and come back again squashed in https://github.com/mate-desktop/caja/commit/6a31ce65e4213d6069e5eb515640233cd27c3937 (PR related https://github.com/mate-desktop/caja/pull/1053 , closing feature request https://github.com/mate-desktop/caja/issues/1027 )

lukefromdc:
I can't see the problem with the indents unless you want the degree of indent of the lines passing variables to `open_windows` changed. Both blocks are supposed to be the same, one copied from the other, both indented 4 spaces to account for being in the if or else blocks. I didn't find any tabs in there. If you want the variable lines indented less I can easily change that and force-push

lukefromdc:
OK, we actually have three options here: this PR keeps the memory leak fix, working around it by executing` open_windows` prior to freeing `splitedOptions` , alternately we could of course remove that line. We can't free it after the original invocation of  ` open_windows` because it then is out of scope. We could of course declare `splitedOptions` at the top of the function (widening the scope) but that would set it unnecessarily when --geometry , --tabs, or --browser have not been passed

sc0w:
> I can't see the problem with the indents



> OK, we actually have three options here: this PR keeps the memory leak fix, working around it by executing open_windows prior to freeing splitedOptions , alternately we could of course remove that line. We can't free it after the original invocation of open_windows because it then is out of scope. We could of course declare splitedOptions at the top of the function (widening the scope) but that would set it unnecessarily when --geometry , --tabs, or --browser have not been passed

I am not sure, feel free to take the decision, for me this PR is ok.

lukefromdc:
That's how I had originally indented those functions, if you all want that changed to 4 spaces only I can do that

lukefromdc:
Rebased on latest git master, then amended the PR's commit to change the indents to 4 more spaces only for the variables in those two function blocks and force-pushed. Keeps this history clean.

sc0w:
The indentantion is still wrong, it seems you didn't understand me

See the changes before and later:

lukefromdc:
Still not sure what you want, but I will try aligning the variables in the lower lines with the variables in the first line. Since this is not a tab/space issue it is not obvious to me, and I have seen all kinds of different things done with this in our code.

I have to rebase anyway, so I will again change the indents in the commit, rebase, and force-push

lukefromdc:
Please recheck indents

PR #1200, opened 2019-04-03
Fix  #1198 "Context menu with menu key broken"

### LINKED ISSUE #1198: Context menu with menu key broken
#### Expected behaviour
Show context menu.

#### Actual behaviour
Nothing.

Press menu key on a selected item. It's only for the menu key, mouse right click shows the context menu. Menu key works otherwise, eg. in the browser.

#### MATE general version
1.22.0-1

#### Package version
1.22.0-1

#### Linux Distribution
Arch Linux x64

I'm not sure what to write here, but what about https://www.archlinux.org/packages/community/x86_64/caja/ ? :D

lukefromdc:
Pass "NULL" for event which is passed ultimately to gtk_menu_popup_at_pointer as an invalid event was being passed when the menu key was used to open the icon views context menu

### DIFF
--- a/src/file-manager/fm-directory-view.c
+++ b/src/file-manager/fm-directory-view.c
@@ -9187,9 +9187,13 @@ fm_directory_view_pop_up_selection_context_menu  (FMDirectoryView *view,
 
 	update_context_menu_position_from_event (view, event);
 
+	/* FIXME: passing event from here won't work
+	 * for gtk_menu_popup_at_pointer (in eel_pop_up_context_menu() )
+	 * if the menu is being triggered from here by the menu key
+	 */
 	eel_pop_up_context_menu (create_popup_menu
 	                         (view, FM_DIRECTORY_VIEW_POPUP_PATH_SELECTION),
-	                          event);
+	                          NULL);
 }
 
 /**
--- a/src/file-manager/fm-icon-view.c
+++ b/src/file-manager/fm-icon-view.c
@@ -2552,8 +2552,12 @@ icon_container_context_click_background_callback (CajaIconContainer *container,
     g_assert (CAJA_IS_ICON_CONTAINER (container));
     g_assert (FM_IS_ICON_VIEW (icon_view));
 
+    /* FIXME: passing event from here won't work
+     * for gtk_menu_popup_at_pointer (in eel_pop_up_context_menu() )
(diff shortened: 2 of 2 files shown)


lukefromdc:
I just found I have another piece of this to fix: selecting an item within the icon view with the mouse, then hitting the menu key still gets no menu and still gets 
`(caja:22948): Gtk-CRITICAL **: 14:48:42.830: gtk_menu_popup_at_rect: assertion 'GDK_IS_WINDOW (rect_window)' failed
`
So I need to find that function and pass NULL to eel_pop_up_context_menu in that case as well. We need to NOT pass NULL in the pathbar's case, but pass NULL in all of the icon view cases and I just found another one. Will work on that this evening, out of time at the moment. 

If you all want to merge this now I can do that in a second PR, if this is not merged this evening I will add it to this one. Either way works for me

raveit65:
I would prefer to see all fixes in this PR , for complete testing.
If someone needs it urgent, he can extract the commit from this branch.

lukefromdc:
Fine be me, should be able to have it all ready tonight

lukefromdc:
Fixed the issue with selecting an item in an icon view and popping up the context menu with the menu key, and rebased for latest changes in master. Right click menus for pathbar work fine. Using the menu key for locations in the pathbar  does not, but I'm not sure if it ever did, and problems there forced passing the event in the first place. To make that work would require something more complex than this. We don't get right click menus for toolbar items etc at all. 

At any rate, this should be ready to go, do check for any regressions. Only context menus for icon views and for places should be affected in any way.

raveit65:
Little rewording, i only moved the link to bugreport from commit header to commit body.
I hope you agree.
Now i start testing.....

lukefromdc:
fine by me

raveit65:
Not related to PR because i see that behaviour with 1.22.0 too.
The context-menu from pathbar-buttons doesn't open with keyboard.
Not sure if this was ever working.

sc0w:
@raveit65 

> Not sure if this was ever working.

https://github.com/mate-desktop/caja/issues/1198#issuecomment-479386002

culprit commit: https://github.com/mate-desktop/caja/commit/c0f159de3fc1fdce082acef36044016e3f7b3527

PR #1202, opened 2019-04-03
eel-string: Use 'memmove' instead of 'memcpy'

sc0w:
Fixes `flawfinder` warnings:

`(buffer) memcpy: Does not check for buffer overflows when copying to destination (CWE-120). Make sure destination can always hold the source data.`

http://cwe.mitre.org/data/definitions/120.html

### DIFF
--- a/eel/eel-string.c
+++ b/eel/eel-string.c
@@ -321,13 +321,13 @@ eel_str_replace_substring (const char *string,
         if (substring_position == NULL)
         {
             remaining_length = strlen (p);
-            memcpy (result_position, p, remaining_length);
+            memmove (result_position, p, remaining_length);
             result_position += remaining_length;
             break;
         }
-        memcpy (result_position, p, substring_position - p);
+        memmove (result_position, p, substring_position - p);
         result_position += substring_position - p;
-        memcpy (result_position, replacement, replacement_length);
+        memmove (result_position, replacement, replacement_length);
         result_position += replacement_length;
     }
 


cwendling:
I don't get the point of this.  `memmove()` does the same checking as `memcpy()`, the only difference is that with `memcpy()` the source and destination buffer cannot overlap (which is not the case here AFAICT).
Also, I didn't check super carefully, but it looks like the proper precautions are taken for copying a correct amount of data; and again, if it's not the case `memmove()` isn't gonna solve it.

This looks like a warning generated to make the developer look more closely at the code, because the tool couldn't make sure the sizes were correct. If anything, that's what we should do: review carefully that the pointer and sizes are always correct.

sc0w:
@cwendling 

I like to avoid security warnings with safer functions if possible, why not?

cwendling:
@sc0w my concern is that AFAICT this doesn't solve anything, as the *only* semantic difference between these two are whether source and destination buffers may overlap or not.
So yes, in this regard `memmove()` is safer as it allows buffers to overlap, but it's not the case here.  The tradeoff is that `memcpy()` might be faster, but again, bounds-wise it's exactly the same.

Look e.g. at https://embeddedartistry.com/blog/2017/3/7/implementing-memset-memcpy-and-memmove, or compare GCC's libiberty's [`memcpy()`](https://github.com/gcc-mirror/gcc/blob/master/libiberty/memcpy.c) and [`memmove()`](https://github.com/gcc-mirror/gcc/blob/master/libiberty/memmove.c) (hint: they are the same, they both simply call `bcopy()`).  If you compare some other libc's implementations you'll likely see a lot more difference, because some take advantage of the fact that `memcpy()` is not supposed to work on overlapping memory regions to optimize a bit further, but the idea is still the same: they copy `n` bytes from `src` to `dest`, no matter where either of those are pointing.

Anyway, my point is that using `memmove()` instead of `memcpy()` will *not* solve a buffer overflow issue.  Yes, it might solve issues related to overlapping buffers, but it's definitely not what the warning you cited is about.  So I'm just trying to point out that if this silences the warning, it ought to be eitehr because the warning was unclear and was supposed to refer to overlapping memory, or the checker simply fails to emit the same warning for `memmove()`.

---
PS:
An untested an terrible, yet should be conforming, implementation of `memmove()` could be:
```c
void *memmove(void *dest, const void *src, size_t n)
{
  if (((char *)src) + n < ((char *)dest) ||
      ((char *)dest) + n < ((char *)src)) {
    /* if regions don't overlap, we can use memcpy() */
    return memcpy(dest, src, n);
  } else {
    char temp[n];
    memcpy(temp, src, n);
    memcpy(dest, temp, n);
    return dest;
  }
}
```

sthibaul:
I agree with @cwendling : using memmove here does not make any sense: since result comes from a fresh malloc(), there is no way that it overlaps with p. What the warning is about is not about overlapping anyway, so memmove won't bring anything.

Here the issue is rather with flawfinder which does not understand that the code is already properly computing the target size before doing the malloc(). Using memmove won't fix this, at best it will hide the warning just because flawfinder does not *yet* print such warning for memmove calls.

raveit65:
Does not make any sense to merge it only make a tool happy.

raveit65:
@sc0w 
Can this be closed if it doesn't solve the real problem?

sc0w:
I can't demostrate problems using memcpy in this function

closing...

PR #1205, opened 2019-04-10
caja-icon-canvas-item: avoid 'cppcheck' warning

sc0w:
Fixes `cppcheck` warning:

`[libcaja-private/caja-icon-canvas-item.c:850]: (error) Uninitialized variable: real_text_height`

### DIFF
--- a/libcaja-private/caja-icon-canvas-item.c
+++ b/libcaja-private/caja-icon-canvas-item.c
@@ -830,21 +830,20 @@ compute_text_rectangle (const CajaIconCanvasItem *item,
         text_rectangle.y0 = icon_rectangle.y1;
         text_rectangle.x1 = text_rectangle.x0 + text_width;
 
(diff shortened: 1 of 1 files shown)


vkareh:
This looks a bit harmless, given the current code, but that safeguard is there to prevent programmer error. If `usage` is not initialized, or otherwise undefined, this function should fail. The code change now defaults it to `BOUNDS_USAGE_FOR_DISPLAY`, which changes the behavior of the function.

Maybe that's fine, or maybe we might want to add an assert at the top of the function so that `usage` _is_ defined.

vkareh:
oh wow, I just saw the new forced push just as I was clicking on "Comment" :stuck_out_tongue_closed_eyes: 

I think the new version looks good! I prefer the `switch` to a bunch of `if`s

sc0w:
@rbuj thanks, it worked too, and the warning is gone.

I think this is false positive warning, anyway, the code looks better with this change.

PR #1206, opened 2019-04-10
Revert "local #include files inside "" instead <>"

lukefromdc:
Reverts mate-desktop/caja#1204

I misread one of the comments reviewing  "local #include files inside "" instead <>" and mistook it for a statement that the change was correct, not that it should not be applied to libraries. We can revert this entirely, or if only some of the local #include files are libraries those and those alone could be put back into <>

### DIFF
--- a/eel/check-program.c
+++ b/eel/check-program.c
@@ -23,15 +23,15 @@
 */
 
 #include <config.h>
+
+#include <eel/eel-debug.h>
+#include <eel/eel-glib-extensions.h>
+#include <eel/eel-lib-self-check-functions.h>
+#include <eel/eel-self-checks.h>
 #include <gdk/gdk.h>
 #include <gtk/gtk.h>
 #include <libxml/parser.h>
 #include <stdlib.h>
-
-#include "eel-debug.h"
-#include "eel-glib-extensions.h"
-#include "eel-lib-self-check-functions.h"
-#include "eel-self-checks.h"
 
 int
 main (int argc, char *argv[])
--- a/eel/eel-accessibility.c
+++ b/eel/eel-accessibility.c
@@ -25,8 +25,7 @@
 #include <config.h>
 #include <gtk/gtk.h>
 #include <atk/atkrelationset.h>
-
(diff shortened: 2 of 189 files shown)


lukefromdc:
This is assuming eel is actually treated as a library of course! I don't see anything called "libeel" or such actually being installed with Caja(instead of being linked into the main executable at build time), so the question is this: Are these files really libraries or not?

sc0w:
Anyway, before https://github.com/mate-desktop/caja/pull/1204 some include files have "" instead <>

My intention was just to reorder the list of includes in all files to make it more readable.

I think if it works as before, there is no reason to revert.

rbuj:
A library is a program by itself, if you use relative include then you are exposing all objects.

sc0w:
> if you use relative include then you are exposing all objects.

I don't understand what do you mean

btw, we have this related commit merged in the past https://github.com/mate-desktop/caja/commit/242176e1d2b8748c47c43a776105285740d9ab30 by the PR https://github.com/mate-desktop/caja/pull/909

It was wrong?

rbuj:
`eel-string.c` and `eel-glib-extensions.h` belong to the same build scope (libeel-2), so it's correct to use `#include "eel-glib-extensions.h"` in `eel-string.c` file. See eel/Makefile.am file.

`caja-search-bar.h` and `caja-query-editor.h `belong to the same build scope (caja), so it's correct to use `#include "caja-search-bar.h"` in `caja-query-editor.h` file. See src/Makefile.am file.

`caja-window-bookmarks.h` and `caja-window.h `belong to the same build scope (caja), so it's correct to use `#include "caja-window.h"` in `caja-window-bookmarks.h` file. See src/Makefile.am file.

sc0w:
please don't merge, I am working in new PR to use <> instead "" when needed

lukefromdc:
OK, closing this in favor of upcoming correction PR

sc0w:
done!

@rbuj please review https://github.com/mate-desktop/caja/pull/1207

PR #1210, opened 2019-04-16
caja-file: 'g_strfreev' instead 'g_free' to free array of strings

sc0w:


sc0w:
`g_strfreev` handles `g_free` in every element of the array:

https://gitlab.gnome.org/GNOME/glib/blob/master/glib/gstrfuncs.c#L2494

```
void
g_strfreev (gchar **str_array)
{
  if (str_array)
    {
      int i;

      for (i = 0; str_array[i] != NULL; i++)
        g_free (str_array[i]);

      g_free (str_array);
    }
}
```

rbuj:
There is no new memory allocation in the elements of the **val** array:
https://github.com/mate-desktop/caja/blob/debc8612d23b5563e1a601bf59cbbe255d7f73c1/libcaja-private/caja-file.c#L3842

sc0w:
I see:

`val = g_new (char *, len + 1);`

And the documentation says:

https://developer.gnome.org/glib/unstable/glib-Memory-Allocation.html#g-new

```
g_new(struct_type, n_structs)

Allocates n_structs elements of type struct_type .
```

rbuj:
You can use `g_strfreev` only if you allocated new memory for the array elements, or you want to free elements data.

For instance `g_strdupv` allocates memory for the elements with `retval[i] = g_strdup (str_array[i]);`, and `retval = g_new (gchar*, i + 1);` is just an array of pointers.
https://gitlab.gnome.org/GNOME/glib/blob/master/glib/gstrfuncs.c#L2520

lukefromdc:
OK, we should probably close this than.

sc0w:
ok, no problem, closing...

PR #1211, opened 2019-04-16
desktop files: do not override caja preferences

rbuj:


raveit65:
Hmm, why removing --no-desktop ?
The desktop shouldn't start from a launcher.
Caja can be used by other desktops, eg. xfce in fedora.
https://src.fedoraproject.org/rpms/caja/blob/master/f/caja_add-xfce-to-desktop-file-1.22.patch

rbuj:
The users can acces to spatial mode using places menu from panel, and all desktop files but the caja-browser.desktop

PR #1212, opened 2019-04-16
C preprocessor: change to pragma once

rbuj:
with several advantages, including: less code, avoidance of
name clashes, and sometimes improvement in compilation speed.

https://en.wikipedia.org/wiki/Pragma_once

### DIFF
--- a/eel/eel-accessibility.h
+++ b/eel/eel-accessibility.h
@@ -21,8 +21,7 @@
    Authors: Anders Carlsson <andersca@gnu.org>
 */
 
-#ifndef EEL_ACCESSIBILITY_H
-#define EEL_ACCESSIBILITY_H
+#pragma once
 
 #include <glib-object.h>
 #include <atk/atkobject.h>
@@ -139,6 +138,3 @@ type_as_function ## _factory_get_type (void)					\
 	atk_registry_set_factory_type (atk_get_default_registry (),		\
 				       object_type,				\
 				       type_as_function ## _factory_get_type ())
-
-
-#endif /* EEL_ACCESSIBILITY_H */
--- a/eel/eel-art-extensions.h
+++ b/eel/eel-art-extensions.h
@@ -23,8 +23,7 @@
             Ramiro Estrugo <ramiro@eazel.com>
 */
 
-#ifndef EEL_ART_EXTENSIONS_H
-#define EEL_ART_EXTENSIONS_H
+#pragma once
 
 #include <glib.h>
 
@@ -116,5 +115,3 @@ extern "C" {
 #ifdef __cplusplus
 }
 #endif
-
-#endif /* EEL_ART_EXTENSIONS_H */
--- a/eel/eel-art-gtk-extensions.h
+++ b/eel/eel-art-gtk-extensions.h
@@ -32,8 +32,7 @@
  * to do operations on ArtIRects (such as intersection)
  */
 
-#ifndef EEL_ART_GTK_EXTENSIONS_H
-#define EEL_ART_GTK_EXTENSIONS_H
+#pragma once
 
 #include "eel-gtk-extensions.h"
 #include "eel-art-extensions.h"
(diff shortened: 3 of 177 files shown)


cwendling:
I would myself rather avoid altering working code to introduce non-standard directives, unless there is clear and measurable advantages and the possible caveats have been well though through.  The list from Wikipedia might be correct, but hardly a hard reference, and definitely not on this specific code.

I know this directive is supposed to be fairly widespread, but I fail to see the large benefit of changing every single header file in a working project to use it.  I myself *very* seldom have found problems related to include guards, clearly not enough in my experience to warrant changing working code.

Just my 2 ¢.

lukefromdc:
And I don't know fsck about this. The link suggested possible faster builds, but my suspicion is this sort of thing could be the source of problems on less common distros.

rbuj:
I saw this new approach in nautilus https://github.com/GNOME/nautilus/commit/95ecfc8ea0629801ae1f8119f3be0b3526e6b147

cwendling:
> And I don't know fsck about this. The link suggested possible faster builds, but my suspicion is this sort of thing could be the source of problems on less common distros.

Agreed.  Well, actually it *should* be fine so long it's built with GCC or CLang, but well.  And another question is quid about so analysis tools, are they happy about it?

> I saw this new approach in nautilus [GNOME/nautilus@95ecfc8](https://github.com/GNOME/nautilus/commit/95ecfc8ea0629801ae1f8119f3be0b3526e6b147)

Ah, actually that's an argument in favor of it, as Nautilus is some kind of an "upstream" in a way.
However, I wouldn't trust Nautilus so much when it comes to portability :)

> I have tested several times, and the build is faster with this change

Is it?  Interesting.  I wouldn't have expected it to be noticeable at all with all the build system overhead.  Out of curiosity, what is "faster" in this case?

rbuj:
TEST:
There are not duplicated symbols in both cases.
```shell
git clone https://github.com/mate-desktop/caja.git
cd caja
git checkout pragma-once

export CFLAGS="-O0"
./autogen.sh --prefix=/usr && time make
LANG=C objdump -T src/.libs/caja > ../dynamic-syms-caja-pragma-once
LANG=C objdump -t src/.libs/caja > ../syms-caja-pragma-once
LANG=C objdump -T libcaja-extension/.libs/libcaja-extension.so > ../dynamic-syms-libcaja-extension-pragma-once
LANG=C objdump -t libcaja-extension/.libs/libcaja-extension.so > ../syms-libcaja-extension-pragma-once

make distclean

git checkout `git log --pretty=%P -n 1`
export CFLAGS="-O0"
./autogen.sh --prefix=/usr && time make
LANG=C objdump -T src/.libs/caja > ../dynamic-syms-caja-include-guards
LANG=C objdump -t src/.libs/caja > ../syms-caja-include-guards
LANG=C objdump -T libcaja-extension/.libs/libcaja-extension.so > ../dynamic-syms-libcaja-extension-include-guards
LANG=C objdump -t libcaja-extension/.libs/libcaja-extension.so > ../syms-libcaja-extension-include-guards

diff ../dynamic-syms-caja-pragma-once ../dynamic-syms-caja-include-guards
diff ../dynamic-syms-libcaja-extension-pragma-once ../dynamic-syms-libcaja-extension-include-guards
diff ../syms-caja-pragma-once ../syms-caja-include-guards
diff ../syms-libcaja-extension-pragma-once ../syms-libcaja-extension-include-guards
```
Execution time: There is no significant difference.
Time pragma once:
```shell
real	3m7,242s
user	1m59,796s
sys	0m52,814s
```
Time include guards:
```shell
real	3m6,135s
user	1m58,041s
sys	0m52,357s
```
Lines of code:  192 (pragma once) vs. 192x3 (include guards)

raveit65:
I don't see that compiling is faster.

```
current master:

real	1m34,491s
user	1m17,426s
sys	0m19,163s

pragma once:

real	1m36,934s
user	1m19,338s
sys	0m19,783s
```

raveit65:
@mate-desktop/core-team 
Any other opinion?

yetist:
I am not sure if `pragma-once` has good portability. If want to improve the compilation speed, I think that porting to meson may be more effective.

PR #1215, opened 2019-04-30
eel: remove date & time functions

rbuj:
Do not need to extend glib library, since it now incorporates these functions.
Extended date format modifiers are available on [glib v2.56](https://bugzilla.gnome.org/show_bug.cgi?id=749206).
eel-glib-extensions.h/c:
```C
GDate * eel_g_date_new_tm (struct tm *time_pieces);
char * eel_strdup_strftime (const char *format, struct tm *time_pieces);
gint64 eel_get_system_time (void);
```

Do not need to test the extensions in `eel_self_check_glib_extensions (void)`.
eel-glib-extensions.c:
```C
static void check_tm_to_g_date (time_t time)
static char * test_strftime (...)
```

Do not need to check if strftime implements extended date format modifiers on the system (available on [glibc v2.27](https://sourceware.org/bugzilla/show_bug.cgi?id=10871)).
configure.ac

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -259,25 +259,6 @@ fi
 
 AC_SUBST(WARNING_CFLAGS)
 
-dnl ===========================================================================
-
-dnl strftime checks
-
-AC_TRY_RUN([#include <time.h>
-                int main ()
-                {
-                  char buf[100];
-                  struct tm tm = {0};
-                  tm.tm_year = 99;
-                  if (strftime(buf, 100, "%EY", &tm) == 4 &&
-                      strcmp (buf, "1999")==0)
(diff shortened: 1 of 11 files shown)


raveit65:
Why this make sense?
A description in a PR is always helpful.
I am sorry to say, but my crystal ball is broken :)

rbuj:
@raveit65 
done

raveit65:
@rbuj 
I meant to decribe what you you want to do in PR.
Maybe this helps that i understand why this is needed.
Well, you don't need to write a book but.....

rbuj:
@raveit65 
done

PR #1217, opened 2019-05-02
libcaja-private: reduce the scope of some variables

sc0w:
Fixes all the `cppcheck` scope variables warnings into the `libcaja-private` folder


raveit65:
What a massive commit. I reviewed it by hand ...
Looks like that none variable wasn't going over board :)
Now i start testing.

PR #1222, opened 2019-05-13
desktop icon view: disconnect callback from correct GSettings instance

monsta:
Taken from https://github.com/linuxmint/nemo/commit/89090ac192638d75359876d917b6d3ce106196af. It's a fix for a random crash which can be probably triggered by changing the desktop font while the desktop icon view is being deleted... I only saw a report about it on Launchpad ([here](https://bugs.launchpad.net/bugs/1598262)). In other words, if nobody can reproduce the crash, just check that things work as before :slightly_smiling_face:


raveit65:
@monsta 
Bugfix?
If yes, please cherry-pick to 1.22.

PR #1229, opened 2019-06-07
Add accessible names for buttons in file transfer operations dialog

TheMuso:


sc0w:
I have tested and I see no differences

@TheMuso can you provide screenshots?

TheMuso:
This change is for assistive technology users. YOu want to test with something like Orca, or examine the accessible object data with accerciser.

lukefromdc:
This is what I get now with caja master accerciser with the file transfer dialog. I assume we are looking for names displayed by those buttons etc? I have never, ever been able to get sound to work with orca, only save dubug output from it to a file.

lukefromdc:
Here's what I got with this PR:

Accerciser_Caja_w_PR

lukefromdc:
Best if someone more familiar than I with accessability reviews this too, but I will merge it if no response in a day or two

PR #1236, opened 2019-06-14
src: reduce the scope of some variables

sc0w:
Fixes all the `cppcheck` scope variables warnings into the `src` folder

Note: I will fix the scope variables in folder `/src/file-manager/` inside a separate PR


sc0w:
currently we have the same style in master at some points

lukefromdc:
OK, no sweat: consider this ready to go.

PR #1237, opened 2019-06-16
Remove trailing whitespaces

rbuj:
```shell
find . -regextype posix-extended -regex '.*\.(c|h|ac|txt|xml)' -exec sed --in-place 's/[[:space:]]\+$//' {} \+
```


rbuj:
The merge conflict is fixed now.

PR #1239, opened 2019-06-24
[notebooks]  Add scroll tabs with mouse wheel

tamplan:
Add scroll tabs with mouse wheel for file's properties and caja's preferences windows.

I plan to work on pane's notebook once this PR merges, i have seen gsettings key 'ctrl-tab-switch-tabs' without GUI that should be add in 'caja-file-management-properties.c'.

Regards

### DIFF
--- a/libcaja-private/Makefile.am
+++ b/libcaja-private/Makefile.am
@@ -184,6 +184,8 @@ libcaja_private_la_SOURCES = \
 	caja-window-slot-info.h \
 	caja-undostack-manager.c \
 	caja-undostack-manager.h \
+	caja-dialog-notebook.c \
+	caja-dialog-notebook.h \
 	$(NULL)
 
 nodist_libcaja_private_la_SOURCES =\
--- a/libcaja-private/caja-dialog-notebook.c
+++ b/libcaja-private/caja-dialog-notebook.c
@@ -1,0 +1,83 @@
+/* -*- Mode: C; indent-tabs-mode: t; c-basic-offset: 8; tab-width: 8 -*- */
+
+/* caja-dialog-notebook.h - GTKNotebook management for dialog windows
+
+   The Mate Library is free software; you can redistribute it and/or
+   modify it under the terms of the GNU Library General Public License as
+   published by the Free Software Foundation; either version 2 of the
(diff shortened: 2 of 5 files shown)


sc0w:
We have here the same function twice, with duplicate maintenance.

I think it can be done one time into one caja library and called with include.

tamplan:
Ok, let's go this way, i will change my code tomorrow...

tamplan:
Done, did you want i squash my commits?

Regards

tamplan:
I see than the build process failed because of marshaller generation but i don't know what file i must change...

If i understand [this page](https://developer.gnome.org/gobject/stable/glib-genmarshal.html), this is because the function 'dialog_page_scroll_event_cb' must be declared **extern**?

Can some one can confirm that point please?

sc0w:
I can't see the function you have done with previous changes, I think you can put that function in an already existing library, for example, into `eel-gtk-extensions.c` maybe?

tamplan:
Quickly summarized, i change 'libcaja-private/Makefile.am' to add 'caja-dialog-notebook.c' that contents the function 'dialog_page_scroll_event_cb' (callback that manage scroll events for notebooks) hoping this function could be called by 'caja-file-management-properties' and caja for file's properties page.

I change **static gboolean dialog_...** to **gboolean dialog_...** but i must, if i understand well, to declare if **extern gboolean dialog_...**.

I purpose to 'caja-dialog-notebook.c' alone file to not pollute another file.

I need to study more the caja's source code and C in general.

Thanks for your suggestion.

Regards

sc0w:
We have all the time needed :)

> Quickly summarized, i change 'libcaja-private/Makefile.am' to add 'caja-dialog-notebook.c' that contents the function 'dialog_page_scroll_event_cb' 

where is the file `caja-dialog-notebook.c` ?

tamplan:
Just used 'commit -a -m ...' but forgot to add tracking for 'caja-dialog-notebook.h' and 'caja-dialog-notebook.c' :-(

sc0w:
@mate-desktop/core-team do you agree to add new file with one unique function?

I think it can be added to `eel-gtk-extensions.c` for example

PR #1240, opened 2019-06-25
desktop: Always draw the items label frame

cwendling:
Draw the frame under the items label even when not selected and not in prelight state, so that it is possible to theme it in the other states. However we still don't draw it when editing, because there is the editing widget frame already and drawing the label's one might slightly overflow the editing widget's borders and lead to slightly ugly results.

Being able to draw a background is especially useful for accessibility purposes (e.g. a high contrast theme would typically make sure there is a background with enough contrast), but might be useful to all as desktop backgrounds are often not uniform at all.

By default this is drawn transparent on the non-selected and non-prelight states (no change in default appearance then), but allows the user to customize it with a background if need be.

Unfortunately for the moment this cannot be altered by theme, so the user will have to modify the GTK custom user CSS directly, but it still is better than nothing and can hopefully be solved better later.

---
The code change is basically reordering the branches and removing the `needs_highlight` from what was the first branch's test.

You can test the change by adding the following to *~/.config/gtk-3.0/gtk.css*:
```CSS
.caja-desktop .caja-canvas-item {
	background-color:white;
	color:black;
	text-shadow:none;
}

.caja-desktop .caja-canvas-item:selected {
	background-color:black;
	color:white;
}
```
and restart Caja.  You should then see a white background behind the desktop items labels.

### DIFF
--- a/data/caja-desktop.css
+++ b/data/caja-desktop.css
@@ -4,6 +4,7 @@
 .caja-desktop-window,
 .caja-desktop:not(:selected):not(:active):not(.rubberband){
 	background-color: transparent;
+	border: none;
 }
 
 /* desktop mode */
--- a/libcaja-private/caja-icon-canvas-item.c
+++ b/libcaja-private/caja-icon-canvas-item.c
@@ -1289,23 +1289,27 @@ draw_label_text (CajaIconCanvasItem *item,
                           "activate_prelight_icon_label", &prelight_label,
                           NULL);
 
-    /* if the icon is highlighted, do some set-up */
-    if (needs_highlight &&
-        !details->is_renaming) {
-            state |= GTK_STATE_FLAG_SELECTED;
-
-            frame_x = is_rtl_label_beside ? text_rect.x0 + item->details->text_dx : text_rect.x0;
-            frame_y = text_rect.y0;
-            frame_w = is_rtl_label_beside ? text_rect.x1 - text_rect.x0 - item->details->text_dx : text_rect.x1 - text_rect.x0;
-            frame_h = text_rect.y1 - text_rect.y0;
-    } else if (!needs_highlight && have_editable &&
-               details->text_width > 0 && details->text_height > 0 &&
-               prelight_label && item->details->is_prelit) {
+    /* if the icon needs a background, do some set-up */
+    if (!needs_highlight && have_editable &&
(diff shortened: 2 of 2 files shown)


cwendling:
The problem with theming is the same as the current situation for foreground color.  I asked on the GTK discussion board to see if they knew a way to fix this so a theme would provide the caja-specific bits, we'll see.

cwendling:
> I DID get a black frame border on the desktop around each item with this after removing that code and restarting Caja

Ah, this might be a problem.  We can add the `border:` property to `caja-desktop.css` so that it is overridden if the theme did set it (until we have a proper theming solution as mentioned above)

cwendling:
This said, you could have seen potentially the same issue before for selected items.

cwendling:
@lukefromdc I added a commit overriding the borders to avoid the issue you have been seeing with your theme.

raveit65:
There is one disadvantage when using external repos for PR.
Clang and cppcheck tests will only run on commits after the PR is merged. This is because of security reasons with depoying to http://caja.mate-desktop.dev/
When you use a branch from our repo for PRs than those test will run before PR is merged.
Here in this PR you see only one link to travis-Ci, but in this PR you see 3 links, for example.
https://github.com/mate-desktop/caja/pull/1244

raveit65:
@cwendling 
We have a collateral damage!
Split view (F3) with you PR.
caja-f3-view-new
Before you PR.
caja-f3-normal

This looks a bit weird with all light themes.

lukefromdc:
Confirmed on my UbuntuStudio_Legacy theme as well.

cwendling:
> We have a collateral damage! […]

ACK, I'll try and see what can be done here

cwendling:
> While this PR does NOT show it would be merged, note that your repo does not seem to include src/caja.about nor attempt to build it. I got a build failure on trying to just drop the changed files into an existing build directory (caja was too old I think) […]

Oops sorry, I didn't realize I PR'd with an older version of Caja.  Apparently I overlooked that my `git pull` failed on untracked files that would get overwritten (because some files that were previously generated are now checked in), my bad.

> There is one disadvantage when using external repos for PR.
> Clang and cppcheck tests will only run on commits after the PR is merged. This is because of security reasons with depoying to http://caja.mate-desktop.dev/

OK, I'll try and remember to push to branches on the main repo then.

PR #1244, opened 2019-06-28
hidpi: ensure all icons are properly scaled

lukefromdc:
In caja-icon-info.c use GTK_ICON_LOOKUP_FORCE_SIZE flag to ensure that icons actually scale when window-scaling-factor is set to 2 or more. Tested at window-scaling-factor=1 and window-scaling-factor=2

### DIFF
--- a/libcaja-private/caja-icon-info.c
+++ b/libcaja-private/caja-icon-info.c
@@ -410,14 +410,14 @@ caja_icon_info_lookup (GIcon *icon,
                                                                  icon,
                                                                  size,
                                                                  scale,
-                                                                 0);
+                                                                 GTK_ICON_LOOKUP_FORCE_SIZE);
 
         if (!gtkicon_info) {
             gtkicon_info = gtk_icon_theme_lookup_icon_for_scale (icon_theme,
                                                                  "text-x-generic",
                                                                  size,
                                                                  scale,
-                                                                 0);
+                                                                 GTK_ICON_LOOKUP_FORCE_SIZE);
         }
 
         icon_info = caja_icon_info_new_for_icon_info (gtkicon_info, scale);


raveit65:
At which place icons are affected?

lukefromdc:
They are affected in icon views both on the desktop and in navigation windows in and  only in certain icon themes (notably mine at
https://github.com/lukefromdc/gtk-theme-ubuntustudio-legacy ) not including any of the MATE themes, but if mine is affected others might be too.

It looks like any icon theme in which the type of use (e.g. apps, actions, mimetypes) is a higher directory level than the size of the icons AND the 48px icons are PNG and not SVG icons may be affected. That's the only difference I could find in my theme that could explain this, and it implies other themes may have this issue UNLESS no other icon theme is still organized that way.

At any rate, GNOME uses  GTK_ICON_LOOKUP_FORCE_SIZE in Nautilus, which does not have this issue with any icon theme. 

Caja_w_Icon_issue

lukefromdc:
Rebased after 1.23.1 release

lukefromdc:
This is no longer necessary in my icon theme after 
https://github.com/lukefromdc/ubuntustudio-legacy-icon-theme/commit/5982d926a33b548ec7b2f6135873b9d06f61016e
which opened the door to some other HiDPI fixes for sharpness and clarity as well. This might still help someone else with a broken theme, but testing now requires an old version of my theme (before July 1 2019 and that commit) or another broken theme. I had problems in index.theme that created the scaling issues. This it appears that normally we don't need  GTK_ICON_LOOKUP_FORCE_SIZE but it does make scaling more robust and less easily broken.

Do others want to merge this to deal with other broken themes, or should I close this?

raveit65:
It's Ok, let me test this.

PR #1248, opened 2019-06-30
src / file-manager: reduce the scope of some variables

sc0w:
Fixes all the `cppcheck` scope variables warnings into the `src / file-manager` folder

### DIFF
--- a/src/file-manager/fm-desktop-icon-view.c
+++ b/src/file-manager/fm-desktop-icon-view.c
@@ -159,7 +159,6 @@ net_workarea_changed (FMDesktopIconView *icon_view,
     int format_returned;
     int length_returned;
     CajaIconContainer *icon_container;
-    GdkScreen *screen;
 
     g_return_if_fail (FM_IS_DESKTOP_ICON_VIEW (icon_view));
 
@@ -225,6 +224,8 @@ net_workarea_changed (FMDesktopIconView *icon_view,
     }
     else
     {
+        GdkScreen *screen;
+
         screen = gdk_window_get_screen (window);
 
         icon_container_set_workarea (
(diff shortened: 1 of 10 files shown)


raveit65:
This is really a huge code change. Maybe it's better to split them in more single commits?
Which is better for `git bisect` when something goes wrong.
src/file-manager/fm-directory-view.c    | 198 ++++++++++++++----------
src/file-manager/fm-properties-window.c | 126 ++++++++++-----
src/file-manager/fm-list-model.c        |  88 +++++++----
src/file-manager/fm-list-view.c         |  68 +++++---
src/file-manager/fm-icon-view.c         |  63 +++++---

Beside from this it seems that none variable deceleration is lost and cppcheck tool warnings are silent.
But what is the technical advantage to move variables through the code?
Please don't let me die stupid ;)

sc0w:
only 10 files changed, I think it is unnecessary to split

in the past I did the same with one commit by folder:

1 commit in `src` folder with 32 changed files

1  commit in `libegg` folder with 3 changed files

1 commit in `libcaja-private` folder with 44 changed files

1 commit in `eel` folder with 15 changed files

To make consistent I think here is sensible to do the same, 1 commit again by folder

raveit65:
Not sure if you understood what i said.
I didn't mean how much files were changed in one commit.
I was talking about how much code lines were changed.
198 changed code lines from one file sounds enough for one commit, or not?

The problem with such huge commits is that `git bisect` shows one commit if you broke something.
And it is very hard see which code line is wrong.

Any way, this is only my personal opinion....

vkareh:
@raveit65 - re-adding you as a reviewer, since you seem to have some concerns with this. There shouldn't be any changes in behavior with this commit, so a `git bisect` wouldn't turn up anything anyway, but please review and test to be sure.

raveit65:
Change seems to be be OK, but for me this commit is too huge.
But i don't have to approve everything.
We are all responsible for the code and if something goes wrong with bisceting the code in future. i will say......i told you so ;)

PR #1249, opened 2019-07-01
Add support for checking .ui accessibility with gla11y

sthibaul:


.travis.yml
```diff
@@ -129,6 +131,7 @@ requires:
     - mate-common
     - mate-desktop-devel
     - pangox-compat-devel
+    - python-lxml
```
raveit65: I am really wondering that this will build on fedora,  see 
https://koji.fedoraproject.org/koji/buildinfo?buildID=1200268

python2-lxml-4.2.5-2.fc30.x86_64.rpm or  python3-lxml-4.2.5-2.fc30.x86_64.rpm

raveit65: Is using python2-lxml correct? See https://travis-ci.org/mate-desktop/caja/jobs/552817105#L373 ?
Python2 is deprecated and it will be removed soon from fedora.

sthibaul: Right, I have updated the branch to use python3 instead.

PR #1252, opened 2019-07-02
Restore transparent background on icon view labels

cwendling:
18e4fbd005bd561e4a12ed388d4d9a8c93f0b15b started always drawing the frame on canvas items for the desktop's sake, but it also impacts other icon views.  This is good in theory, but conflicts with some themes that used to work fine and now display an unwanted background because of generic rules matching these elements.

Unfortunately there don't seem to be another solution than providing overrides for this yet, so do that, as anyway before recently the background wasn't drawn at all even if specified.

Spotted in https://github.com/mate-desktop/caja/pull/1240#issuecomment-507039580

### DIFF
--- a/data/caja-desktop.css
+++ b/data/caja-desktop.css
@@ -2,7 +2,10 @@
 /* This is loaded with GTK_STYLE_PROVIDER_PRIORITY_APPLICATION and overrides themes */
 
 .caja-desktop-window,
-.caja-desktop:not(:selected):not(:active):not(.rubberband){
+.caja-desktop:not(:selected):not(:active):not(.rubberband),
+/* unfortunately we also have to apply this to all views normal states for the
+ * split view items not to have background */
+.caja-canvas-item:not(:selected):not(:active):not(.rubberband){
 	background-color: transparent;
 	border: none;
 }


raveit65:
> Unfortunately there don't seem to be another solution than providing overrides for this yet, so do that, as anyway before recently the background wasn't drawn at all even if specified.

Maybe a style class for actice an inactive view?
Wondering how this is solved in nautilus/nemo?

cwendling:
> Maybe a style class for active and inactive view?

(IIUC) The problem we have is that Caja's views have the `view` CSS class, and that get caught by the theme's rule for `.view`, regardless of us providing a custom rule with priority *theme* even if it's more precise.  So AFAICT adding a class won't help, maybe removing `.view` will, but that doesn't sound like a sensible move.

> Wondering how this is solved in nautilus/nemo?

It is a good question, but remember that before my patching, the items label background wasn't drawn *at all* but for the selected state.  Thus, the problem doesn't appear, but it's also not possible to draw any background even if wanted.

raveit65:
I think the `.view` class should be stay, you need that to set the background-color in themes, otherwise the background-color of the window is used.

I found out that the split view was dropped in nautius.
And nemo use `vertical.nemo-window-pane` and `vertical.nemo-window-pane.nemo-inactive-pane` to match the split view.

But such a specific style-class isn't a solution for Adwaita and HighContrast themes from gtk+.

PR #1253, opened 2019-07-04
file-conflict-dialog: don't scale icons twice on hidpi monitors

lukefromdc:
Hardcode scale factor to 1 in  caja_file_get_icon_pixbuf functions in caja-file-conflict as these icons are already getting scaled elsewhere. Otherwise they get scaled twice and the whole dialog is ballooned oversize on hidpi

*Based on similar change made in Nautilus as a small part of fd01842f996b7d07baf2aafb0f68308fc5ed2055 "file-conflict-dialog: separate file logic from UI management " which was a more general refactoring of the file conflict dialog and split out the frontend and backend code.

### DIFF
--- a/libcaja-private/caja-file-conflict-dialog.c
+++ b/libcaja-private/caja-file-conflict-dialog.c
@@ -76,7 +76,7 @@ file_icons_changed (CajaFile *file,
     pixbuf = caja_file_get_icon_pixbuf (fcd->details->destination,
                                         CAJA_ICON_SIZE_LARGE,
                                         TRUE,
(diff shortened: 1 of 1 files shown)


lukefromdc:
Fixes this problem: 1080p resolution on left, 4K resolution on right
Standard_and_Hidef

Before and after results, Nautilus gives similar results:
caja-file-conflict_dialog_results

raveit65:
Thanks for picking this up, i was still wondering why this icons are so huge with HIDPI.

PR #1254, opened 2019-07-04
Add support for theme-specific desktop CSS overrides

cwendling:
This allows to have a theme file to better fit desktop theming to a specific theme.  This is designed for themes with specific semantics, like High Contrast ones.

Support for HighContrast themes (both of the ones provided by GTK3, plus both current MATE's and an older of MATE's) is included, and can be used to test the feature: enable one of the High Contrast themes, and see the desktop items labels get drawn on plain background for optimal readability.

Adding support for a new theme is a simple matter of adding the rules in a file named `$caja_datadir/caja-desktop-$THEME.css`, it'll get automatically loaded when that theme `$THEME` is in use.

### DIFF
--- a/data/Makefile.am
+++ b/data/Makefile.am
@@ -39,6 +39,8 @@ cajadata_DATA = \
 	caja.css \
 	caja-desktop.css \
 	a11y-caja-desktop-base.css \
+	caja-desktop-ContrastHigh.css \
+	caja-desktop-ContrastHighInverse.css \
 	caja-desktop-HighContrast.css \
 	caja-desktop-HighContrastInverse.css \
 	$(NULL)
--- a/data/Makefile.am
+++ b/data/Makefile.am
@@ -38,6 +38,9 @@ cajadata_DATA = \
 	caja-suggested.placeholder \
 	caja.css \
 	caja-desktop.css \
+	a11y-caja-desktop-base.css \
+	caja-desktop-HighContrast.css \
+	caja-desktop-HighContrastInverse.css \
 	$(NULL)
 
 # app data file
--- a/data/a11y-caja-desktop-base.css
+++ b/data/a11y-caja-desktop-base.css
@@ -1,0 +1,18 @@
+/* base rules for highly contrasted desktop items */
+
+.caja-desktop.caja-canvas-item {
+  text-shadow: none;
+}
+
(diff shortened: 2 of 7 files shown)


raveit65:
This works, but i would use a light font with dark background for canvas items (desktop) with HC-Inverse, or not?
But it's your decision.

raveit65:
Forget what i said, my desktop didn't redraw correctly, after a restart of caja it works fine when i switch themes.
This happens sometimes with HIDPI.

alexarnaud:
@raveit65 @lukefromdc It seems this has been broken by another commit. The default theme for High Contrast was applied at the time of this commit but not in 1.24 from my testing on Debian 10.

raveit65:
@cwendling  ^^^^^

cwendling:
OK, this is not a change in Caja, but a change in GTK HighContrast theme that somehow breaks the snippets I had here.  I'll dig into that, but it's fine with GTK 3.24.14 (and should be good up to 3.24.16, breaking at 3.24.17 IIUC).

cwendling:
Looks like it is a fallout from https://gitlab.gnome.org/GNOME/gtk/-/issues/2641, which would be fixed if/when this hits GTK3.

raveit65:
Than someone needs to ask gtk+ to cherry-pick the fix to stable branch, or not?

raveit65:
I guess this fallout is also responsible for all other problems with HC theme that you posted in the last 24h.

cwendling:
> Than someone needs to ask gtk+ to cherry-pick the fix to stable branch, or not?

I did: https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/1730#note_777302

I think there's a fair chance it'll make it as I think I saw a wip branch for GTK3.  Let's wait and see.

> I guess this fallout is also responsible for all other problems with HC theme that you posted in the last 24h.

That's likely indeed.

cwendling:
I just tested https://gitlab.gnome.org/GNOME/gtk/-/tree/wip/jimmac/gtk3-public-hc-colors and it fixes the issue, so that was indeed the culprit.

cwendling:
It has been merged upstream: https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/1749
We just have to wait on a release now

PR #1255, opened 2019-07-05
Render file-conflict icons as cairo_surface vectors

vkareh:
This reverts the changes in #1253 and instead renders icons using `cairo_surface_t` vectors, rather than `GdkPixbuf`. The result is the correct size, correct scale icon shown in higher quality.

Before:
gdkpixbuf

After:
cairo_surface


vkareh:
Noticed that I forgot to change two instances of `g_object_unref` to `cairo_surface_destroy`. Fixed it just now

PR #1259, opened 2019-07-10
Fix not running gla11y when not available

sthibaul:


lukefromdc:
Also, someone needs to check this (all of these) with gla11y installed and make sure it works in that case

sthibaul:
CI did test that case

yetist:
I would suggest to modify the `gla11y.mk` file at the same time, which is more general.
```
diff --git a/gla11y.mk b/gla11y.mk
index abe9f53..5f7f805 100644
--- a/gla11y.mk
+++ b/gla11y.mk
@@ -10,7 +10,9 @@ a11y_verbose_1 =
 
 all-local: $(GLA11Y_OUTPUT)
 $(GLA11Y_OUTPUT): $(ui_files)
-       $(a11y_verbose) python3 $(GLA11Y) -P $(srcdir)/ -f $(srcdir)/$(GLA11Y_FALSE) -s $(srcdir)/$(GLA11Y_SUPPR) -o $@ $(ui_files:%=$(srcdir)/%)
+       $(a11y_verbose) if [[ $(GLA11Y) == *gla11y ]]; then \
+               python3 $(GLA11Y) -P $(srcdir)/ -f $(srcdir)/$(GLA11Y_FALSE) -s $(srcdir)/$(GLA11Y_SUPPR) -o $@ $(ui_files:%=$(srcdir)/%); \
+               fi
 
 clean-local: clean-local-check
 clean-local-check:
```

raveit65:
@sthibaul 
Sounds like good suggestion to update gla11y.mk here and in other open PRs, or not?
Than we can merge this.

vkareh:
I can confirm that this change works both with and without gla11y, but I agree with @yetist and @raveit65 - it makes sense to change gla11y.mk

raveit65:
@sthibaul 
We don't have all the time of the world!
Several main package doesn't build because of gla11y.
Please answer and add the suggestion.
Otherwise you force us to revert the culprit commit which causes the build issues.
Thanks.

raveit65:
@lukefromdc 
fixed by https://github.com/mate-desktop/caja/commit/3fb05765664fd045997a9827c3ba9598a8962a1c

lukefromdc:
I can confirm a new build with master after https://github.com/mate-desktop/caja/commit/3fb05765664fd045997a9827c3ba9598a8962a1c completed fine

sthibaul:
I was on holidays off-line last week, so couldn't even know that you wanted my feedback on the other approach, which eventually got committed. Before going offline I had assumed that the question was for the mate maintainers.

Anyway, AIUI it is fixed now.

cwendling:
> […]
> ```diff
> […]
> +       $(a11y_verbose) if [[ $(GLA11Y) == *gla11y ]]; then \
> […]
> ```

beware that `[[ … ]]` is NOT POSIX syntax, and is likely not portable.  It works with Bash, but does not with Dash, and probably many others.

lukefromdc:
With what is in current Caja master and gla11y not installed I just got a successful build in a dash terminal.

cwendling:
@lukefromdc yes apparently `if` always returns successfully even if the command it had to run wasn't found, so all you'd get is a bunch of `dash: [[: not found`.  However, this also means the body of the `if` won't ever be executed, so gla11y won't ever be executed with dash even if it's present.

sthibaul:
I have submitted https://github.com/mate-desktop/caja/pull/1271 which just avoids the issue

PR #1263, opened 2019-07-18
Support querying files by contained text

ptrNine:
ODF files also supported via odt2txt

### DIFF
--- a/libcaja-private/caja-query.c
+++ b/libcaja-private/caja-query.c
@@ -39,6 +39,7 @@ struct CajaQueryDetails
     GList *tags;
     gint64 timestamp;
     gint64 size;
+    char *contained_text;
 };
 
 static void  caja_query_class_init       (CajaQueryClass *class);
@@ -484,4 +485,15 @@ void caja_query_set_size(CajaQuery *query, gint64 size)
 gint64 caja_query_get_size(CajaQuery *query)
 {
     return query->details->size;
+}
+
+void caja_query_set_contained_text (CajaQuery *query, const char *text)
+{
+    g_free (query->details->contained_text);
+    query->details->contained_text = g_strdup (text);
+}
+
+char *caja_query_get_contained_text (CajaQuery *query)
+{
+    return g_strdup (query->details->contained_text);
 }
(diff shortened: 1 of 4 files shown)


ptrNine:
There was a mistake with mime-types. I used `g_content_type_equals` with `"text/plain"`. But content type of text files may be different (`"text/x-c"` for C files, `"text/x-diff"` for diff files, etc.)
Now used `g_content_type_is_mime_type` with `"text/plain"`. That matches all text files.
I had the same problem before that fix. Now all works fine.

lukefromdc:
Out of time today, will have to retest tomorrow at the earliest

raveit65:
Steps to reproduce the issue?
Steps to reproduce the PR?

lukefromdc:
This is a new feature: adding the ability to search files for contained text in Caja's search function as you can in mate-search-tool

lukefromdc:
I just looked over the memory management, matching up new dynamically allocated variables to g_free statements, we should be OK on that assuming I didn't miss anything.

raveit65:
I created a file `test` with the content `sex sells`
But search with `sex sells`, `sells` or `sex` doesn't find this file.
Please provide detailed `steps to test` if you want that i spend more of my spare time to review your PR.

lukefromdc:
The overall UI is a pain to use, the text string to use goes in the box labelled "matches files that contain specified text. The problem is the top  "search:" entry also exists and is very confusing. Ideally when using additional options that would be somehow cleared up

raveit65:
Ok, now it works.
I would really like to see some more explanation from author how to use it.
Than our user-guide can be updated by @mate-desktop/documentation-team

raveit65:
@ptrNine ^^^^^^^

ptrNine:
@raveit65 
For using search by contained text you should:

1. Enable search by clicking on the magnifier sign (or CTRL + F);
2. Run search for any text for enabling  'Search results' bar;
3. Add row by clicking '+' and select 'Contained text' in the combo box;
4. Enter search text in the label and click 'Reload'.

It works for all text files. 
ODF files (Texts, Presentations, Sheets) also supported if you have odt2txt on your system. odt2txt is a simple converter from ODF to plain text available on many repositories. Basically you should simply install odt2txt package for enabling ODF files search.

PR #1266, opened 2019-07-21
add git.mk to generate .gitignore

yetist:
See https://raw.github.com/behdad/git.mk


raveit65:
Correct link to source is https://raw.githubusercontent.com/behdad/git.mk/master/git.mk

PR #1271, opened 2019-07-24
gla11y.mk: fix compatibility with dash

sthibaul:


cwendling:
Why not simply prepend `python3` to `$(GLA11Y)` in configure?  Sounds to me like the thing to instead of fiddling with the command in the makefile.

```diff
 AC_PATH_PROG([GLA11Y], [gla11y], [true])
+AS_IF([test "$GLA11Y" != true], [GLA11Y="python3 $GLA11Y"])
```

sthibaul:
I don't think that `test "$GLA11Y" != true` will work, because it'll find true as /bin/true or something else.

cwendling:
> I don't think that `test "$GLA11Y" != true` will work, because it'll find true as /bin/true or something else.

It shouldn't, [AC_PATH_PROG](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/autoconf.html#index-AC_005fPATH_005fPROG-318) only searches in the `PATH` for `prog-to-check-for`, not for `value-if-not-found`, which can be anything.

cwendling:
Actually the whole forcing of python3 seems weird to me.  If needed in the builder, why not simply use a `export GLA11Y="python3 /path/to/gla11y"` there and remove the complexity here?  And if it's the will of the gla11y maintainer to drop Python2 support, why update the shebang there?

sthibaul:
Anyway, let's just stop bikeshedding, I have just released a v0.3 which defaults to  python3 already, so we don't have to care at all any more. I have updated the pull request accordingly

gla11y.mk
```diff
@@ -10,9 +10,7 @@ a11y_verbose_1 =
 
 all-local: $(GLA11Y_OUTPUT)
 $(GLA11Y_OUTPUT): $(ui_files)
-	$(a11y_verbose) if [[ $(GLA11Y) == *gla11y ]]; then \
-	        python3 $(GLA11Y) -P $(srcdir)/ -f $(srcdir)/$(GLA11Y_FALSE) -s $(srcdir)/$(GLA11Y_SUPPR) -o $@ $(ui_files:%=$(srcdir)/%); \
-	        fi
+	$(GLA11Y) -P $(srcdir)/ -f $(srcdir)/$(GLA11Y_FALSE) -s $(srcdir)/$(GLA11Y_SUPPR) -o $@ $(ui_files:%=$(srcdir)/%)
```
cwendling: you probably shouldn't remove `$(a11y_verbose)`

sthibaul: indeed, fixed

Issue #1280, opened 2019-08-08
Caja crashes when a file move in 19.2 mate


donlet:
#### Expected behaviour
Not crashed

#### Actual behaviour
Caja crashed when a file moved. The file has moved up
Copy and Delete are no problem.
 
Aug  8 20:43:46 tina kernel: [   25.937635] pool[1718]: segfault at 0 ip 00007f38a9339f98 sp 00007f38866fc800 error 4 in libglib-2.0.so.0.5600.4[7f38a92bc000+114000]
Aug  8 20:43:59 tina kernel: [   39.234093] pool[1867]: segfault at 0 ip 00007f34026a6f98 sp 00007f33f6147800 error 4 in libglib-2.0.so.0.5600.4[7f3402629000+114000]

#### Steps to reproduce the behaviour
Step 1. Pick a file with CTRL + X
Step 2. Change another directory
Step 3. CTRL + V

#### MATE general version
1.22.0

#### Package version
MATE caja 1.22.1

#### Linux Distribution
Linux mint 19.2

#### Link to downstream report of your Distribution
https://linuxmint.com/edition.php?id=261

I upgraded Linux mint 19.2 from 19.1

sc0w:
I can't reproduce this issue with git master (Aug 8, 2019)

lukefromdc:
Can this than be closed?

donlet:
Two PC at home and company are in the same problem.

donlet:
Workaround @.@

Step 1. Open multiTab A and B
Step 2. Pick a file of multiTab A with CTRL + X
Step 3. Change another directory of multiTab B
Step 4. CTRL + V

sc0w:
it works fine here, @donlet can you attach a video?

donlet:
https://youtu.be/YuchB9nmUb4

Sorry. I mistook the workaround.
If source and taget are different partition, Caja doesn't crash.
The source was in ext4 and the target was in ntfs.

donlet:
@sc0w Could you explain how to caja debug?

rbuj:
To enable Caja debug create ~/caja-debug-log.conf file and restart caja:
```shell
cat << EOF > ~/caja-debug-log.conf
[debug log]
max lines = 1000
enable domains = async;GLog
EOF

kill -s SIGUSR1 $(pgrep caja)
```

To see Caja log:
```shell
cat ~/caja-debug-log.txt
```

rbuj:
[...]

PR #1282, opened 2019-08-10
Update Korean translation which prevents a segfault on file move

rbuj:
See https://github.com/mate-desktop/caja/issues/1280

### DIFF
--- a/po/ko.po
+++ b/po/ko.po
@@ -4,21 +4,23 @@
 # FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
 # 
 # Translators:
-# Youngbin Han <sukso96100@gmail.com>, 2018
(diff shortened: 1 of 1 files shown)


raveit65:
This needs to be done at transifex to avoid that a pull from transifex will override this fix.

rbuj:
The translation was edited at [transifex](https://www.transifex.com/mate/MATE/translate/#ko/1-22--caja) first, then it was pulled using `tx pull -l ko -f`:
```shell
git clone git@github.com:mate-desktop/caja.git
cd caja/
git checkout 1.22
git checkout -b moving-files-in-Korean-122
tx pull -l ko -f
git add po/ko.po 
git commit -m "Update Korean translation which prevents a segfault on file move"
git commit --amend
git push origin moving-files-in-Korean-122
```

raveit65:
Ok, i didn't know that :)

PR #1283, opened 2019-08-10
Copy file and keep timestamps of original file including remotes

### LINKED ISSUE #1226: Copying to/from SSH server connection should keep timestamps
- Use Caja to connect to an SSH server.
- Copy files to/from the server.

#### Expected behaviour
The "Date Modified" timestamps of the files should stay unchanged.

#### Actual behaviour
The "Date Modified" timestamps are reset to the current time.

#### Versions:
MATE general version: 1.20.0
Package version: 1.20.2
Linux Distribution: Linux Mint 19.1 MATE 64bit

N/A

#### Remark
According to [gvfs](https://gitlab.gnome.org/GNOME/gvfs/issues/346#note_357921),
>  it is up to the client to use G_FILE_COPY_ALL_METADATA, or copy some attributes selectively over g_file_get_attribute() resp. g_file_set_attribute()  functions.

rbuj:
The timestamps are reset to the current time, even if the source &
target folders support these GFileInfo attributes.

Closes #1226

### DIFF
--- a/libcaja-private/caja-file-operations.c
+++ b/libcaja-private/caja-file-operations.c
@@ -3718,7 +3718,7 @@ copy_move_directory (CopyMoveJob *copy_job,
 
 	if (create_dest) {
 		flags = (readonly_source_fs) ? G_FILE_COPY_NOFOLLOW_SYMLINKS | G_FILE_COPY_TARGET_DEFAULT_PERMS
-					     : G_FILE_COPY_NOFOLLOW_SYMLINKS;
+					     : G_FILE_COPY_NOFOLLOW_SYMLINKS | G_FILE_COPY_ALL_METADATA;
 		/* Ignore errors here. Failure to copy metadata is not a hard error */
 		g_file_copy_attributes (src, *dest,
 					flags,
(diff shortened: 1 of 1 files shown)


lukefromdc:
How should I test this? I have not had issues with metadata that I have noticed given my use-case e.g I still have to strip it from jpegs after copying, and file creation times seem to be preserved.

rbuj:
@lukefromdc 

You can test file attributes with `stat` after copying them from caja, it's not about exif data.

vkareh:
@lukefromdc - I tested this by using caja to open a folder in an SSH connection. If you drag-drop a file from a local folder to the SSH folder on `master`, then right-click -> Properties, the Modified Date will show up as _now_, regarless of when the file was modified. With this patch, the file shows the correct modified date (which should be the same on both the local and the remote folder).

vkareh:
@rbuj, maybe add a better description in the commit message - the only indications that this is an issue with SSH is the branch name and the linked bug. Maybe the commit message should specify that

rbuj:
I only changed the title and the description of the commit.

lukefromdc:
You can test this with a loopback ssh mount (connect-to-server 127.0.0.1 and have ssh-server running), just confirmed the issue, about to test the fix

lukefromdc:
Note that the Ubuntu build failure in Travis is a timeout in installing dependencies, meaning it had nothing to do with the build at all. Are we ready to merge this?

PR #1291, opened 2019-08-19
Add --keyword=ngettext:1,2 to XGETTEXT_OPTIONS

rbuj:


raveit65:
For what is it good for?

rbuj:
It's not necessary to add this keywordspec, since it's already included for the C language by default:

The default keywordspecs for C, if not disabled (--keyword without keywordspec in XGETTEXT_OPTIONS):

gettext, dgettext:2, dcgettext:2, ngettext:1,2, dngettext:2,3,
dcngettext:2,3, gettext_noop, and pgettext:1c,2, dpgettext:2c,3,
dcpgettext:2c,3, npgettext:1c,2,3, dnpgettext:2c,3,4,
dcnpgettext:2c,3,4.

https://www.gnu.org/software/gettext/manual/gettext.html#Language-specific-options

PR #1295, opened 2019-08-21
Travis CI: 'autogen.sh' and 'scan-build' with '--enable-empty-view'

sc0w:
it is needed to build the files:
```
	fm-empty-view.c 
	fm-empty-view.h
```


raveit65:
Does anybody know what empty-view function is?

PR #1297, opened 2019-08-28
caja-window-slot: "folder-saved-search" icon instead "text-x-generic"

sc0w:
Before:

Later:

Before:

Later:

### DIFF
--- a/src/caja-window-slot.c
+++ b/src/caja-window-slot.c
@@ -394,7 +394,10 @@ caja_window_slot_update_icon (CajaWindowSlot *slot)
              */
             if (g_strcmp0 (icon_name, gtk_window_get_icon_name (GTK_WINDOW (window))) != 0)
             {
-                gtk_window_set_icon_name (GTK_WINDOW (window), icon_name);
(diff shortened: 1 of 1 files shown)


sc0w:
> A question concerning the code:
> This unconditionally replaces "text-x-generic" in Caja's window decoration titlebar. Are there any other instances that would bring up "text-x-generic" there and for which "folder-saved-search" would be inappropriate?

I don't know, this needs to test some time. Here all the other icons seems to be fine.

> Ideally this would be set by name for search windows directly, and probably the fallback icon should be "folder" which is the generic icon for a directory. If other cases are using the fallback that should be the fallback icon.

I don't like the "folder" icon in the search results, if we can put other icon, it is useful to improve accessibility.

I think this is the best icon:

lukefromdc:
I am only worried about folder-saved-search coming up in some other case where we are not in a search window or a saved search. Even that is probably better than text-x-generic, which is never appropriate for anything that opens in caja and instead is correct for Pluma. I will thus approve this as-is, just recommend checking for any such fallout.

PR #1303, opened 2019-09-19
caja-extensions: allows the translation of copyright

rbuj:


lukefromdc:
@raveit65 , I assume you have tested this in a non-English language?

raveit65:
@lukefromdc 
yes, i have tested it in my language. But i don't think we see a difference before new translations from transifex are available.

Issue #1284, opened 2019-08-10
Caja started to do hyphenation on long file names


piotrjurkiewicz:
Caja started to insert hyphens in long file names. See attached screenshot. It is probably related to the recent "enhancements" in Pango:

https://blogs.gnome.org/mclasen/2019/07/27/more-text-rendering-updates/

https://blogs.gnome.org/mclasen/2019/08/07/pango-1-44-wrap-up/

> As a consequence, there is now a text attribute to suppress the insertion of hyphens.

Maybe this attribute should be enabled in Caja?

![image](https://user-images.githubusercontent.com/2237894/62826218-eeb8ab80-bbb7-11e9-9474-06b139747b1c.png)

#### MATE general version

1.22

#### Package version

community/caja 1.22.1-1
extra/pango 1:1.44.3-1
extra/pangomm 2.42.0-1

#### Linux Distribution

Arch

baldr14:
Here is the Bug-Report for pango: https://gitlab.gnome.org/GNOME/pango/issues/401
The bug seems to be caja related. The bug still exists.


Distribution: Arch Linux
Desktop: gnome 3.34

Packages:
caja 1.22.2
pango 1.44.7
pangomm 2.42.0

raveit65:
Confirmed with fedora 31.
Ihmo, hyphens produce bad grammar ;)
Folder `AndroidStudioProjects` looks like this in caja with normal icon-size and HiDIPI resolution.

```
AndroidStudioProje-
cts
```
I am not a native English speaker but is this correct English grammar?

arteme:
As far as pango is concerned, it's not a bug, but a feature. Other projects had dealt with it:
https://gitlab.gnome.org/GNOME/nautilus/merge_requests/482/diffs

sc0w:
I think it is a crazy new feature with no sense

It was fixed in nautilus with https://gitlab.gnome.org/GNOME/nautilus/commit/9738d8561ed38a87b4d72db3e49653641edd9a09

raveit65:
> As far as pango is concerned, it's not a bug, but a feature. 

Who said this is a bug?

Breaking `AndroidStudioProjects` wrong is in grammar error.

raveit65:
@sc0w 
I agree with adding nautilus fix to caja.

baldr14:
> Who said this is a bug?

I said, but have correct it in my comment ;-)

raveit65:
[...]

PR #1320, opened 2019-11-15
canvas-item: Don't hyphenate filenames

raveit65:
fixes https://github.com/mate-desktop/caja/issues/1284

Pango 1.44 got the ability to automatically hyphenate on line breaks,
which is on by default, but can be set off by a new attribute.

As a result, we now hyphenate filenames, which is confusing, because
a filename may already include hyphens.

To restore the previous behavior, let's not insert hyphens when
breaking filenames in multiple lines.

Inspired by https://gitlab.gnome.org/GNOME/nautilus/commit/9738d85

Please test.

### DIFF
--- a/libcaja-private/caja-icon-canvas-item.c
+++ b/libcaja-private/caja-icon-canvas-item.c
@@ -60,6 +60,14 @@
 #define MAX_TEXT_WIDTH_BESIDE 90
 #define MAX_TEXT_WIDTH_BESIDE_TOP_TO_BOTTOM 150
 
+#ifndef PANGO_CHECK_VERSION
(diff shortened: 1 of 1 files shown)


lukefromdc:
We need a review from someone with pang 1.44 installed

raveit65:
@mate-desktop/core-team 
Any one like to test this bugfix before i back port it to 1.22 branch?

lukefromdc:
Any more reviews on this? This is a must-fix bug so we need either to merge this or if there is a problem to fix it and merge the results. No more tests and we have to merge this and deal with any problems as they are found. I don't have pango 1.44 though I suppose if this sits long enough Debian Unstable will get it eventually.

PR #1322, opened 2019-11-23
avoid redundant redeclarations

sc0w:
Fixes the warnings:

[caja-Wredundant-decls.txt](https://github.com/mate-desktop/caja/files/3882578/caja-Wredundant-decls.txt)


sc0w:
@lukefromdc 

> I assume these had to move to avoid issues from the removals?

The declarations must to be at first place, and `G_DEFINE_...` does the job.

`..._init` and `..._class_init` functions must to be later.

PR #1330, opened 2019-12-18
Fix a french grammar mistake in fr.po ("Copie" instead of "Copy")

Wolferos:
Fixed "Copie" instead of "Copy" in **fr.po** for _caja-file-operations_


raveit65:
We are using transifex for translations. https://www.transifex.com/mate/MATE/dashboard/
Please create an account become a member of your language team and fix it there.
Thanks

PR #1331, opened 2019-12-19
Updated French Localisation from Transifex

Wolferos:
As requested by @raveit65 https://github.com/mate-desktop/caja/pull/1330#issuecomment-567109522 I've joined the french translation team on _Transifex_.

I've modified small misspellings ("copie" instead of "copy") on Transifex and the French Loc from this PR is now synchronized with the last version from _Transifex_.

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -4,35 +4,38 @@
 # FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
 # 
 # Translators:
-# Louis-Martin Carrière <louismc@gmail.com>, 2018
-# Lionel Brianto <kinobi@kinobiweb.com>, 2018
-# Scoubidou <david.vantyghem@free.fr>, 2018
-# Benjamin Teissier <enatheme@enatheme.org>, 2018
-# jeremy shields <jeremy.shields@orange.fr>, 2018
-# yoplait <yoplait@tememe.org>, 2018
-# Xorg, 2018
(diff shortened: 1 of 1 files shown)


raveit65:
Thanks, but you only need to fix it at transifex (master and 1.22 branch).
Before a new package release we always do a pull from transifex to update .po files with latest changes from translators.
SO, no need to a pull request.

Wolferos:
Ok, but there is a very big problem with French Loc.

At the moment, _Caja_ crashes with default locale set on **fr_FR.UTF-8** when copying a folder containing files (no crash when copying an empty folder). This bug doesn't occurs with the default locale set on **en_GB.UTF-8** and it has already been reported in March 2019: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=922699

raveit65:
Ok, good to know.
I will do a pull from transifex for master and 1.22 branch
But debian use 1.20 which reached EOL status since nearly 2 years. 
So debian needs to use the commit as patch to fix it for 1.20.

Wolferos:
Ok, thank you

raveit65:
for master https://github.com/mate-desktop/caja/commit/13ae87df1f21a70d5e1fec93d91f70bffc653460
and 1.22
https://github.com/mate-desktop/caja/commit/46ddf950b8f962b8733290ee952385ae5bc0deec
Is this sufficient and we can close PR?

Wolferos:
Yes you can close it ;)

PR #1334, opened 2019-12-21
editable-label: Don't hyphenate filenames

raveit65:
more fixes for https://github.com/mate-desktop/caja/issues/1284

Renaming a file can make you crazy without this.
Please test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,6 +78,7 @@ requires:
     - libsm
     - mate-common
     - mate-desktop
+    - xorgproto
     - which
 
   debian:
--- a/eel/eel-editable-label.c
+++ b/eel/eel-editable-label.c
@@ -41,6 +41,14 @@
 #include <gdk/gdkx.h>
 #include <gdk/gdkkeysyms.h>
 
+#ifndef PANGO_CHECK_VERSION
+#define PANGO_CHECK_VERSION(major, minor, micro)                          \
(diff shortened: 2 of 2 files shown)


raveit65:
@yetist 
Looks like `xorgproto` isn't in default installation any more for ArchLinux.
I guess this will break building for most repos, or is this a package dependency issue by ArchLinux?
https://travis-ci.org/mate-desktop/caja/jobs/628108244#L878

raveit65:
> why the seemingly unrelated Travis change? Is it needed for evaluating the results of this change?

Yes.

> Was this caused by the first commit, since presumably builds on master completed?

No, first commit is fine in general, and is unrelated to archlinux build failure which happens on master too.

lukefromdc:
No sweat: if combining these is OK w you it is OK with me.

yetist:
Archlinux is doing the Xorg cleanup, so we need add `xorgproto` to the `.travis.yml` file.

https://www.archlinux.org/news/xorg-cleanup-requires-manual-intervention/
https://bugs.archlinux.org/task/64892

raveit65:
@yetist 
Seems that was only temporarily needed, build for 1.22 runs fine without adding `xorgproto` to `.travis.yml` file.
https://travis-ci.org/mate-desktop/caja/builds/628661277?utm_source=github_status&utm_medium=notification

PR #1340, opened 2020-01-09
Translate eject

johndescs:
Both eject device popups were not translatable. Make them translatable.

I also provide a translation into French, but maybe not the best way here since you are using an external tool, but it may still be used as an example…


raveit65:
Can you please remove french translation from PR.
We are using transifex for translations. https://www.transifex.com/mate/MATE/dashboard/
Every pull from transifex server will override translations at github.
In other words, become a member of your translation team and translate it there when your PR is merged.
Thanks

johndescs:
Okay, done.

PR #1342, opened 2020-01-16
Set a real title on the desktop window (as provided to ATK)

johndescs:
Even if we think that the desktop window title is never seen, it is used
by the window selector in "all windows" mode and speech synthesis.

### DIFF
--- a/src/caja-desktop-window.c
+++ b/src/caja-desktop-window.c
@@ -204,6 +204,8 @@ caja_desktop_window_new (CajaApplication *application,
         XFree(xch);
     }
 
+    gdk_window_set_title (gdkwin, _("Desktop"));
+
     g_signal_connect (window, "delete_event", G_CALLBACK (caja_desktop_window_delete_event), NULL);
 
     /* Point window at the desktop folder.
@@ -313,12 +315,6 @@ draw (GtkWidget *widget,
     return GTK_WIDGET_CLASS (caja_desktop_window_parent_class)->draw (widget, cr);
 }
 
-static char *
-real_get_title (CajaWindow *window)
-{
-    return g_strdup (_("Desktop"));
-}
-
 static CajaIconInfo *
 real_get_icon (CajaWindow *window,
                CajaWindowSlot *slot)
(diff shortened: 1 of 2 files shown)


rbuj:
`x-caja-desktop://` is not a standard URI. If you try to open this URI from a caja window, then the window title is set to `x-caja-desktop`. Are you trying to fix this behavior?

johndescs:
Not exactly, but the result of using this URI is that automatically it becomes the title of the window. But for the desktop window, this is not very user-friendly. By desktop window, I mean the one that's set as root, takes the whole screen behind everything, not a standard caja listing window with the content of the desktop.

alexarnaud:
It is useful for the control + alt + tab window to see "desktop" and hear "desktop" with Orca when moving between top bar, bottom bar and desktop.

rbuj:
a11y name is already set:
https://github.com/mate-desktop/caja/blob/410cfe6090513cfb93f799b1335765191a29aff6/src/caja-desktop-window.c#L132

https://github.com/GNOME/gtk/blob/3c50e5324f08e41dd9838334a631677684ac7ca1/gtk/a11y/gtkwindowaccessible.c#L132-L134

johndescs:
Yes, so it sounds not illogical to set it as "real/normal" window title too. The a11y name works well in mouse review mode for example, when the mouse is over the desktop we already get "Desktop" and not "x-caja-desktop".

rbuj:
let's see what the other team members think.

johndescs:
> I agree with you. From my point of view, you do no need to override parent's `get_title()` in class initializer because you set desktop window title using gdk_window_set_title.

Right, this can be dropped in the meantime as it will default to a meaningful title now.

raveit65:
No idea about, go ahead if it works for you without collateral damages.

cwendling:
> I agree with you. From my point of view, you do no need to override parent's `get_title()` in class initializer because you set desktop window title using gdk_window_set_title.

Good point, :+1: if it indeed works as it should.

johndescs:
I've updated the PR in accordance to all this and tested that it still works for this issue and in mouse review mode :)

cwendling:
> […] using gdk_window_set_title.

Wait, can't we simply use g**t**k_window_set_title() here?

johndescs:
Does it matter? I used gdk's just because we had already a ref to it available in the function. In a previous iteration with code added somewhere else I had one with gtk's but it was exactly the same outcome.

raveit65:
Any open questions or can this be merged?

alexarnaud:
From the user point of view it's OK for me.

@cwendling Is it OK for you ?

raveit65:
Anyone like to approve it for 1.24?

cwendling:
Sorry for the delay, I completely overlooked I had to rely here.  Yes, this is good :+1:

PR #1352, opened 2020-02-15
Update authors

rbuj:
```
$ git diff --word-diff=color --word-diff-regex=. master
```


raveit65:
Same problem here with no-reply addresses. First we like to find a solution here https://github.com/mate-desktop/engrampa/pull/366

rbuj:
@raveit65 done

raveit65:
Rebase to master and forced-push to fix travis builds.

Issue #1353, opened 2020-02-19
Pressing Ctrl+z while editing filename undoes previous copy operation


defterade:
#### Expected behaviour

The text is in the rename field is undone instead of the copy operation.

#### Actual behaviour

The copy operation is undone, removing the file while its name is being edited.

#### Steps to reproduce the behaviour

Copy a file using caja
Right click > Rename the file
Press Ctrl + z before pressing enter on the filename field

#### MATE general version

1.22.2

#### Package version

MATE caja 1.22.3

#### Linux Distribution

Arch Linux (5.5.4-arch1-1)

sc0w:
confirmed, one more related issue: https://github.com/mate-desktop/caja/issues/1106

lukefromdc:
DATA LOSS likely here:

This has an even more general case, because ctrl-Z is "undo" and undoes the last FINISHED operation without invoking any warning dialogs. Create a test file, rename it. Finish renaming it (click elsewhere), then hit ctrl-Z. The renaming operation is reversed, Repeat it, the file is deleted. Note that invoking "undo" by ANY means (ctrl-z or from the Edit menu) after creating/copying files/directories deletes them without invoking the "are you sure you want to permanently delete..." dialog. 

The default (or user set) preference "Ask before emptying trash or deleting files" is entirely ignored, and THAT is the true issue here. 

Use of ctrl-Z to invoke undo is a special case of it, as is the subcase of doing this while renaming a file and attempting to stop or undo it with ctrk-Z prior to editing the rename box. This can come as a complete suprise to a user.

This also applies to directories as stated above, and thus could also undo any mass copy job in the same run of caja unless the user has done something to break the undo/redo list. I just confirmed this: mass copied a bunch of .debs from their usual folder to a test folder on the desktop, started renaming one and hit ctrl-Z, all were deleted WITHOUT invoking the confirmation dialog. You could redo the copy, but NOT if you had since done anything else in Caja-and of course not if you no longer have access to the originals. 

 Note that while renaming a file it is easy on some keyboards to hit ctrl when you wanted shift, with potentially catastrophic results.

This could get seriously ugly if someone is renaming files they just copied from say, a friend's flash drive. Data loss results if they no longer have access to that drive, or have deleted the original file.

lukefromdc:
https://github.com/mate-desktop/caja/issues/1106 is the more general case, this issue is simply one way it can be invoked

lukefromdc:
Try
[...]

Issue #1106, opened 2019-01-09
we can loss data pressing CONTROL + Z accidentally


sc0w:
#### Expected behaviour

show dialog asking with every operations with `CONTROL + Z`

#### Actual behaviour

we can loss data pressing `CONTROL + Z `accidentally

#### Steps to reproduce the behaviour

- secondary mouse click -> create document -> empty file

- it creates an empty file with the name "new file"

- open the file "new file" (with pluma or other)

- add some text

- save the file

- go to caja and press CONTROL + Z (we can do it accidentally)

- the file is gone and we loss all of its contents

#### MATE general version

mate desktop 1.20.3

#### Package version

caja 1.20.3

#### Linux Distribution

debian testing/buster

lukefromdc:
`CONTROL + Z` is UNDO, and should respect the default (or user set) preference "Ask before emptying trash or deleting files" and bring up the warning dialog for deleting files unless the user has that turned off.

lukefromdc:
It looks like we have code that attempts to respect the delete confirmation here but does not actually work:

In caja-undostackmanager.c we have the function `caja_undostack_manager_undo`

containing this snippet of code:

```
        if (priv->confirm_delete) {
          caja_file_operations_delete (uris, NULL,
              undo_redo_done_delete_callback, action);
    	  g_list_free_full (uris, g_object_unref);
        } else {
          /* We skip the confirmation message
           */
          GList *f;
          for (f = uris; f != NULL; f = f->next) {
            char *name;
            name = g_file_get_uri (f->data);
            g_free (name);
            g_file_delete (f->data, NULL, NULL);
            g_object_unref (f->data);
          }
          g_list_free (uris);

```

and I found that replacing (for test purposes only) 
` if (priv->confirm_delete) {`
with
` if (priv) {` (which should always return TRUE), then we get the confirmation dialog for deleting files if and only if the user preference for asking before deleting files or emptying the trash is turned on (the default behavior). Somehow priv->confirm_delete is getting set to TRUE unconditionally. Since just using ` caja_file_operations_delete` seems to work fine and respect the user preference for the confirmation dialog, I will now write a PR for that

lukefromdc:
Try
https://github.com/mate-desktop/caja/pull/1354

cwendling:
[...]

PR #1354, opened 2020-02-22
Confirm undo if deleting

lukefromdc:
Fix https://github.com/mate-desktop/caja/issues/1106
Remove data loss threat from https://github.com/mate-desktop/caja/issues/1353 by preventing silent deletion of files in all cases except where user has turned off  "Ask before emptying trash or deleting files"

Use the general case  `caja_file_operations_delete` when undo operations involve deleting a file in all cases, as this respects "Ask before emptying trash or deleting files". Remove the direct use of `g_file_delete` and now unused variable `priv->confirm_delete` which was not being properly read or set for some reason. This is also shorter, simpler code and in the (default) case of " Ask before emptying trash or deleting files" being turned on is the intended codepath anyway.

Note that if the user cancels deletion of a file/files/directory/directories the job is still removed from the undo stack, the next previous job will now be next item that can be undone if "undo" is invoked again.

### DIFF
--- a/libcaja-private/caja-undostack-manager.c
+++ b/libcaja-private/caja-undostack-manager.c
@@ -265,7 +265,6 @@ caja_undostack_manager_init (CajaUndoStackManager * self)
   priv->index = 0;
   priv->dispose_has_run = FALSE;
   priv->undo_redo_flag = FALSE;
-  priv->confirm_delete = FALSE;
 }
 
 static void
@@ -319,9 +318,6 @@ caja_undostack_manager_set_property (GObject * object, guint prop_id,
         g_mutex_unlock (&priv->mutex);
         do_menu_update (manager);
       }
-      break;
-    case PROP_CONFIRM_DELETE:
-      priv->confirm_delete = g_value_get_boolean (value);
       break;
     default:
       G_OBJECT_WARN_INVALID_PROPERTY_ID (object, prop_id, pspec);
@@ -423,7 +419,7 @@ caja_undostack_manager_redo (CajaUndoStackManager * manager,
         uris = construct_gfile_list (action->sources, action->src_dir);
         caja_file_operations_copy (uris, NULL,
             action->dest_dir, NULL, undo_redo_done_transfer_callback, action);
-    	g_list_free_full (uris, g_object_unref);
+        g_list_free_full (uris, g_object_unref);
         break;
       }
       case CAJA_UNDOSTACK_CREATEFILEFROMTEMPLATE:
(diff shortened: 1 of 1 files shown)


sc0w:
typo in commit message: "Confirm**n**"

lukefromdc:
Commit message typo fixed

lukefromdc:
Had to squash the original two commits to get git to actually apply the commit message edit

sc0w:
the behavior is better than before, it fixes https://github.com/mate-desktop/caja/issues/1106 , but I think is better to ask always in every control+z operation, for example, if we rename a file, press enter, and press control+z accidentally, caja doesn't ask us. and it renames the file again with the previous name.

I think it doesn't fix https://github.com/mate-desktop/caja/issues/1353 completely, for example: we delete text in the name of file, press control+z, and caja does nothing with this PR, the expected is come back the deleted text.

lukefromdc:
It looks like part of https://github.com/mate-desktop/caja/issues/1353 is that while a rename operation is pending (user has not clicked elsewhere to close the entry and show the normal label), no part of the renaming operation is on the list of operations that can be undone.  Until that label is closed, Caja has not executed a rename operation, thus it can't be undone, is not on the list, and the PRIOR job (creation of the file in this case) is last on the list to undo. Thus we might want to edit the commit body to close only https://github.com/mate-desktop/caja/issues/1106

To change that might be rather complex, as it would require the user's individual keystrokes to be stored character by character on the list of operations that can be undone.

To ask for every undo operation would require creating a new dialog entirely and again is much more complex. Deletion of a file is unique in that if it is undone, it can't be redone unless it is a copy from a mount that still exists, and of a file that has not been deleted by some other program. Renaming a file by comparison can be undone and redone freely.  If someone wants to create a new dialog for this I have no trouble approving that, but we need to deal with the data loss potential now in my opinion.

lukefromdc:
Should we wait for another review or merge this? We have an incredible number of open PR's at the moment competing for everyone's time and this one relates to data loss

lukefromdc:
We have a Travis issue affecting only the Ubuntu build, does not mention anything related to this PR's file or code change:
https://travis-ci.org/mate-desktop/caja/jobs/657919561?utm_medium=notification&utm_source=github_status
```
Setting environment variables from .travis.yml

$ export DISTRO="ubuntu:19.10"

$ bash -c 'echo $BASH_VERSION'

4.3.48(1)-release

before_install.1

0.39s$ curl -Ls -o docker-build https://github.com/mate-desktop/mate-dev-scripts/raw/master/travis/docker-build
before_install.2

0.62s$ curl -Ls -o gen-index https://github.com/mate-desktop/mate-dev-scripts/raw/master/travis/gen-index.sh
before_install.3

0.00s$ chmod +x docker-build gen-index
install.1

4.61s$ sudo apt-get install -y python3-pip python3-setuptools
install.2

2.41s$ sudo pip3 install --upgrade pip
install.3

3.42s$ sudo pip install PyGithub

0.01s$ ./docker-build --name ${DISTRO} --config .build.yml --install

./docker-build: line 1: syntax error near unexpected token `newline'

./docker-build: line 1: `<!DOCTYPE html>'

The command "./docker-build --name ${DISTRO} --config .build.yml --install" failed and exited with 2 during .

Your build has been stopped.
```

libcaja-private/caja-undostack-manager.c
```diff
@@ -614,25 +610,11 @@ caja_undostack_manager_undo (CajaUndoStackManager * manager,
...
-            g_object_unref (f->data);
-          }
-          g_list_free (uris);
-          /* Here we must do what's necessary for the callback */
-          undo_redo_done_transfer_callback (NULL, action);
-        }
+
+        caja_file_operations_delete (uris, NULL,
+            undo_redo_done_delete_callback, action);
+    	g_list_free_full (uris, g_object_unref);
```
rbuj: don't mix spaces with tabs in the line above, remove trailing space in the line below.

lukefromdc:
Just fixed that tab/space issue. Most likely I had cut and pasted the tab from previous code containing the same line. This indent issue is so hard for me to see in pluma I often cut and paste the whitespace from ahead of lines trying to avoid creating problems, but sometimes (as here) that propagates a previous problem

libcaja-private/caja-undostack-manager.c
```diff
@@ -423,7 +419,7 @@ caja_undostack_manager_redo (CajaUndoStackManager * manager,
         uris = construct_gfile_list (action->sources, action->src_dir);
         caja_file_operations_copy (uris, NULL,
             action->dest_dir, NULL, undo_redo_done_transfer_callback, action);
-    	g_list_free_full (uris, g_object_unref);
+        g_list_free_full (uris, g_object_unref);
```
rbuj: don't mix spaces with tabs in the line above

raveit65:
There is a simple rule `no tabs = no pain = more fun at the beach` :P

lukefromdc:
Turns out we need https://github.com/mate-desktop/caja/pull/1357 to clean up a warning caused by a newly unused property

PR #1359, opened 2020-03-17
caja-file: remove warning GTimeVal is deprecated

rbuj:
Test: Move a file/folder to trash if empty, next open the trash and check that the Trashed on date is displayed correctly. It requires glib 2.61+ for tests.


raveit65:
The date on trashed file is displayed correctly.

PR #1363, opened 2020-04-03
use MATE_COMPILE_WARNINGS from mate-common

raveit65:


rbuj:
There are other components that would like this change: caja-extensions, mate-indicator-applet, mate-sensors-applet, mate-settings-daemon, mate-user-share

raveit65:
> There are other components that would like this change: caja-extensions, mate-indicator-applet, mate-settings-daemon, mate-user-share

Yes, we should do that, It's on my list.

PR #1370, opened 2020-04-08
Remove the warning: the address of ‘iter’ will never be NULL

rbuj:
Test:
Launch `caja-file-management-properties`, next select `Extensions` tab, finally enable/disable an extension.

Removed warning:
```
caja-file-management-properties.c:743:12: warning: the comparison will always evaluate as ‘true’ for the address of ‘iter’ will never be NULL [-Waddress]
  743 |  if (&iter != NULL)
      |            ^~
```


raveit65:
@monsta 
Why not merging with 2 approvals?

PR #1374, opened 2020-04-17
Travis CI: debian build: add "make check"

sc0w:


raveit65:
Beside that there is no description why this is needed and what `make check` does, this will increase build time.
I see that it will do a `make install`. But `make install/uninstall` is already tested by `make distcheck`.

Any way, can this be done only for distros at travis which do not other important tasks like releasing, gh-pages, etc. ?
I don't think that we need to see the result in every build.

sc0w:
This change is to avoid the issue mentioned in https://github.com/mate-desktop/caja/commit/312394dfa42c07a3afa5089c6e3e5b9a1c1001ae#commitcomment-34152266 in the future

With this change, without make install, it only works in fedora, I dont know why:
```
-  - make install
-  - if xvfb-run make check; then :; else cat eel/test-suite.log; cat src/test-suite.log; false; fi
   - make distcheck
+  - if xvfb-run make check; then :; else cat eel/test-suite.log; cat src/test-suite.log; false; fi
```
The builds: https://travis-ci.org/github/mate-desktop/caja/builds/676591603

anyway, I did new changes, now it works only in debian

raveit65:
DISTRO="archlinux:latest"    12 min 55 sec
DISTRO="debian:testing"     18 min 57 sec
DISTRO="fedora:latest"        14 min 43 sec
DISTRO="ubuntu:devel"       13 min 41 sec

PR #1375, opened 2020-04-23
Enable file selection on moved files and folders on samba shares

adam083b:
The call to the method caja_file_changes_queue_file_moved prevents the selection on moved files and folders on samba shares.
I'm not a gtk expert and don't understand clearly the cause. Replacing caja_file_changes_queue_file_moved by a call to caja_file_changes_queue_file_removed and caja_file_changes_queue_file_added works and should have the same behavior.

### DIFF
--- a/libcaja-private/caja-file-operations.c
+++ b/libcaja-private/caja-file-operations.c
@@ -4294,7 +4294,8 @@ copy_move_file (CopyMoveJob *copy_job,
 			g_hash_table_replace (debuting_files, g_object_ref (dest), GINT_TO_POINTER (TRUE));
 		}
 		if (copy_job->is_move) {
-			caja_file_changes_queue_file_moved (src, dest);
+			caja_file_changes_queue_file_removed (src);
+			caja_file_changes_queue_file_added (dest);
 		} else {
 			caja_file_changes_queue_file_added (dest);
 		}
@@ -4917,7 +4918,8 @@ move_file_prepare (CopyMoveJob *move_job,
 			g_hash_table_replace (debuting_files, g_object_ref (dest), GINT_TO_POINTER (TRUE));
 		}
 
-		caja_file_changes_queue_file_moved (src, dest);
+		caja_file_changes_queue_file_removed (src);
+		caja_file_changes_queue_file_added (dest);
 
 		if (position) {
 			caja_file_changes_queue_schedule_position_set (dest, *position, job->screen_num);


raveit65:
I am using only nfs shares.
@mate-desktop/core-team 
Does anyone use samba shares from  the team for testing?

lukefromdc:
I use neither one myself.

On 4/27/2020 at 5:35 AM, "raveit65"  wrote:
	I am using only nfs shares.
 @mate-desktop/core-team
 Does anyone use samba shares from  the team for testing?

echo083:
> The description is wrong, since it's not only applied to remote files. It goes without saying that your proposal has an overstrain on local file systems.

Hello,

You are right, my modification request also impacts the local file systems for move actions, but it's because after a move action is done, the method caja_file_changes_consume_changes is called which calls the methods caja_directory_notify_files_* depending of the elements in the caja files queue (CajaFileChangesQueue singleton) and triggers many signals. I'm afraid that if we don't apply the same logic for local file systems and remote file systems we would encounter unexpected side effects very difficult to understand. So my recommendation is to apply my proposal as it is or not at all.

Also I'm looking forward anybody to reproduce the bug and check that with my modification proposal fixes it. For information the fie I modified shares the same logic as this file in the nautilus project https://github.com/GNOME/nautilus/blob/mainline/src/nautilus-file-operations.c and I had the same bug but I didn't recompile it with my modification. Maybe I could request a review on their side too. What's your opinion?

Thanks

lukefromdc:
On the one hand, we cannot afford to make caja too heavy for old/small
systems to handle, or even worse import the bugs we have (freezes etc)
on remote mounts that bog down or disappeae to local filesystem
handling. If this requires limiting features on remote filesystem
management I'm in favor of so doing
On the other hand, Nautilus (from which caja was forked) was
originally intended right from the start to handle both local and
remote filesystwms

On 4/27/2020 at 9:28 AM, "echo083"  wrote:
	The description is wrong, since it's not only appliedo  to remote
files. It goes without saying that your proposal has an overstrain on
local file systems.  

	Hello, 

	You are right, my modification request also impacts the local file
systems for move actions, but it's because after a move action is
done, the method caja_file_changes_consume_changes is called which
calls the methods caja_directory_notify_files_* depending of the
elements in the caja files queue (CajaFileChangesQueue singleton) and
triggers many signals. I'm afraid that if we don't apply the same
logic for local file systems and remote file systems we would
encounter unexpected side effects very difficult to understand. So my
recommendation is to apply my proposal as it is or not at all. 

	Also I'm looking forward anybody to reproduce the bug and check that
with my modification proposal fixes it. For information the fie I
modified shares the same logic as this file in the nautilus project
https://github.com/GNOME/nautilus/blob/mainline/src/nautilus-file-operations.c
and I had the same bug but I didn't recompile it with my modification.
Maybe I could request a review on their side too. What's your opinion?


	Thanks

lukefromdc:
Any updates on this

echo083:
> Any updates on this

Hello,

Since 23.04 I'm using caja with my patch and didn't get any issue and I'm using it to move files on a daily basis si I'm quite confident of the reliability of my pull request. However a single proof of work is not sufficient in my opinion, except if you are voluntary to recompile caja and apply my patch and test it for a while on your own. You would require a samba share for information.

Thanks

raveit65:
> > Any updates on this
> 
> Hello,
> 
> Since 23.04 I'm using caja with my patch and didn't get any issue and I'm using it to move files on a daily basis si I'm quite confident of the reliability of my pull request. However a single proof of work is not sufficient in my opinion, except if you are voluntary to recompile caja and apply my patch and test it for a while on your own. You would require a samba share for information.
> 
> Thanks

Since April there is a change request.......

lukefromdc:
Any updates on this or should we close it?

echo083:
Hello, I'm still using my patch since April 2020 and had no issue. I requested a review, now proceed as you want if you know a developer or you are able yourself to review it and decide if it can be integrated to the main branch please go ahead otherwise you can close it.
Thanks

rbuj (changes requested):
I don't agree with removing the function **caja_file_changes_queue_file_moved** and the operation kind **CHANGE_FILE_MOVED**.
```
$ git grep caja_file_changes_queue_file_moved
libcaja-private/caja-file-changes-queue.c:caja_file_changes_queue_file_moved (GFile *from,
libcaja-private/caja-file-changes-queue.h:void caja_file_changes_queue_file_moved                      (GFile      *from,
```
```
$ git grep CHANGE_FILE_MOVED
libcaja-private/caja-file-changes-queue.c:    CHANGE_FILE_MOVED,
libcaja-private/caja-file-changes-queue.c:    new_item->kind = CHANGE_FILE_MOVED;
libcaja-private/caja-file-changes-queue.c:                            && change->kind != CHANGE_FILE_MOVED
libcaja-private/caja-file-changes-queue.c:                            && change->kind != CHANGE_FILE_MOVED;
libcaja-private/caja-file-changes-queue.c:        case CHANGE_FILE_MOVED:
```

echo083:
@rbuj If you don't agree you can dismiss my pull request! Thanks

rbuj:
@echo083 it's just my opinion, other team members may find your proposal useful or may point you to a better solution. Or you can find another solution that doesn't go through the indirect removal of an operation kind.

lukefromdc:
closing this as it makes caja too heavy

PR #1396, opened 2020-05-23
caja-window-slot: fix window icon replacement/alternation

### LINKED ISSUE #1394: Custom Caja icon in Windows List applet remains after first visit of a folder with Custom Icon selected
#### Expected behaviour
Caja icon in Windows List follows style of icon for current folder.

#### Actual behaviour
Caja icon in Windows List is default, but after visiting the folder with Custom Icon it remains custom forever.

1. Open Caja
2. Select some folder, do right mouse click on it, on Basic tab Select Custom Icon for it
3. Visit the folder with custom icon set
4. Using same Caja window visit Home folder with default icon

See screencast:

caja-icon

#### MATE general version
1.24.0

#### Package version
1.24.0-1

#### Linux Distribution
Ubuntu MATE 20.04 LTS

https://bugs.launchpad.net/ubuntu/+source/caja/+bug/1880281

rbuj:
Closes #1394

### DIFF
--- a/src/caja-window-slot.c
+++ b/src/caja-window-slot.c
@@ -384,34 +384,40 @@ caja_window_slot_update_icon (CajaWindowSlot *slot)
     icon_name = NULL;
     if (info)
     {
+        GtkWindow *gtk_window = GTK_WINDOW (window);
+
         icon_name = caja_icon_info_get_used_name (info);
         if (icon_name != NULL)
         {
+            /* frees the icon if it was previously set by using a pixbuf */
+            if (gtk_window_get_icon (gtk_window))
+                gtk_window_set_icon_list (gtk_window, NULL);
+
             /* Gtk+ doesn't short circuit this (yet), so avoid lots of work
              * if we're setting to the same icon. This happens a lot e.g. when
              * the trash directory changes due to the file count changing.
              */
-            if (g_strcmp0 (icon_name, gtk_window_get_icon_name (GTK_WINDOW (window))) != 0)
+            const gchar *current_icon_name;
+
+            current_icon_name = gtk_window_get_icon_name (gtk_window);
+            if ((current_icon_name == NULL) || (strcmp (icon_name, current_icon_name) != 0))
             {
-                if (g_strcmp0 (icon_name, "text-x-generic") == 0)
-                    gtk_window_set_icon_name (GTK_WINDOW (window), "folder-saved-search");
+                if (strcmp (icon_name, "text-x-generic") == 0)
+                    gtk_window_set_icon_name (gtk_window, "folder-saved-search");
                 else
-                    gtk_window_set_icon_name (GTK_WINDOW (window), icon_name);
+                    gtk_window_set_icon_name (gtk_window, icon_name);
             }
         }
         else
         {
-            GdkPixbuf *pixbuf;
-
-            pixbuf = caja_icon_info_get_pixbuf_nodefault (info);
+            GdkPixbuf *pixbuf = caja_icon_info_get_pixbuf_nodefault (info);
 
             if (pixbuf)
             {
-                gtk_window_set_icon (GTK_WINDOW (window), pixbuf);
+                gtk_window_set_icon (gtk_window, pixbuf);
                 g_object_unref (pixbuf);
             }
         }
-
         g_object_unref (info);
     }
 }


raveit65:
Beside that i can restore the original folder icon with the revert button in icon-ui-dialog, what is the different to master?

src/caja-window-slot.c
```diff
@@ -384,34 +384,43 @@ caja_window_slot_update_icon (CajaWindowSlot *slot)
...
         {
             /* Gtk+ doesn't short circuit this (yet), so avoid lots of work
              * if we're setting to the same icon. This happens a lot e.g. when
              * the trash directory changes due to the file count changing.
              */
-            if (g_strcmp0 (icon_name, gtk_window_get_icon_name (GTK_WINDOW (window))) != 0)
+            const gchar *current_icon_name;
+
+            current_icon_name = gtk_window_get_icon_name (gtk_window);
+            if ((current_icon_name == NULL) || (strcmp (icon_name, current_icon_name) != 0))
```
sc0w, line 406: why `strcmp` instead `g_strcmp0` ?

I noticed this commit in the past: https://github.com/mate-desktop/caja/commit/ebf820f112dd840b73183c693a832001f53b7bae

maybe `g_strcmp0` is better?

src/caja-window-slot.c
```diff
@@ -384,34 +384,43 @@ caja_window_slot_update_icon (CajaWindowSlot *slot)
...
             {
-                if (g_strcmp0 (icon_name, "text-x-generic") == 0)
-                    gtk_window_set_icon_name (GTK_WINDOW (window), "folder-saved-search");
+                if (strcmp (icon_name, "text-x-generic") == 0)
```
sc0w, line 408: the same here

rbuj: `icon_name` can't be NULL... take a look at the line 397 `if (icon_name != NULL)`

src/caja-window-slot.c
rbuj: strcmp (icon_name, current_icon_name:

- `icon_name` can't be NULL... take a look at the line 397 if (icon_name != NULL)
- `current_icon_name` can't be NULL...   take a look at the line 406 `(current_icon_name == NULL)`

rbuj:
> `strcmp` can't handle `NULL`, please use `g_strcmp0` instead

It's not required to hande NULL values here.

sc0w:
for security reasons, if we use `strcmp` with `NULL` it will  cause unexpected behavior, probably it will crash, for this reason is better `g_strcmp0`

`current_icon_name = gtk_window_get_icon_name (gtk_window);` <- it can be unexpectedly `NULL`

You are doing an additional change unneeded to fix the issue, I don't understand why you don't keep it as before

sc0w:
oops, you are right, it is checking before if it is `NULL`

src/caja-window-slot.c
```diff
@@ -384,34 +384,43 @@ caja_window_slot_update_icon (CajaWindowSlot *slot)
...
     if (info)
     {
+        GtkWindow *gtk_window;
+        GdkPixbuf *pixbuf;
+
+        gtk_window = GTK_WINDOW (window);
+
+        pixbuf = gtk_window_get_icon (gtk_window);
+        if (pixbuf)
+            gtk_window_set_icon (gtk_window, NULL);
```
sc0w, line 394: little question, we need `g_object_unref (pixbuf);` inside this condition?

rbuj: I added a comment in the code. You don't need to call g_object_unref, calling gtk_window_set_icon or gtk_window_set_icon_list frees the icon/list when the new value is NULL.

sc0w:
New warning:

```
caja-window-slot.c:394:43: warning: incompatible pointer types passing 'CajaWindow *' (aka 'struct CajaWindow *') to parameter of type 'GtkWindow *' (aka 'struct _GtkWindow *') [-Wincompatible-pointer-types]
                gtk_window_set_icon_list (window, NULL);
                                          ^~~~~~
/usr/include/gtk-3.0/gtk/gtkwindow.h:297:65: note: passing argument to parameter 'window' here
void       gtk_window_set_icon_list                (GtkWindow  *window,
                                                                ^
```

rbuj:
done

PR #1407, opened 2020-06-06
caja properties-window: display dir last modified

### LINKED ISSUE #1406: Directory properties does not display last modified date
https://gitlab.gnome.org/GNOME/nautilus/-/merge_requests/409

#### Expected behaviour
Display last modified date on directories

#### Actual behaviour
Last modified date not displayed in UI properties

 - Select folder, open properties

wesinator:
closes #1406 

based on https://gitlab.gnome.org/GNOME/nautilus/-/merge_requests/409/diffs

### DIFF
--- a/src/file-manager/fm-properties-window.c
+++ b/src/file-manager/fm-properties-window.c
@@ -2652,11 +2652,19 @@ should_show_accessed_date (FMPropertiesWindow *window)
 	 * day decide that it is useful, we should separately
 	 * consider whether it's useful for "trash:".
 	 */
-	if (file_list_all_directories (window->details->target_files)) {
+	if (file_list_all_directories (window->details->target_files)
+	    || is_multi_file_window (window))
+ 	{
 		return FALSE;
 	}
 
 	return TRUE;
+}
+
+static gboolean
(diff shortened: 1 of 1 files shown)


rbuj:
`Modified` should not be displayed on a mount point:

Before:

Screenshot at 2020-06-30 14-41-47

After:

Screenshot at 2020-06-30 14-56-49

raveit65:
Oh, i didn't check that.
But why it is bad?

@wesinator 
Any idea to exclude that for mount points?

wesinator:
> Oh, i didn't check that.
> But why it is bad?
> 
> @wesinator
> Any idea to exclude that for mount points?

No - I just adopted the fix from the upstream nautilus change 

I don't think it's really a problem 
It seems from the screenshot the mount point is identified as a directory in the file manager, hence why it shows modified date.

rbuj:
PR #1424

rbuj:
From my point of view, it's a date that makes no sense for mount points and it can lead to confusion, since it displays the date on which the volume was mounted, which has nothing to do with the date on which the volume was last modified.

Issue #1412, opened 2020-06-14
fm-properties doesn't show all tabs when opening on folders


raveit65:
#### Expected behaviour

Fm-properties should show all tabs, we have only 4.
#### Actual behaviour
![Screenshot at 2020-06-14 14-28-16](https://user-images.githubusercontent.com/961604/84595307-76fe8180-ae57-11ea-9ccb-ecbf34e91719.png)
Only 3 are shown and we see unneeded GtkNotebook arrows
Issue is caused by https://github.com/mate-desktop/caja/commit/0b4c7a6b8c25afb987d08bfd2c4c9be57de23960
Fix coming soon.
#### Steps to reproduce the behaviour
Open fm-properties from context menu of folders

#### MATE general version
1.24

#### Package version
1.24.0

#### Linux Distribution
fedora

#### Link to downstream report of your Distribution
I don't file out report against myself.

lukefromdc:
I just found that I never saw this because the tabs on my gtk theme are too small to use up all the space. Had to narrow the dialog and switch to Menta to see it, but yes this is now confirmed here too. About to test the PR

raveit65:
Well, the meaning of `gtk_notebook_set_scrollable`is a bit misleading.
https://developer.gnome.org/gtk3/stable/GtkNotebook.html#gtk-notebook-set-scrollable
```
Sets whether the tab label area will have arrows for scrolling if there are too many tabs to fit in the area.
```

PR #1413, opened 2020-06-14
fm-properties: Don't show notebook arrows

raveit65:
fixes https://github.com/mate-desktop/caja/issues/1412

Unwanted arrows won't be shown and notebook-tabs mouse-scrolling works like before.


tamplan:
Sorry for that.

Regards.

PR #1415, opened 2020-06-20
[1.20] Reset move/copy time before beginning operation

### LINKED ISSUE #1132: Transfer rate calculated wrong on enqueued operations
#### Expected behaviour
In dialog "File Operations" the transfer speed should be calculated only while the file is actually transferred.

#### Actual behaviour
When I enqueue multiple big files to copy via network on a not-so-fast connection, it seems that the transfer speed is calculated over the whole time that the file is in the queue. So the time the transfer is paused, seems also to be taken into account for the transfer speed.

I had files of several 100 MB (between 100 and 800 MB) and copied them to a networked (scp) destination. In the attached screenshot you see "leftovers" that were enqueued a lot earlier and paused and started automatically by Caja. The true transfer rate of the link was about 1.5 MB/sec. This speed is only displayed on the initial transfer that brings up that dialog for the first time. The displayed transfer rates of later files always increase, but never reach the true rate, that you can measure with the begin and end time of the transfer and the filesize.

#### MATE general version
1.20.3

#### Package version
caja 1.20.2-5

#### Linux Distribution
Ubuntu Release 18.10 x86_x64

n/a

jnwatts:
Tested on v1.20.3-13-g01b11be, can apply on latest as well but I'm unable to test.

Closes #1132

### DIFF
--- a/libcaja-private/caja-file-operations.c
+++ b/libcaja-private/caja-file-operations.c
@@ -4196,6 +4196,10 @@ copy_move_file (CopyMoveJob *copy_job,
 		goto out;
 	}
 
+	/* Timer was already running from g_timer_new()...
+	 * Restart now that we are actually moving or copying
+	 */
+	g_timer_start( job->time );
 
  retry:
 	error = NULL;


raveit65:
Thanks,
we only accept PRs on master, and 1.20 is EOL since 1.5 years. 1.22 is EOL since January this year
So we will never do new releases for 1.20 or 1.22.
We support only one stable release because we are a small team.
Of course  we will cherry-pick into 1.24 branch when this is a bug fix.
But you can ask your distro use this commit when is is merged as patch for older Mate releases in their distro.

Sorry.....

jnwatts:
No worries! Thanks for the quick response anyway :-)

I already filed a PR against master a little while ago: #1416

PR #1416, opened 2020-06-20
Reset move/copy time before beginning operation

### LINKED ISSUE #1132: Transfer rate calculated wrong on enqueued operations

jnwatts:
Queued transfers start their timer upon being queued, which negatively
affects transfer-speed calculation by including the time spent *not
transfering*.

Closes #1132
(Mirror of #1415 for master, untested)

### DIFF
--- a/libcaja-private/caja-file-operations.c
+++ b/libcaja-private/caja-file-operations.c
@@ -4237,6 +4237,10 @@ copy_move_file (CopyMoveJob *copy_job,
 		goto out;
 	}
 
+	/* Timer was already running from g_timer_new()...
+	 * Restart now that we are actually moving or copying
+	 */
+	g_timer_start (job->time);
 
  retry:
 	error = NULL;


raveit65:
Not related to  PR, but start/pause/stop logic seems quite broken in master.

libcaja-private/caja-file-operations.c
```diff
@@ -4237,6 +4237,10 @@ copy_move_file (CopyMoveJob *copy_job,
 		goto out;
 	}
 
+	/* Timer was already running from g_timer_new()...
+	 * Restart now that we are actually moving or copying
+	 */
+	g_timer_start( job->time );
```
rbuj, line 4243: Fix Code Formatting:
```patch
diff --git a/libcaja-private/caja-file-operations.c b/libcaja-private/caja-file-operations.c
index 8079cd9..8462d22 100644
--- a/libcaja-private/caja-file-operations.c
+++ b/libcaja-private/caja-file-operations.c
@@ -4240,7 +4240,7 @@ copy_move_file (CopyMoveJob *copy_job,
        /* Timer was already running from g_timer_new()...
         * Restart now that we are actually moving or copying
         */
-       g_timer_start( job->time );
+       g_timer_start (job->time);
 
  retry:
        error = NULL;
```

jnwatts: Whoops! Not sure where my head was, usually I'm better at picking up near-by style :-)

Fixed in 4ec34f7

rbuj:
There is no need to reset the timer before copying or moving each of the files, take a look at delete_files operation, where it can be seen that the timer is reset before the loop:
https://github.com/mate-desktop/caja/blob/4ec34f73c36f0655eb084e60a2e8c8cdee401747/libcaja-private/caja-file-operations.c#L1792-L1832

jnwatts:
> There is no need to reset the timer before copying or moving each of the files, take a look at delete_files operation, where it can be seen that the timer is reset before the loop:
> https://github.com/mate-desktop/caja/blob/4ec34f73c36f0655eb084e60a2e8c8cdee401747/libcaja-private/caja-file-operations.c#L1792-L1832

With a brief look, appears right, but I'm unable to test (doesn't

PR #1423, opened 2020-06-30
caja-file-management-properties: Reduce the height of the dialog

### LINKED ISSUE #1381: Preference is too big
#### Expected behaviour
More compact preference dialog(or more tabs and less options in one tab).

#### Actual behaviour
The preference dialog is so big that I can't get the "OK" button.
2020-04-29 15-57-32 创建的截图

1. Change the environment language to Chinese 
2. Open Caja
3. Open 编辑 -> 首选项

#### MATE general version
1.24.0

#### Package version
1.24.0

#### Linux Distribution
Ubuntu MATE 20.04 LTS

Thanks!

rbuj:
Untitled

Untitled-2

resolves #1381

### DIFF
--- a/src/caja-file-management-properties.ui
+++ b/src/caja-file-management-properties.ui
@@ -538,67 +538,265 @@
                   </packing>
                 </child>
                 <child>
-                  <object class="GtkBox" id="vbox3">
+                  <object class="GtkFrame">
                     <property name="visible">True</property>
                     <property name="can_focus">False</property>
-                    <property name="orientation">vertical</property>
-                    <property name="spacing">6</property>
+                    <property name="hexpand">True</property>
+                    <property name="vexpand">True</property>
+                    <property name="label_xalign">0</property>
+                    <property name="shadow_type">none</property>
                     <child>
-                      <object class="GtkLabel" id="label5">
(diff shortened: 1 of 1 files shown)


raveit65:
Should adding support for `on_view_defaults_notebook_scroll_event` not be done in a second commit?
Commit message doesn't describe this and is misleading.
And you screenshot looks like you did reduce the width and not the height, or not?

rbuj:
@raveit65 ok, we can do that later.

rbuj:
It's a visual effect of the theme, only the height is changed.

The Chinese screenshot is narrower than English screenshot because Chinese uses logograms instead of characters.

raveit65:
> ok, we can do that later.

Then let us first review the other open PRs.

> It's a visual effect of the theme, only the height is changed.

> The Chinese screenshot is narrower than English screenshot because Chinese uses logograms instead of characters.

Ok, good.

raveit65:
I consider this as bugfix.

raveit65:
I think we can't cherry-pick this to 1.24 branch because of https://github.com/mate-desktop/caja/commit/270534f373b8fd1abceae02049b7aa21a2dca334

PR #1429, opened 2020-07-06
fm-icon-view: totem-audio-preview was removed in totem 3.25.90

rbuj:


lukefromdc:
have working sound previews after building today's master (with this merged) and restarting caja, no problem

PR #1435, opened 2020-07-25
eel-editable-label: Consecutive break/return statements are unnecessary

rbuj:
reported by cppcheck


raveit65 (approved):
Builds and runs fine, warnings are gone.

Issue #1401, opened 2020-05-28
Saving web links via drag and drop from Firefox is buggy


vjayer:
#### Expected behaviour
*.desktop files are created containing link url and expected meta data.

Note: I am not sure if an additional starting interpreter line of
`#!/usr/bin/env xdg-open`
should also be required in said *.desktop files.
Currently this does **not** occur from drag-n-dropping into the Caja file manger, when it is successful. It does occur for *.desktop files created from drag-n-dropping into the MATE panel.

#### Actual behaviour
1) Sometimes *.desktop files are created -- however see Note above about differences between *.desktop files created by Caja and by the MATE panel

2) Sometimes local html copies are created instead

3) Sometimes an error occurs:
`HTTP Error: Method Not Allowed`

#### Steps to reproduce the behaviour
Normally behaviors 1-2 above occurs randomly. If I get a local copy saved, which usually takes a moment as it actually scrapes the web page, I'll just try again and most of the time it will save a link (behavior 1) in the next try, or 2nd or 3rd.

But I have found web pages were I can finally reproduce all 3 behaviors above consistently:
1) https://letstalkbitcoin.com/blog/post/understanding-urbit-1-introducing-the-personal-server-1
- no problem, seems to always save *.desktop links so far

2) https://letstalkbitcoin.com/profile/user/legendface66
- **problem**, always saves local html copy

3) https://www.wiley.com/en-us/Discord+For+Dummies-p-9781119688099
- **problem**, always results in
`HTTP Error: Method Not Allowed`

#### MATE general version
1.22.2-1+tricia

#### Package version
1.22.2-1+tricia

#### Linux Distribution
Linux Mint Mate 19.3

#### Link to downstream report of your Distribution
https://blog.linuxmint.com/?p=3833

I created a thread about this here:
https://forums.linuxmint.com/viewtopic.php?f=206&t=320347

I have also upgraded from Linux Mint Mate 18.x if that makes a difference

there is also a related bug:
https://github.com/mate-desktop/caja/issues/1323
but in his case, it does not create links--nothing happens at all.

vjayer:
There is one more thing I forgot: A user in the linuxmint forum mentioned what turned out to be a workaround, and might provide insight to a solution:

[...]

PR #1436, opened 2020-07-26
caja-dnd: Create links by default on dnd from Web Browsers

### LINKED ISSUE #1433: Caja inconsistent creation of link-type launchers by dragging from browsers
Hopefully some pictures are worth a thousand words...

Dragging a URL from Chrome or Firefox to **mate-panel** creates launchers...
bitmap3

Dragging a URL from Chrome or Firefox to **pluma** pastes the text of the URL...
bitmap2

Dragging a URL from Firefox to **caja** **DOES** create launchers, but from Chrome it **DOES NOT**...
bitmap

This issue has been around for years on numerous releases.  I mostly **only** use LTS releases of Ubuntu MATE so that'd mean I've seen this on MATE 1.12, 1.20 and more recently 1.24.

I can also confirm that this problem exists if dragging into a normal **caja** window (i.e. not the Desktop)... it just reacts as if you're not dragging anything at all (i.e. no little question mark appears next to your mouse cursor, like it does when dragging from Firefox).

There is a work around, which is to drag them to **mate-panel**, then drag that launcher to **caja**, then finally remove the launcher from **mate-panel**.  Not only is this time consuming and messy... but also the launchers created by **mate-panel** seem to be inconsistent with those created by **caja** (e.g. if using something that works like Firefox).  The launcher made by **mate-panel** seem to lack the 'Link to' in the name, use a different icon and other slightly different properties.

Any thoughts?

This is a really useful feature if you're trying to archive off some unfinished work (and half-read tabs) into a folder to pick up a project later.  Naturally on Linux this means freedesktop ".desktop" launcher files... where as on a Mac you'd have ".webloc" files and on Windows you'd have ".url" files (nothing a little bash script can't handle and pass to xdg-open)... but the same deal ultimately... it'd be nice to restore this functionality to regain some feature parity with other OS's.

**UPDATE**

I have just noticed that this doesn't even behave consistently with Firefox either.  Depending on the website (or perhaps contents of the URL or title of the page maybe?) Caja will sometimes instead try to copy the web page instead!  This has been witnessed when dragging from Konqueror too.  However in both cases, dragging to a **mate-panel** is still fine.

Give this URL a go in Firefox to see what I mean...   [https://www.linuxfoundation.org/](https://www.linuxfoundation.org/)

rbuj:
closes #1433

### DIFF
--- a/libcaja-private/caja-dnd.c
+++ b/libcaja-private/caja-dnd.c
@@ -349,20 +349,28 @@ caja_drag_items_on_desktop (const GList *selection_list)
 GdkDragAction
 caja_drag_default_drop_action_for_netscape_url (GdkDragContext *context)
 {
-    /* Mozilla defaults to copy, but unless thats the
-       only allowed thing (enforced by ctrl) we want to ASK */
-    if (gdk_drag_context_get_suggested_action (context) == GDK_ACTION_COPY &&
-            gdk_drag_context_get_actions (context) != GDK_ACTION_COPY)
+    /* Mozilla defaults to copy, however by default caja creates a link
+     * if available.
+     *
+     * Enforced action X:
+     * ((actions & X) != 0) && (suggested_action == X)
+     *    - GDK_ACTION_LINK enforced by Alt
+     *    - GDK_ACTION_MOVE enforced by Shift
+     *    - GDK_ACTION_COPY enforced by Ctrl
+     */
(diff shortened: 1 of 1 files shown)


vjayer:
hello I have some test urls cited here:
https://github.com/mate-desktop/caja/issues/1401
does this commit also fix those?

PR #1437, opened 2020-07-27
Update Desktop Entry keys for created links to http/https/ftp urls

rbuj:
See https://github.com/mate-desktop/caja/issues/1433#issuecomment-664245428


rbuj:
It really isn’t necessary.

PR #1461, opened 2020-11-09
Fix i18n in backgrounds and emblems dialog

rbuj:
Test: Select "Edit -> Background and emblems" from menu bar.

Before

Captura de pantalla a 2020-11-09 19-01-24

After

Captura de pantalla a 2020-11-09 18-51-17


raveit65 (approved):
LGTM,
and bring back the translation.
Simple change and one step forward not to use `eel` functions.

cwendling (approved):
LGTM. The XML seems to now be in English rather than containing the translations themselves (since 83fa9ede5c267fdd7ddca3f70801b838ee348190), and this PR fixes translating those strings (WFM in French).

PR #1463, opened 2020-11-12
fm-properties-window: make the location of system folders portable

rbuj:


lukefromdc (approved):
This builds and installs fine, icons and pixmaps go to their normal places in my builds (for packing into Debian packages which use the postinst script to compile the schemas) with 
`./autogen.sh --prefix=/usr --libdir=/usr/lib/x86_64-linux-gnu  --sysconfdir=/etc --enable-shared=yes --enable-static=no --disable-maintainer-mode  --disable-schemas-compile`

raveit65 (approved):
Works fine.

PR #1467, opened 2020-11-22
caja-icon-container: warning -Wold-style-declaration

rbuj:
```
caja-icon-container.c:842:1: warning: 'inline' is not at beginning of declaration [-Wold-style-declaration]
  842 | static void inline
      | ^~~~~~
```

### DIFF
--- a/libcaja-private/caja-icon-container.c
+++ b/libcaja-private/caja-icon-container.c
@@ -839,7 +839,7 @@ clear_keyboard_focus (CajaIconContainer *container)
     container->details->keyboard_focus = NULL;
 }
 
-static void inline
+static inline void
 emit_atk_focus_state_change (CajaIcon *icon, gboolean focused)
 {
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
caja-icon container built with no warnings after this, with my normal build options. Build finished, nothing broke. No runtime suprises, none were expected but I checked anyway.

raveit65:
I found some links where this warning is fixed like this:
```
-static void inline
+static inline void
```
https://oss.oracle.com/pipermail/ocfs2-devel/2011-January/007790.html
https://github.com/Evervolv/android_kernel_htc_qsd8k/commit/9f1253577230eb23b75b149d35c27c9e86b0161e

rbuj:
yes it makes sense

rbuj:
@raveit65 done

raveit65 (approved):
Builds and runs fine, and build warning is gone.

Issue #1466, opened 2020-11-21
[Suggest] Add emblems in the suspicious .desktop files


HacKurx:
Hello to all the developers of Caja,

While reading this [article](https://null-byte.wonderhowto.com/how-to/pop-reverse-shell-with-video-file-by-exploiting-popular-linux-file-managers-0196078/), I think it would be a good idea to impose a danger emblem to all *.desktop files if they don't come from a system-approved location such as /usr/share/.

For more understanding :
```
~$ cd ~
~$ ls -lah *.desktop
~$ -rwxrwxr-x 1 loic loic 338 nov.  21 13:36 fake_video.desktop
~$ -rwxrwxr-x 1 loic loic 369 nov.  21 13:37 fake_video_emblems.desktop
```

![Capture du 2020-11-21 13-38-13](https://user-images.githubusercontent.com/4661917/99877708-61f4fe00-2c00-11eb-8a98-62957a71d386.png)

Thanks.

Regards,

artebin:
Hello,

I agree, removing the `.desktop` extension and not indicating that the file is a .desktop file following the  [Desktop Entry Specification](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html) is one of the bad UI/UX designs that we can find in the today desktop metaphors (one among plenty... ).

It would be more correct to hide the `.desktop` extension **if and only if** the file follows the Desktop Entry Specification.
All other cases: show the real file name (i.e. complete file name, i.e. show the extension). 
And if the file is indeed a valid `.desktop` file then always add emblem to the icon.

It is also critical to have an option in the Preferences of Caja: `Always show complete file name in file explorer window`.
Hiding the `.desktop` extension might be useful for the icons in the desktop panel, but in the _file explorer_ it is a non-sense to me to not show the complete file name.

rbuj:
The desktop files can be used in the user space, for example to create links to local files...

HacKurx:
@rbuj have you read my link?

rbuj:
No I don't. For security, I don't usually read untrusted external links.

HacKurx:
> No I don't. For security, I don't usually read untrusted external links.

![Screenshot_2020-11-23 How to Pop a Reverse Shell with a Video File by Exploiting Popular Linux File Managers](https://user-images.githubusercontent.com/4661917/99996455-0ec7aa80-2dbc-11eb-8b66-3743ab8f146b.png)

rbuj:
[...]

PR #1468, opened 2020-11-23
caja-directory-async: trustred launchers

rbuj:
See #1466


lukefromdc:
Is this intended to let users manually mark a .desktop file as trusted after examination or not?

rbuj:
I haven't decided this part yet, but we could store a new gvfs metadata (.local/share/gvfs-metadata) such as trusted-launcher="yes"

lukefromdc:
I am opposed to force-disabling user-created .desktop files with no option to override. In fact, I use them myself

rbuj:
@lukefromdc done

HacKurx:
For me, only the .desktop files coming from /usr/share/applications/ are trusted (because it belongs to root).
The goal is not to prevent the user from creating *.desktop files just distinguish them visually.

There are many different attacks with *.desktop files. Here is another one:

Imagine that Bob forgets to log out. Someone passes and changes the firefox shortcut in mate-panel or on the desktop to replace the legitimate command with a malicious one.
Bob comes back and clicks it because for him it's still firefox (same icon, no visual warning that the shortcut has changed).

Ideally, the file '/home/Bob/Desktop/firefox.desktop' should look the same only if it is identical (sha256) to the file '/usr/share/applications/firefox.desktop'. If the file '/home/Bob/Desktop/firefox.desktop' is modified or created by the user then it should have a visual emblem imposed.

rbuj:
The alert emblem can also be replaced with a transparent image by changing the icon theme, which may be located in user home.

HacKurx:
So for the more cautious among us, we need a new option in caja :

@artebin
> It is also critical to have an option in the Preferences of Caja: Always show complete file name in file explorer window.

Thanks.

artebin:
Or maybe the other way around: an option FALSE by default named "Hide .desktop extensions in Desktop folder only"

lukefromdc (commented):
Here's what I got on a test build and install (which went fine, and no surprises elsewhere at runtime):

Existing user-made desktop file on desktop no change in behavior
Moved that file to another directory, and the "launch anyway/mark as trusted/cancel dialog came up and worked as expected. 

NEW (made in same run of caja) user-made desktop file on desktop no change in behavior

Moved that file to another directory, and the "launch anyway/mark as trusted/cancel dialog comes up, but marking as trusted did not work until caja is restarted. When caja was restarted, the launcher was now trusted but not before. Selecting "launch anyway" always worked as expected

This was verified by testing it twice, after caja restarts

lukefromdc:
Does anyone else want to review this (it is a behavior change) before merging it? I forgot to re-select coreteam for another review

raveit65:
LiveCDs are coming with desktop files like `Install-Live-CD-to-disc.`
So this PR will mark them as un-trusted which isn't true.
I am not sure if we really should do this.

rbuj:
> LiveCDs are coming with desktop files like `Install-Live-CD-to-disc.`
> So this PR will mark them as un-trusted which isn't true.

They are marked as trusted in `caja-directory-async: trust also launchers from user desktop`.

lukefromdc:
We cannot break liveCD environments using MATE. That could kill things like Ubuntu-MATE outright

libcaja-private/caja-file-utilities.c
```diff
@@ -1134,6 +1134,33 @@ caja_is_in_system_dir (GFile *file)
     return res;
 }
 
+gboolean
+caja_is_in_desktop_dir (GFile *file)
+{
+    const char *desktop_dir;
```
raveit65, line 1140: Build warning.
```
caja-file-utilities.c:1140:17: warning: declaration shadows a variable in the global scope [-Wshadow]

    const char *desktop_dir;
```
see https://travis-ci.org/github/mate-desktop/caja/jobs/745617527#L4604

raveit65:
> > LiveCDs are coming with desktop files like `Install-Live-CD-to-disc.`
> > So this PR will mark them as un-trusted which isn't true.
> 
> They are marked as trusted in `caja-directory-async: trust also launchers from user desktop`.

Yes. now i see it and it works like expected.
Desktop files on the desktop are marked as trusted but not desktop files in eg. home dir of user account.

libcaja-private/caja-file-utilities.c
rbuj: @raveit65 done

raveit65 (approved):
LGTM,
does what it promise.

vkareh (approved):
This is a nice feature.

LGTM

PR #1469, opened 2020-11-24
file-operations: don't attempt to move into the source dir

ericek111:
Dragging a file in caja onto the same active tab in tab bar shows a conflict resolution dialog. This is annoying, sometimes I start dragging an item, only to realize that it doesn't need to be moved, so I drop it onto the tab, subconsciously thinking it's the safest place to do so.

### DIFF
--- a/libcaja-private/caja-file-operations.c
+++ b/libcaja-private/caja-file-operations.c
@@ -5882,6 +5882,7 @@ caja_file_operations_copy_move (const GList *item_uris,
 {
 	GList *locations;
 	GList *p;
+	GFile *src_dir;
 	GFile *dest;
 	GtkWindow *parent_window;
 	gboolean target_is_mapping;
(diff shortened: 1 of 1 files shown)


libcaja-private/caja-file-operations.c
```diff
@@ -5918,10 +5919,10 @@ caja_file_operations_copy_move (const GList *item_uris,
 		parent_window = (GtkWindow *)gtk_widget_get_ancestor (parent_view, GTK_TYPE_WINDOW);
 	}
 
+	src_dir = g_file_get_parent (locations->data);
```
rbuj, line 5922: There is a memory leak:
```C
g_object_unref (src_dir);
```

ericek111:
I have fixed the leak (sorry for that), but now build fails for Debian and Arch with some obscure error. Perhaps you know what to do. Should I push a commit to trigger another build or is this fine for merging the PR?

libcaja-private/caja-file-operations.c
```diff
@@ -5966,10 +5962,14 @@ caja_file_operations_copy_move (const GList *item_uris,
 					       done_callback, done_callback_data);
 	}
 
-    	g_list_free_full (locations, g_object_unref);
+    g_list_free_full (locations, g_object_unref);
```
rbuj, line 5965: This line has not been modified. Could you leave it as it was before?

rbuj:
You should squash all commits into a single one.

libcaja-private/caja-file-operations.c
ericek111: That padding is gonna haunt me. :( Caja is overdue for some serious linting. Rolled back, as it has nothing to do with this PR.

rbuj (approved):
LGTM, no action is taken, this could be improved by displaying a message in the log, however this avoids the file replacement dialog.

raveit65 (approved):
Does what it promise.
Thx.

PR #1470, opened 2020-11-26
caja-file-management-properties: use g_list_free_full

rbuj:


raveit65:
I am still wondering that `g_list_foreach ()` can be replaced with `g_list_free_full ()`.
Are you sure?
There description sound different.
https://developer.gnome.org/glib/unstable/glib-Doubly-Linked-Lists.html#g-list-foreach
`Calls a function for each element of a GList.`
`It is safe for func to remove the element from list , but it must not modify any part of the list after that element.`
https://developer.gnome.org/glib/unstable/glib-Doubly-Linked-Lists.html#g-list-free-full
`Convenience method, which frees all the memory used by a GList, and calls free_func on every element's data.`

lukefromdc (approved):
I checked https://developer.gnome.org/glib/stable/glib-Doubly-Linked-Lists.html#g-list-free-full for myself and this should work. A test build and install worked fine and didn't throw any terminal warnings. I ran it from gdb under SU as I didn't want to kill my running instance in my session, was also testing gdb (had an earlier symbol reading issue w it now fixed), and wanted to see any warnings thrown

rbuj:
https://github.com/GNOME/glib/blob/5e1d368eecc3311a172b015b6fb1667761334058/glib/glist.c#L218-L246
```C
/**
 * g_list_free_full:
 * @list: the first link of a #GList
 * @free_func: the function to be called to free each element's data
 *
 * Convenience method, which frees all the memory used by a #GList,
 * and calls @free_func on every element's data.
 *
 * @free_func must not modify the list (eg, by removing the freed
 * element from it).
 *
 * It can be combined with g_steal_pointer() to ensure the list head pointer
 * is not left dangling ­— this also has the nice property that the head pointer
 * is cleared before any of the list elements are freed, to prevent double frees
 * from @free_func:
 * |[<!-- language="C" -->
 * GList *list_of_owned_things = …;  /<!-- -->* (transfer full) (element-type GObject) *<!-- -->/
 * g_list_free_full (g_steal_pointer (&list_of_owned_things), g_object_unref);
 * ]|
 *
 * Since: 2.28
 */
void
g_list_free_full (GList          *list,
                  GDestroyNotify  free_func)
{
  g_list_foreach (list, (GFunc) free_func, NULL);
  g_list_free (list);
}
```

raveit65 (approved):
LGTM

PR #1474, opened 2020-12-21
caja-bookmarks-window: upgrade ui file

rbuj:


lukefromdc (approved):
Builds and runs fine. For all the changed lines, this is a simple change: require GTK 3.22 or later in the file, and replace - with _  in all the property names.  Bookmark editing window works same as before. Will merge this

PR #1475, opened 2020-12-28
Remove warning -Wtype-limits

rbuj:
```
caja.make.log:/usr/include/glib-2.0/glib/gmacros.h:813:63: warning: comparison of unsigned expression in '< 0' is always false [-Wtype-limits]
caja.make.log-  813 | #define CLAMP(x, low, high)  (((x) > (high)) ? (high) : (((x) < (low)) ? (low) : (x)))
caja.make.log-      |                                                               ^
```


raveit65 (approved):
LGTM,
warning is gone.

PR #1476, opened 2021-01-02
caja-places-sidebar: upgrade format drive item

### LINKED ISSUE #873: [Feature request] Format option missing for USB devices
#### Expected behaviour
On the devices side, right click on the USB device plugged, then select "Format..."
The "Format" option is already in Nautilus, see capture attached.

**This option is useful for beginner users, in particular for old people.**

nautilus format usb

#### Actual behaviour
On the devices side, right click on the USB device plugged, the "Format..." option is missing, see capture attached.
mate format missing

Plug an USB device, as a stick. Open Caja. Right click on the USB plugged.

#### Linux Distribution
Format option missing in Ubuntu Mate 14.04, 16.04 and 17.10

rbuj:
based on https://github.com/GNOME/nautilus/commit/a73c8c7736cee89bf5470de47aec876b86878bcd
closes #873

### DIFF
--- a/src/caja-places-sidebar.c
+++ b/src/caja-places-sidebar.c
@@ -26,9 +26,11 @@
 #include <cairo-gobject.h>
 
 #include <gdk/gdkkeysyms.h>
+#include <gdk/gdkx.h>
 #include <gtk/gtk.h>
 #include <glib/gi18n.h>
 #include <gio/gio.h>
+#include <gio/gdesktopappinfo.h>
 #include <libnotify/notify.h>
 
 #include <eel/eel-debug.h>
@@ -1735,6 +1737,18 @@ check_visibility (GMount           *mount,
 
         if (*show_stop)
             *show_unmount = FALSE;
+
+        if (volume != NULL)
+        {
+            gchar *unix_device_id;
+            gchar *disks_path;
+
+            unix_device_id = g_volume_get_identifier (volume, G_VOLUME_IDENTIFIER_KIND_UNIX_DEVICE);
+            disks_path = g_find_program_in_path ("gnome-disks");
+            *show_format = (unix_device_id != NULL) && (disks_path != NULL);
+            g_free (unix_device_id);
+            g_free (disks_path);
+        }
     }
 
     if (volume != NULL)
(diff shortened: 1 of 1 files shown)


lukefromdc (commented):
This formatted a flash drive as VFAT without issue, and the reformatted  device appeared in the sidebar immediately.  

An attempt to format as EXT4 also worked, though the device disappeared from the sidebar and did not reappear. If the device was left in place a while, removed,and then reinserted, it then appeared as reformatted and worked as expected.  If the user removes the device before this process finishes, an error dialog pops up to indicate a failed attempt to format the (no longer present) drive. Since the partially formatted drive is NOT mountable, it does not appear in the sidebar and recovery is only possible from the commandline.

This may be due to the time needed to fully create the ext4 filesystem. Thus, something needs to indicate "pending" while the filesystem is still being written to prevent the user from prematurely removing the drive. Also if we are going to format removable devices from the sidebar, we should show them in the sidebar whether or not they are mountable, as there is a valid interaction with them in either case.

Not listing this as "changes requested" as I will be very busy (and in harm's way) this coming week.

raveit65 (approved):
Works fine

raveit65:
Not sure if it is possible. But having a notification when gnome-disk process has been done might be a good idea.
But maybe this is to complicated.

rbuj:
gnome-disk-utility already provides an application called gsd-disk-utility-notify that monitors dbus messages and triggers notifications.

raveit65:
ok,
but i didn't saw an notification when formating of my usb-stick was finished.

rbuj:
Yes you're right, apparently it's only for SMART (Self-Monitoring, Analysis, and Reporting Technology)
https://gitlab.gnome.org/inigomartinez/gnome-disk-utility/-/blob/master/src/notify/gdusdmonitor.c

raveit65:
But i think it is OK, because it should be be impossible to unmount the disk or write to the disk, when the formatting process has access on the disk and isn't finished.

rbuj:
Before formatting, It makes sure the volume not in use:
```C
      /* ensure the volume is unused (e.g. unmounted) before formatting it... */
      gdu_utils_ensure_unused (data->client,
                               GTK_WINDOW (data->parent_window),
                               data->object,
                               (GAsyncReadyCallback) ensure_unused_cb,
                               NULL, /* GCancellable */
                               data);
```
https://github.com/GNOME/gnome-disk-utility/blob/c84a863c1d78d3b2e6cfca71bdd14c04578e4db7/src/disks/gducreateformatdialog.c#L399-L405

The changes on volumes are monitored using gvfs-udisks2-volume-monitor
https://gitlab.gnome.org/GNOME/gvfs/-/tree/master/monitor/udisks2

> The gvfs-udisks2-volume-monitor process is responsible for the disks, media, mounts and fstab entries shown in the desktop user interface. In particular, GNOME Shell, GNOME Files as well as any other application using the GLib APIs, is using information from this process.[5] In Debian it is distributed as part of the gvfs-daemons package.[10] It is not to be confused with udisksd and udisksctl, which are part of the udisks2 package.[11]
> udisks: The udisks daemon serves as an interface to system block devices, implemented via D-Bus. It handles operations such as querying, mounting, unmounting, formatting, or detaching storage devices such as hard disks or USB thumb drives.
https://en.wikipedia.org/wiki/GVfs

I checked the code, and it only displays an error dialog when it failed to format a volume:
```C
  if (!udisks_block_call_format_finish (UDISKS_BLOCK (source_object), res, &error))
    {
      gdu_utils_show_error (GTK_WINDOW (data->parent_window), _("Error formatting volume"), error);
      g_error_free (error);
    }
```
https://github.com/GNOME/gnome-disk-utility/blob/c84a863c1d78d3b2e6cfca71bdd14c04578e4db7/src/disks/gducreateformatdialog.c#L219-L223

I also think it should send a notification when the formatting finished successfully or an error occurred.

rbuj:
https://gitlab.gnome.org/GNOME/gnome-disk-utility/-/merge_requests/38

raveit65:
Notification works fine with gnome-disk-utility PR.
Thanks :)

lukefromdc:
Any progress on this? Do we plan to port over the notifications from gnome-disk-utility, merge this as-is, drop it, or do something else?

raveit65:
I think it is OK to merge it without notifications by g-d-u.
It is impossible to do any action with the filesystem before the `formatting` process is finished.
No need to wait of a probably slow gnome MR process, ....ihmo.

danboid:
Great news!

Thanks for adding this important missing feature @rbuj ! Now gnome-disks just needs to add exfat support.

rbuj:
https://github.com/rbuj/mate-disk-utility shows the notification... if you want we can add it to MATE projects list ;-)

raveit65:
Sure, we can add it to mate-wiki.
But what is the problem with your MR?

rbuj:
I forked gnome-disks to avoid future issues, as the author said that the xid parameter is obsolete, so it won't be possible to open the formatting dialog in a modal way from the command line. I don't like that gnome-disks uses a hardcoded header bar, ie no titlebar :-(

https://gitlab.gnome.org/GNOME/gnome-disk-utility/-/commit/8818a11010a07800638fa9b33163d3793bcf5f4a

> The HdyWindow widget is a subclass of GtkWindow which has no titlebar area and provides rounded corners on all sides, ensuring they can never be overlapped by the content. This makes it safe to use headerbars in the content area as follows:

https://gnome.pages.gitlab.gnome.org/libhandy/doc/1.0/HdyWindow.html

PR #1477, opened 2021-01-16
Update copyright to 2021

rbuj:


raveit65:
@rbuj @mate-desktop/core-team 
I don't think that a pull request is necessary for updating date of copyright ;)
No need for extra work by a reviewer.
If the package builds for locally you can commit it directly to master, please....

Btw., updating copyright all over the place is the first step for a new major release.

rbuj:
Ok, I'm going to do as you said.
Translations should also be updated before the major release in order to give time to translate it.

raveit65:
There is also a problem with a new language by transifex, locale `tzm`. after `tx pull -af` or ` tx pull -l tzm` package building is broken. 
I always dropped this new locale after a pull in the the last months.

Issue #1480, opened 2021-01-22
Backgrounds and emblems dialog shows only "patterns" after commit 40f061c638ae125fc6a4170807fa890362050e0f


lukefromdc:
#### Expected behaviour
When backgrounds and emblems dialog is opened, "patterns" should be the default screen shown, plus selectable boxes to pick "colors" and "emblems"

#### Actual behaviour
Only the default "patterns" screen is shown, the three selectable boxes to the left in the dialog do NOT appear. Culprit commit traced to
https://github.com/mate-desktop/caja/commit/40f061c638ae125fc6a4170807fa890362050e0f
which is also the only commit affecting this dialog since our last "good" version in Sep 2020

#### Steps to reproduce the behaviour
Install caja from current git master or 1.25.1 release.
Open the Backgrounds and Emblems dialog

#### MATE general version
1.25

#### Package version
1.25.2+git20210122 locally complied from today's git master

#### Linux Distribution
Debian Unstable

#### Link to bugreport of your Distribution (requirement)
None as this is a locally compiled package. Note that Debian unstable at this time should NOT have this bug as they are shipping 1.24.0-1 and the offending commit does not appear in the 1.24 branch.

lukefromdc:
What the dialog should look like:
![Complete_Dialog](https://user-images.githubusercontent.com/14036792/105546144-ea714700-5cf4-11eb-9069-ccba57fa0c0b.jpg)

lukefromdc:
What we are actually getting after https://github.com/mate-desktop/caja/commit/40f061c638ae125fc6a4170807fa890362050e0f
![InComplete_Dialog](https://user-images.githubusercontent.com/14036792/105546200-feb54400-5cf4-11eb-8acd-e8400c15bbc3.jpg)

rbuj:
I couldn't reproduce your issue on Fedora:

![Screenshot at 2021-01-22 22-09-05](https://user-images.githubusercontent.com/10171411/105547244-966b6000-5cfe-11eb-9578-e5fc1b8e93ac.png)

Build script for testing: 
- https://gist.github.com/rbuj/bc62164be10a7290f5213c47711c7998#file-test-mate-desktop-sh

lukefromdc:
I get this on Debian Unstable repeatedly, with different builds from different versions. Note that I have unused locales purged, though localpurge doesn't clear them until I actively run it. That seems to make no difference.

I am using the LANG=en_US.UTF-8 locale

[...]

PR #1481, opened 2021-01-22
Do not make browser.xml from browser.xml.in

rbuj:
The translations are stored in the po files, ie they are retrieved via gettext, for this reason it's not necessary to generate the xml file including the translations.


lukefromdc:
This didn't fix
https://github.com/mate-desktop/caja/issues/1480
either, I don't know enough about i18n to have any idea what is going on

rbuj:
#1483 renames `browser.xml` to `browser.xml.in` and converts the generic target pattern `%.xml: %.xml.in` to the specific pattern `$(mime_DATA): $(mime_in_files)`.

PR #1484, opened 2021-01-24
caja-application: Use PACKAGE in notify_init instead of GETTEXT_PACKAGE

rbuj:


lukefromdc (approved):
Eject notifications work same as before, though note that I cannot evaluate the code change as I do not know how i18n and gettext work. Should get a test by someone using a language other than English.

raveit65 (approved):

PR #1488, opened 2021-01-28
Fix desktop file validation warning

rbuj:
```
$ find -type f -name *.desktop -exec desktop-file-validate {} \;
./data/caja-browser.desktop: hint: value "GTK;System;Utility;Core;" for key "Categories" in group "Desktop Entry" contains more than one main category; application might appear more than once in the application menu
```
https://specifications.freedesktop.org/menu-spec/latest/apa.html


raveit65 (approved):
LGTM

PR #1490, opened 2021-02-03
caja-directory-async: drop istr_set_insert and use g_hash_table_add instead

rbuj:


raveit65 (approved):
LGTM,
builds and runs fine.

lukefromdc (approved):
File builds with no warnings on my usual settings, and the file copy dialogs work as expected. 
At the moment I am copying from a backup hard drive to another hard drive, not a flash drive and this is doubling as a basic file dialog test

Note that what I have seen for the "delayed sync" issue is flash drives waiting on eject (ejecting from the caja sidebar) until the data is finished copying before showing the notification that it is safe to eject the drive. That notification (another portover from Nemo as I recall) seems to work correctly.

I never just pull flash drives because that is asking for data loss. I've never after waiting for that notification had the drive turn around and not have the data physically written to flash. The other indicator is seeing the drive disappear from the sidebar: if that doesn't happen I leave it in.

That said, if users are pulling drives as soon as a copy dialog finishes, that should hold up for syncing the files

lukefromdc:
Is this intended to fix
https://github.com/mate-desktop/caja/issues/1494 ?

rbuj:
> Is this intended to fix
> #1494 ?

No, it isn't. From my point of view there is no issue in #1494 , perhaps the user should tune up the system settings regarding the disk write policy, write-back caching, ...

PR #1497, opened 2021-02-21
Rename wallpaper pattern name from MATE to GNOME

rbuj:
browser.xml: use two white spaces for indentation


lukefromdc (approved):
This works and testing it is not affected by https://github.com/mate-desktop/caja/issues/1480 as the "patterns" window is the one that shows by default in the backgrounds and emblems dialog.
I assume this is to reflect GNOME's authorship of the artwork in question?

raveit65:
Why renaming back after 8 years?

lukefromdc:
I have no idea, I only did a functional test

rbuj:
The footprint is the [logo of gnome](https://www.gnome.org/logo-and-trademarks/).

raveit65:
Indeed, there is a gnome footprint :)

raveit65 (approved):
LGTM

PR #1500, opened 2021-02-25
caja-window: Fix cppcheck warning unusedAllocatedMemory

rbuj:
```
src/caja-window.c:1530:16: style: Variable 'copy' is allocated memory that is never used. [unusedAllocatedMemory]
```

### DIFF
--- a/src/caja-window.c
+++ b/src/caja-window.c
@@ -1525,12 +1525,8 @@ real_sync_title (CajaWindow *window,
 {
     if (slot == window->details->active_pane->active_slot)
     {
-        char *copy;
-
-        copy = g_strdup (slot->title);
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
This builds and runs fine, though note that caja-window.c builds with so many warnings I was not able to see if any were new. At runtime window titles (as shown in the titlebar) change when changing the directory displayed, same as before.

The code here is interesting: it looks like someone at one point decided to copy the string possibly for safety reasons, only to either not use the resulting` copy` variable in calling 
`g_signal_emit_by_name ` or for this to later be changed out for using the string that had been copied directly. Anyway works exactly the same as before, as that variable is only valid in the scope of this function and is never used. 

Simple change, merging

Issue #1498, opened 2021-02-24
Files dragged on the home icon are sometimes copied


robertxgray:
I am experiencing the issue #786 in Ubuntu 20.04.

#### Expected behaviour
When dragging files from desktop to the home icon, they should be moved.

#### Actual behaviour
The files are copied, not moved.

#### Steps to reproduce the behaviour
1. Restart caja
2. Drag a file from desktop to the home icon

#### MATE general version
1.24.0

#### Package version
1.24.0-1

#### Linux Distribution
Ubuntu Mate 20.04.2 LTS

robertxgray:
Just checked Fedora 33 (caja 1.24.0-5.fc33) and it has the same problem. The easiest way to reproduce the incorrect behaviour is:
1. Reboot
2. Login
3. Drag a file on the home icon before anything else

I have replaced the desktop home icon with a symlink as a temporary (and ugly) workaround.

PR #1502, opened 2021-02-27
Fix unknown filesystem on home icon dnd

robertxgray:
When files are dropped on the user home icon, it's file system data can be unavailable. This happens if the home folder has not been browsed before. Therefore, caja copies the files to the home folder even if home and desktop share the same file system.

On caja-dnd.c:458, the function caja_file_get_existing_by_uri returns a null target_file, causing check_same_fs to get a wrong result. The function with caja_file_get_by_uri returns a CajaFile, but with no filesystem_id.

### DIFF
--- a/libcaja-private/caja-dnd.c
+++ b/libcaja-private/caja-dnd.c
@@ -374,28 +374,42 @@ caja_drag_default_drop_action_for_netscape_url (GdkDragContext *context)
 }
 
 static gboolean
-check_same_fs (CajaFile *file1,
-               CajaFile *file2)
+check_same_fs (const char *target_uri, 
+               CajaFile *target_file,
+               const char *dropped_uri,
+               CajaFile *dropped_file)
 {
     gboolean result;
 
     result = FALSE;
 
-    if (file1 != NULL && file2 != NULL)
+    char *target_fs = NULL, *dropped_fs = NULL;
+
+    if (target_file != NULL)
     {
-        char *id1, *id2;
-
-        id1 = caja_file_get_filesystem_id (file1);
-        id2 = caja_file_get_filesystem_id (file2);
-
-        if (id1 != NULL && id2 != NULL)
-        {
-            result = (strcmp (id1, id2) == 0);
-        }
-
-        g_free (id1);
-        g_free (id2);
+        target_fs = caja_file_get_filesystem_id (target_file);
     }
+    if (target_fs == NULL)
+    {
+        target_fs = caja_get_filesystem_id_by_uri (target_uri, TRUE);
+    }
+
+    if (dropped_file != NULL && !caja_file_is_symbolic_link (dropped_file))
+    {
+        dropped_fs = caja_file_get_filesystem_id (dropped_file);
+    }
+    if (dropped_fs == NULL)
+    {
+        dropped_fs = caja_get_filesystem_id_by_uri (dropped_uri, FALSE);
+    }
+
+    if (target_fs != NULL && dropped_fs != NULL)
+    {
+        result = (strcmp (target_fs, dropped_fs) == 0);
+    }
+
+    g_free (target_fs);
+    g_free (dropped_fs);
 
     return result;
 }
@@ -519,7 +533,7 @@ caja_drag_default_drop_action_for_icons (GdkDragContext *context,
         target = g_file_new_for_uri (target_uri_string);
     }
 
-    same_fs = check_same_fs (target_file, dropped_file);
+    same_fs = check_same_fs (target_uri_string, target_file, dropped_uri, dropped_file);
 
     caja_file_unref (dropped_file);
     caja_file_unref (target_file);
--- a/libcaja-private/caja-file-utilities.c
+++ b/libcaja-private/caja-file-utilities.c
@@ -1299,6 +1299,42 @@ caja_restore_files_from_trash (GList *files,
     caja_file_list_unref (unhandled_files);
 }
 
+char *
+caja_get_filesystem_id_by_location (GFile *location, gboolean follow)
+{
+    GFileInfo *info;
+    GFileQueryInfoFlags flags;
+    char *filesystem_id = NULL;
+
+    if (follow) {
+        flags = G_FILE_QUERY_INFO_NONE;
+    } else {
+        flags = G_FILE_QUERY_INFO_NOFOLLOW_SYMLINKS;
+    }
+
+    info = g_file_query_info (location, G_FILE_ATTRIBUTE_ID_FILESYSTEM, flags, NULL, NULL);
+    if (info) {
+        if (g_file_info_has_attribute (info, G_FILE_ATTRIBUTE_ID_FILESYSTEM)) {
+            filesystem_id = g_strdup (
+                g_file_info_get_attribute_string (info, G_FILE_ATTRIBUTE_ID_FILESYSTEM));
+        }
+        g_object_unref (info);
+    }
+    return filesystem_id;
+}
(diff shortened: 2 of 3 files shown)


robertxgray:
This pull request fixes #1498

libcaja-private/caja-dnd.c
```diff
@@ -517,6 +517,12 @@ caja_drag_default_drop_action_for_icons (GdkDragContext *context,
     else
     {
         target = g_file_new_for_uri (target_uri_string);
+
+        if (target_file == NULL)
+        {
+            target_file = caja_file_get_by_uri (target_uri_string);
```
rbuj, line 523: You called twice `caja_file_get_internal`:
1. `caja_file_get_internal (location, FALSE); ` on `target_file = caja_file_get_existing_by_uri (target_uri_string);`
2. `caja_file_get_internal (location, TRUE);` on `target_file = caja_file_get_by_uri (target_uri_string);`.

So, if the directory doesn't exist in the directories lookup table (1), you add it on (2).

```C
CajaFile *
caja_file_get_existing_by_uri (const char *uri)
{
        GFile *location;
        CajaFile *file;

        location = g_file_new_for_uri (uri);
        file = caja_file_get_internal (location, FALSE);
        g_object_unref (location);

        return file;
}

CajaFile *
caja_file_get_by_uri (const char *uri)
{
        GFile *location;
        CajaFile *file;

        location = g_file_new_for_uri (uri);
        file = caja_file_get_internal (location, TRUE);
        g_object_unref (location);

        return file;
}
```

```C
CajaDirectory *
caja_directory_get_internal (GFile *location, gboolean create)
{
    CajaDirectory *directory;

    /* Create the hash table first time through. */
    if (directories == NULL) {
        directories = g_hash_table_new (g_file_hash, (GCompareFunc) g_file_equal);
        caja_global_preferences_init ();
    }

    /* If the object is already in the hash table, look it up. */

    directory = g_hash_table_lookup (directories,
                                     location);
    if (directory != NULL)
    {
        caja_directory_ref (directory);
    }
    else if (create)
    {
        /* Create a new directory object instead. */
        directory = caja_directory_new (location);
        if (directory == NULL)
        {
            return NULL;
        }

        /* Put it in the hash table. */
        g_hash_table_insert (directories,
                             directory->details->location,
                             directory);
    }

    return directory;
}
```

rbuj: What is the purpose of `caja_file_refresh_info (target_file);`?

rbuj: we may be using two entries for the same location:
VirtualBox_11 Fedora Rawhide_27_02_2021_21_06_22

robertxgray: My first fix attempt was replacing caja_file_get_existing_by_uri with caja_file_get_by_uri on caja-dnd.c:458.

```diff
- target_file = caja_file_get_existing_by_uri (target_uri_string);
+ target_file = caja_file_get_by_uri (target_uri_string);
```
This solves target_file being null, but then target_file->details->filesystem_id is null. So you still get the same result in check_same_fs.

The purpose of caja_file_refresh_info is to load the right value in filesystem_id. I had a look at the existing functions and that was the most appropriate for the task that I could find, but note that I'm not familiar with the code of caja.

I'm aware that caja_file_get_internal is being called twice. I left it as is in order to keep the fix as simple and less intrusive as possible. You could also revert to the original replacement on line 458 and then use a different criteria to determine if it's necessary to update filesystem_id or not.

rbuj: Maybe the issue is due the actual target does not exist in the lookup table of directories because the virtual file system uri was added instead, so it failed to check whether the source and target belong to the same file system, which is why the copy operation is selected instead of the move operation.

robertxgray: You know the internals of caja far better than me, please feel free to solve the issue in any other way that you consider appropriate. Just tell me if there's any way I can help in order to have the problem solved asap, because I find it annoying. Thank you very much.

rbuj (changes requested):
It also affects other links such as `computer:///`
https://github.com/mate-desktop/caja/blob/82762a4c10d3e2ab99eb9ae6fea2b27e117549f0/libcaja-private/caja-desktop-link.c#L210-L284

It should be enough to use the link activation address:

https://github.com/mate-desktop/caja/blob/82762a4c10d3e2ab99eb9ae6fea2b27e117549f0/libcaja-private/caja-file.c#L4219-L4233

robertxgray:
I don't understand the change you are requesting. Care to explain please?

rbuj:
You need to fix why the activation location of the virtual home link that is placed on desktop isn't in the hash, instead to force its addition which may affect other virtual links.

robertxgray:
I pushed a different solution to have files referenced by links in the hash, please have a look.

The bug also affects links created with desktop files in any path (and the solution fixes them), I mean something like this:
```desktop
[Desktop Entry]
Name=My Link
Icon=folder
Type=Link
URL=/path/in/same/fs
```

rbuj:
Do not mix the path to the desktop file, if it exists, with its contents.
There is no need to monitor the paths to the desktop files.

robertxgray:
I have been analyzing this problem in depth. It affects not only desktop icon files, but also caja links and symbolic links. All these elements include a reference to another path that is potentially absent in the hash.

When they are part of dnd operations, the original path is used when the element is dragged and the referred path is used when the element is the drop target. However, we are currently storing only the original filesystem for caja links and the referred filesystem for symbolic links. Because of this, filesystem identification is wrong when dragging symbolic links or dropping over caja links. Desktop icon files are wrong in both cases, but only dropping files on them makes sense.

The easiest solution is to identify and calculate the missing filesystems in caja_drag_default_drop_action_for_icons. It's not the most efficient solution but my first clumsy attempt can be improved.

Monitoring the referred paths fixes the problem for caja links and desktop icons. The current version doesn't fix symbolic links but maybe it can be improved. This solution doesn't add overload in dnd operations but in the extra monitoring instead.

I have also been working in a third solution that stores both filesystems in the CajaFilePrivate structure. The problem is that we're currently using the referred paths in all the dnd operations and we would require the original paths instead. It's a complex solution with significant impact in the existing dnd code, but I can continue with it if you think it's worth it.

I would appreciate more proposals or feedback. Thank you.

robertxgray:
I have pushed another proposal with an improved version of caja-dnd.check_same_fs. I think it covers all the use cases.

libcaja-private/caja-dnd.c
```diff
@@ -454,8 +467,8 @@ caja_drag_default_drop_action_for_icons (GdkDragContext *context,
     }
 
     dropped_uri = ((CajaDragSelectionItem *)items->data)->uri;
-    dropped_file = caja_file_get_existing_by_uri (dropped_uri);
-    target_file = caja_file_get_existing_by_uri (target_uri_string);
+    dropped_file = caja_file_get_by_uri (dropped_uri);
+    target_file = caja_file_get_by_uri (target_uri_string);
```
rbuj, line 471: I love the approach of checking the file system id to see whether the two files are in the same file system.
However, I don't think it's right adding the files to the hash table here if they don't exist yet.

robertxgray: I have updated the code. Thank you.

rbuj (approved):
LGTM

rbuj:
Can you squash all commits in a single one?

raveit65 (approved):
LGTM,
runs fine.

PR #1504, opened 2021-03-06
caja-places-sidebar: gnome-disk-utility ignores the --xid option

rbuj:
https://gitlab.gnome.org/GNOME/gnome-disk-utility/-/commit/9f657153


raveit65 (approved):
LGTM

PR #1510, opened 2021-03-23
icon view: Add a directory setting to block moving icons

cwendling:
This PR adds a new menu item in the icon view context menu to lock icons in place:
caja-lock-directory

This is mostly useful on the desktop (which is the main user of the free-placement icon view) to be able to lock the layout and avoid
unintentionally moving icons around.  Especially useful for less computer-literate users.


libcaja-private/caja-icon-container.c
```diff
@@ -8740,6 +8746,16 @@ caja_icon_container_set_keep_aligned (CajaIconContainer *container,
...
+void
+caja_icon_container_set_lock_icons_position (CajaIconContainer *container,
+                                             gboolean lock_icons_position)
+{
+    if (container->details->lock_icons_position != lock_icons_position)
+    {
+        container->details->lock_icons_position = lock_icons_position;
+    }
+}
+
```
raveit65, line 8758: Build warnings. https://travis-ci.org/github/mate-desktop/caja/jobs/764131695#L801
```
libcaja-private/caja-icon-container.c:8753:49: style: The statement 'if (container->details->lock_icons_position!=lock_icons_position) container->details->lock_icons_position=lock_icons_position' is logically equivalent to 'container->details->lock_icons_position=lock_icons_position'. [duplicateConditionalAssign]

    if (container->details->lock_icons_position != lock_icons_position)

                                                ^

libcaja-private/caja-icon-container.c:8755:49: note: Assignment 'container->details->lock_icons_position=lock_icons_position'

        container->details->lock_icons_position = lock_icons_position;

                                                ^

libcaja-private/caja-icon-container.c:8753:49: note: Condition 'container->details->lock_icons_position!=lock_icons_position' is redundant

    if (container->details->lock_icons_position != lock_icons_position)
```

src/file-manager/fm-icon-view.c
```diff
@@ -990,6 +996,38 @@ fm_icon_view_set_directory_keep_aligned (FMIconView *icon_view,
...
+    }
+
+    return caja_file_get_boolean_metadata
+            (file,
+             CAJA_METADATA_KEY_ICON_VIEW_LOCK_ICONS_POSITION,
+             get_default_directory_lock_icons_position ());
+}
+
+static void
+fm_icon_view_set_directory_lock_icons_position (FMIconView *icon_view,
```
raveit65, line 1021: Warning.
https://travis-ci.org/github/mate-desktop/caja/jobs/764131695#L11413
```
fm-icon-view.c:1021:61: warning: unused parameter 'icon_view' [-Wunused-parameter]

fm_icon_view_set_directory_lock_icons_position (FMIconView *icon_view,
```

raveit65:
Nice idea, i think a lot of people will love it....

lukefromdc (commented):
Despite the warning, this builds and runs as expected and locking icons in place from the desktop right-click menu works

libcaja-private/caja-icon-container.c
cwendling: Indeed, I could ditch the test. Will do.

src/file-manager/fm-icon-view.c
cwendling: Fair enough; what am I supposed to do here?  I recall last time I used `G_GNUC_UNUSED` it started a abundant thread on how we should suppress that, so I'd rather not replicate that here and use whatever is preferred :slightly_smiling_face:

cwendling: Actually it could make sense to test the view settings, so I'll do and use that parameter instead.

cwendling:
I addressed the comments, and changed the following:
* Use the more conservative default of *not* locking icon positions by default.  It was an oversight on my part, as I first prepared it for a environment where `default=true` was the best fit, but I'm not convinced it is for regular Caja.
* Use a more reasonable guard for setting/getting the locking state -- although the previous of *keep aligned* works effectively exactly the same, it's conceptually not a good fit, *manual placement* support seems a better one.

Effectively, the only expected change is that by default Desktop should *not* be locked anymore.  Note that if you tested the previous iteration the setting will be already written in `~/.config/caja/desktop-metadata`, so to really test the default you'd have to manually delete the relevant line near the top of that file and test again.

raveit65:
> Effectively, the only expected change is that by default Desktop should not be locked anymore. 

I think this is much better ;)

raveit65 (approved):
Does what it promise.
And build warnings are gone.

Lapazlane:
I'm not clear - will this enable lock icon position in folder as well as desktop?  Such a feature would be extremely useful as I often have large numbers of files where their placement and organisation is essential in making sense of and easily locating the information they hold. Thank you.

lukefromdc (approved):
Approving as this worked in my original test and the build warnings are not reported fixed

lukefromdc:
Also, just tested behavior in a folder: if and only if manual arrangement of icons is selected for a folder, the option to lock icon positions will be shown and it works. Best of all, this is stored (and thus can be set) separately for each directory for which manual arrangement has been selected. The desktop is simply a special case of this, where manual arrangement is the default.

PR #1512, opened 2021-03-25
Limit dnd operations on desktop icons

robertxgray:
Some drag and drop operations involving desktop icon files are not defined and produce an error as result. IMHO the message errors are meaningless for most users and their experience would be better if such operations cannot be performed.

Expected behaviour with this change:

- Computer and network icons cannot be used as a drop target.
- Desktop icon files cannot be dropped on other files or folders.

### DIFF
--- a/libcaja-private/caja-dnd.c
+++ b/libcaja-private/caja-dnd.c
@@ -432,7 +432,9 @@ caja_drag_default_drop_action_for_icons (GdkDragContext *context,
     GdkDragAction actions;
     CajaFile *dropped_file, *target_file;
 
-    if (target_uri_string == NULL)
+    if ((target_uri_string == NULL) ||
+        (strcmp (target_uri_string, "computer:///") == 0) ||
+        (strcmp (target_uri_string, "network:///") == 0))
     {
         *action = 0;
         return;
@@ -457,12 +459,31 @@ caja_drag_default_drop_action_for_icons (GdkDragContext *context,
     dropped_file = caja_file_get_existing_by_uri (dropped_uri);
     target_file = caja_file_get_existing_by_uri (target_uri_string);
 
+    if (eel_uri_is_desktop (dropped_uri))
+    {
+        if (eel_uri_is_desktop (target_uri_string))
+        {
+            /* Only move to Desktop icons */
+            if (actions & GDK_ACTION_MOVE)
+            {
+                *action = GDK_ACTION_MOVE;
+            }
+        }
+        else
+        {
+            *action = 0;
+        }
+
+        caja_file_unref (dropped_file);
+        caja_file_unref (target_file);
+        return;
+    }
     /*
      * Check for trash URI.  We do a find_directory for any Trash directory.
      * Passing 0 permissions as mate-vfs would override the permissions
      * passed with 700 while creating .Trash directory
      */
-    if (eel_uri_is_trash (target_uri_string))
+    else if (eel_uri_is_trash (target_uri_string))
     {
         /* Only move to Trash */
         if (actions & GDK_ACTION_MOVE)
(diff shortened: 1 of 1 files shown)


rbuj (changes requested):
no longer shows the message to the user about the operation is not supported.

Screenshot at 2021-03-25 13-57-33

rbuj:
It involves other locations, including places side panel.

libcaja-private/caja-dnd.c
```diff
@@ -457,12 +459,31 @@ caja_drag_default_drop_action_for_icons (GdkDragContext *context,
     dropped_file = caja_file_get_existing_by_uri (dropped_uri);
     target_file = caja_file_get_existing_by_uri (target_uri_string);
 
+    if (eel_uri_is_desktop (dropped_uri))
```
rbuj, line 462: If `eel_uri_is_desktop(uri)` returned TRUE means that the uri starts with `x-caja-desktop:` prefix, for instance the **Computer** icon placed on user's desktop (x-caja-desktop:///computer)

rbuj (changes requested):
drag and drop modifier keys do not work:
- GDK_ACTION_LINK is enforced by Alt key
- GDK_ACTION_MOVE is enforced by Shift key
- GDK_ACTION_COPY is enforced by Ctrl key

Edit: The modifier keys work as expected, GDK_ACTION_COPY is enforced by pressing Ctrl+Shift.

libcaja-private/caja-dnd.c
robertxgray: That code is related to the second expected behaviour.

- Desktop icon files cannot be dropped on other files or folders.

With desktop icon files I mean any CajaDesktopIconFile, as described in caja-desktop-icon-file.h. My intention was to stop the user from performing dnd operations that lead to a sure error and leave the dnd animation as the only feedback. I think that improves the user experience, but I don't know if it's a majority opinion.

robertxgray:
To be honest I didn't know about those modifier keys, but I just performed some tests with both my proposal and the original code and couldn't find any problem. Can you provide more information about the use cases where the modifier keys are not working please?

libcaja-private/caja-dnd.c
```diff
@@ -457,12 +459,31 @@ caja_drag_default_drop_action_for_icons (GdkDragContext *context,
...
+        caja_file_unref (target_file);
+        return;
+    }
     /*
      * Check for trash URI.  We do a find_directory for any Trash directory.
      * Passing 0 permissions as mate-vfs would override the permissions
      * passed with 700 while creating .Trash directory
      */
-    if (eel_uri_is_trash (target_uri_string))
+    else if (eel_uri_is_trash (target_uri_string))
```
rbuj, line 486: It still shows the error dialog when dragging an icon with the x-caja-desktop schema to the Trash.

rbuj:
> To be honest I didn't know about those modifier keys, but I just performed some tests with both my proposal and the original code and couldn't find any problem. Can you provide more information about the use cases where the modifier keys are not working please?

Don't worry, when I tried to enforce the link creation I pressed the Alt key, however to force a link creation you have to press Shift and Control keys. I'm sorry.

libcaja-private/caja-dnd.c
```diff
@@ -491,20 +512,6 @@ caja_drag_default_drop_action_for_icons (GdkDragContext *context,
...
-            if (actions & GDK_ACTION_MOVE)
-            {
-                *action = GDK_ACTION_MOVE;
-            }
-
-            g_object_unref (target);
-            caja_file_unref (dropped_file);
-            caja_file_unref (target_file);
-            return;
-        }
```
rbuj, line 507: This makes #1502 unnecessary.

libcaja-private/caja-dnd.c
rbuj: IMHO it should be handled by gvfs backends. `caja_drag_default_drop_action_for_icons` sets the default dnd action, when it's available, and it doesn't stop/end a dnd job by setting `*action = 0;`

robertxgray:
Okay, I'll try to implement the change in gvfs backends.

PR #1520, opened 2021-04-21
Fix backgrounds and emblems dialog

lukefromdc:
Fetch name and not display_name due to xmlGetProp returning NULL on the latter. Note that use of accelerators is the only difference within the XML file

### DIFF
--- a/data/browser.xml
+++ b/data/browser.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <categories>
-  <category name="patterns" display_name="_Patterns" image="backgrounds.png" columns="5" mode="directory" path="caja/patterns" type="property/bgimage" description="Drag a pattern tile to an object to change it">
+  <category name="patterns" display_name="Patterns" image="backgrounds.png" columns="5" mode="directory" path="caja/patterns" type="property/bgimage" description="Drag a pattern tile to an object to change it">
     <pattern display_name="Blue Ridge" filename="blue_ridge"/>
     <pattern display_name="Blue Rough" filename="blue_gray_rough"/>
     <pattern display_name="Blue Type" filename="blue_type"/>
@@ -33,7 +33,7 @@
     <pattern display_name="Wavy White" filename="wavy_white"/>
   </category>
 
-  <category name="colors" display_name="C_olors" image="colors.png" columns="5" mode="inline" type="application/x-color" description="Drag a color to an object to change it to that color">
+  <category name="colors" display_name="Colors" image="colors.png" columns="5" mode="inline" type="application/x-color" description="Drag a color to an object to change it to that color">
     <color name="Mango">#E0D656</color>
     <color name="Orange">#D2AB50</color>
     <color name="Tangerine">#C6864B</color>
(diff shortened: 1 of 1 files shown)


lukefromdc:
Fix https://github.com/mate-desktop/caja/issues/1480

raveit65:
Just a rebase with master.

raveit65:
Buttons are there with commit but they are untranslated.
Bildschirmfoto zu 2021-04-27 19-08-11

lukefromdc:
OK, someone who knows how to work on that will have to fix translations as I cannot test or work on those. 
Otherwise I have a one-line fix suitable for my own use, might be good enough if nobody else ever gets this. Still no idea the reason for that though

lukefromdc:
Called for another review, we need to decide what to do with this. I cannot fix the translations

vkareh:
@lukefromdc I cannot reproduce this issue in F34.

However, I _think_ you can try translating the `name` attribute by making this change:
```diff
diff --git a/its/browser.its b/its/browser.its
index 163cca01..9c091109 100644
--- a/its/browser.its
+++ b/its/browser.its
@@ -3,6 +3,7 @@
            xmlns:gt="https://www.gnu.org/s/gettext/ns/its/extensions/1.0"
            version="2.0">
        <its:translateRule selector="/categories" translate="no"/>
+       <its:translateRule selector="//category/@name" translate="yes"/>
        <its:translateRule selector="//category/@display_name" translate="yes"/>
        <its:translateRule selector="//category/@description" translate="yes"/>
        <its:translateRule selector="//category/pattern/@display_name" translate="yes"/>
```
However, if this works, then it begs the question, what's different between `name` and `display_name`? And it's the accelerator. So if you just remove the underscores from the `display_name` attributes, and revert all other changes, does it work? If so, then that should be the solution for now, as using `name` means we lose the accelerators anyway.

Translating the `name` attribute might have unintended consequences, as it's probably used as an internal ID. I think we should try removing the accelerator markers in the `display_name` and see if that fixes it for you. Basically just change `_Patterns` to `Patterns`, `C_olors` to `Colors`, and `_Emblems` to `Emblems`.

lukefromdc:
OK,I will test this as soon as I get home from this road trip(2 daya expected)

On 7/15/2021 at 8:31 AM, "Victor Kareh" ***@***.***> wrote:
>
***@***.*** I cannot reproduce this issue in F34.
>
>However, I _think_ you can try translating the `name` attribute by 
>making this change:
>```diff
>diff --git a/its/browser.its b/its/browser.its
>index 163cca01..9c091109 100644
>--- a/its/browser.its
>+++ b/its/browser.its
>@@ -3,6 +3,7 @@
>            
>xmlns:gt="https://www.gnu.org/s/gettext/ns/its/extensions/1.0"
>            version="2.0">
>        <its:translateRule selector="/categories" translate="no"/>
>+       <its:translateRule ***@***.***" 
>translate="yes"/>
>        <its:translateRule ***@***.***_name" 
>translate="yes"/>
>        <its:translateRule ***@***.***" 
>translate="yes"/>
>        <its:translateRule 
***@***.***_name" translate="yes"/>

lukefromdc:
Just confirmed that removing accelerators works, we get back the patterns/colors/emblems options without the changes that would affect translations.  Question then becomes why do accelerators not work with the new code?

lukefromdc:
Updates this PR to just remove accelerators and rebase against master

vkareh (approved):
LGTM

raveit65:
Same problem as before. Only the emblems button is translated in my language.
Normal screenshot without this code.
caja_emblems_full_translated
With this code!
caja-emblems_new
Is it better with other languages?
Any chance that someone with debian or unbuntu installed can confirm the issue before we break translations?

lukefromdc:
We can reopen this if someone else ever gets this issue, but the issue this fixes is now gone in Debian Unstable.

I am guessing something was wrong related to libxml and got fixed. Libxml itself hasn't changed since May, so wondering what is different but the problem is now gone and this change breaks translations it seems

PR #1528, opened 2021-05-16
caja-directory.c: Fix memory leak.

mtwebster:
When creating a new file (using a template, for instance), file->
details->is_added could potentially be FALSE, and cause this file to
not be finalized along with other files if the view directory is
destroyed.

This can cause issues when re-entering that directory, with the file
being in an undefined state, and could prevent the view from fully
loading the location (this is identical behavior to that described
in https://github.com/mate-desktop/python-caja/pull/64.

To reproduce:
- Create an svg file and save in ~/Templates.
- Right-click, Create document-> svg file, name it whatever.
- Navigate out of the folder.
- Modify the file in a visible manner.
- Re-enter the folder, note that it never finishes loading.

Ref:
https://github.com/linuxmint/nemo/issues/2736

### DIFF
--- a/libcaja-private/caja-directory.c
+++ b/libcaja-private/caja-directory.c
@@ -923,7 +923,6 @@ caja_directory_notify_files_added (GList *files)
              * If it was renamed this could be ignored, but
              * queue a change just in case */
             caja_file_changed (file);
-            caja_file_unref (file);
         }
         else
         {
@@ -931,6 +930,7 @@ caja_directory_notify_files_added (GList *files)
                                      directory,
                                      g_object_ref (location));
         }
+        caja_file_unref (file);
         caja_directory_unref (directory);
     }
 


lukefromdc:
When cropping an image to a different shape with gimp and having navigated out of the directory, it loaded right back up when I navigated back into that directory. I don't have Templates so i used another, though that should have no effect. Thus I could not duplicate the bug

lukefromdc (approved):
Builds and runs fine, no problem with reloading the thumbnail after changing the file contents, though that worked with master as well on my system. Someone who can duplicate the issue should test this

lukefromdc:
This may have fixed another similar problem, which may have been an instance of the test case: when rendering out a video file to the desktop or doing another job where a file is not readable until an extended job is done (including some transfers to flash media), the thumbnail would never finish loading and hang on the "clock" thumbnail. With this installed, the thumbnail came up as soon as a test video render job finished. Thus we now have duplication of the issue and this fixed it.

lukefromdc:
Will merge this as it works and nobody else seems to be paying attention to it

lukefromdc:
Travis error is same Fedora problem

PR #1535, opened 2021-06-25
Add native support for using a bulk renaming tool.

mtwebster:
This allows the normal method of renaming a file launch a bulk
renamer if one is configured and multiple files are selected.

Ported from nemo.

### DIFF
--- a/src/file-manager/fm-directory-view.c
+++ b/src/file-manager/fm-directory-view.c
@@ -6421,6 +6421,91 @@ real_action_redo (FMDirectoryView *view)
 	caja_undostack_manager_redo (manager, GTK_WIDGET (view), finish_undoredo_callback);
 }
 
+static const gchar *known_bulk_renamers[][2] = {
+    { "bulky",  "bulky"     },
+    { "thunar", "thunar -B" }
+};
+
+static char *
+get_bulk_rename_tool (void)
+{
+    gchar *bulk_rename_tool = NULL;
+    gint i;
+
+    for (i = 0; i < G_N_ELEMENTS (known_bulk_renamers); i++) {
+        gchar *binary = g_find_program_in_path (known_bulk_renamers[i][0]);
+        if (binary != NULL) {
+            bulk_rename_tool = g_strdup (known_bulk_renamers[i][1]);
+            g_free (binary);
+            break;
+        }
+    }
+
+    if (bulk_rename_tool == NULL) {
+        g_settings_get (caja_preferences,
+                        CAJA_PREFERENCES_BULK_RENAME_TOOL,
+                        "^ay", &bulk_rename_tool);
+    }
+
+    return g_strstrip (bulk_rename_tool);
+}
+
+static gboolean
+have_bulk_rename_tool (void)
+{
+    gchar *bulk_rename_tool;
+    gboolean have_tool;
+
+    bulk_rename_tool = get_bulk_rename_tool ();
+    have_tool = ((bulk_rename_tool != NULL) && (*bulk_rename_tool != '\0'));
+    g_free (bulk_rename_tool);
+
+    return have_tool;
+}
+
+static void
+invoke_external_bulk_rename_utility (FMDirectoryView *view,
+                                     GList           *selection)
+{
+    GList *iter;
+    GString *cmd;
+    gchar *bulk_rename_tool;
+    gint i;
+
+    /* assemble command line */
+    bulk_rename_tool = get_bulk_rename_tool ();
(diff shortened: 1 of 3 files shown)


mbkma:
Thanks! It is very handy to have bulk rename built in. However, a few comments:
 - I do not think it is a good idea that the user has to enter a command manually in caja-preferences. The bulk rename tool should be automatically detected by the code, or hardcoded (such as in mate-panel window list applet).
 - It is currently possible to enter any command, so e.g. `atril` will open two instances of atril instead of renaming. This sould be avoided.
 
I tested it with the thunar bulk renamer ( `thunar --bulk-rename` ) it works fine.

mtwebster:
Hi, thanks for the feedback -

1 - I'm not sure why we shouldn't allow the user to define the bulk renamer - this allows them to use whichever one they prefer - there are a number of them out there. A default can always be defined in the settings gschema file, as well as overridden by distributions. At the very least the setting should exist, even if you'd prefer not to have it in the preferences ui.

2 - We can't detect what is a bulk renamer and what isn't. Of course the user can put anything in there, but I think you have to give them some benefit of the doubt and assume they won't enter something in there that *isn't* actually a bulk renamer.

Thank you

mbkma:
Yep, I think just a gsettings key with a default is perfect. No need to expose this to the user in the preferences ui.

For convenience you could try to find some hard coded bulk-rename-tools. Maybe sth like this works:
```c
static const char* bulk_rename_tools[] = {
	"bulky",
        ...
};

int i;
gchar *progname;
for (i = 0; i < G_N_ELEMENTS(bulk_rename_tools); i++)
{
    if ((progname = g_find_program_in_path(bulk_rename_tools[i])) != NULL);
    {
         g_free(progname);
         break;
    }
    g_free(progname);
}

g_settings_set(settings, "bulk-rename-tool", "^ay", bulk_rename_tools[i]);

```

mtwebster:
I removed the preferences ui portion of this, and added some hardcoded renamers to check in addition to the gsettings key. If there are any popular ones I missed let me know and I'll add them.

mbkma:
Thanks a lot! In a quick test everything works fine.

src/file-manager/fm-directory-view.c
```diff
@@ -6421,6 +6421,87 @@ real_action_redo (FMDirectoryView *view)
...
+        file = walk->data;
+        parameter = caja_file_get_uri (file);
+        quoted_parameter = g_shell_quote (parameter);
+        g_free (parameter);
+        cmd = g_string_append (cmd, " ");
+        cmd = g_string_append (cmd, quoted_parameter);
+        g_free (quoted_parameter);
+    }
+
+        guint i = 0;
```
mbkma, line 6490: this `i` can be in the for-loop

mtwebster: I cleaned up the legacy code a bit (I should in nemo also), though I'm not really keen on declaring i in the loop init - it's not consistent with the rest of caja (and personally I think it worsens readability, but I'm old and blind).

lukefromdc:
What is the status of this given the issues Mint had with Bulky?

mtwebster:
That issue isn't relevant - it was a problem with the pinned caja version in Mint, and its hardcoded use of bulky. I reviewed this PR since then to be sure, but it already guards against that problem.

lukefromdc:
What is the status of this? Are people still interested in working on it, should we close it, or something else?

lukefromdc:
Closing as no response from originator in over a year. We can reopen this if folks want to finish it

Pikku-a:
> What is the status of this? Are people still interested in working on it, should we close it, or something else?

> Closing for now as no response from originator in over a year. We can reopen this if folks want to finish it

May I ask what changes would be necessary for this pull request?

lukefromdc:
I really don't know, all I did on this was some testing as I've never used such a tool

lukefromdc:
Reopening as someone might be interested in working on it

lukefromdc:
Comments by others would need to be resolved

Pikku-a:
I tested this a bit and it seems to work well. At least I didn't notice anything to improve.

mtwebster:
The PR probably works because it's pointed at an older version of caja.

Quite a bit of refactoring has occurred since and it became no longer trivial to rebase this PR. I don't have any spare time to work on it currently.

lukefromdc:
OK, I'll close this for now. It can be reopened if folks want to work in it in the future, or a new PR can be submitted if major code changes from this are needed

PR #1554, opened 2021-09-18
fixed sort order in list view

DennisHeine:
sort order in list view was inverted


lukefromdc (approved):
Got a surprise with this: I was able to verify that on selecting "Date Modified" as the sort criterion the default sorting order of files (but not folders for some reason) was inverted with master as of today.
On applying this PR,  that default sorting order was fixed-but the order in which the columns appeared also changed, in this case with "Date Modified" being all the way to the left. That needs to be fixed

DennisHeine:
thanks alot, i changed the wrong file when editing.

PR #1557, opened 2021-09-21
fixed sort order (now the correct file)

decipher2k:
hopefully everything is working now

### DIFF
--- a/libcaja-private/caja-file.c
+++ b/libcaja-private/caja-file.c
@@ -1,4 +1,4 @@
-/*- Mode: C; indent-tabs-mode: t; c-basic-offset: 8; tab-width: 8 -*-
+/* -*- Mode: C; indent-tabs-mode: t; c-basic-offset: 8; tab-width: 8 -*-
 
    caja-file.c: Caja file model.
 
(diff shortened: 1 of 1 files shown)


lukefromdc (commented):
OK, this keeps the columns in the same order. When selecting list view and selecting sorting by date, both files and directories sort oldest to newest going down, by default. Not sure if this is the desired default (it is easily reversed by clicking the column header), but this does make directory and file sorting consistant with each other, which is important

lukefromdc:
Also just found a big problem: sorting by name in icon view is now reversed when arrange items>reverse order is NOT selected, and first to last when it is selected. That needs to be fixed. Same defaults in list view and compact view. We do NOT want reversed order to be shown when it is not selected by the user.

lukefromdc:
In fact, I have realized on further testing with directories not previously experimented with on this issue that with "reversed order" NOT selected (my normal setting) we seem to get both directories and files sorted oldest to newest going down, which is probably correct for that setting. Thus we should not change this at all.

Note that directories that have been open before in list view, once you have ever clicked on the column header twice save your setting, making this quite difficult to test.

The only thing I would change is removing "reversed order" as the global default for a new Caja install not configured by the user. I propose we close this and move on

DennisHeine:
all right. should i raise an issue concerning default settings?

lukefromdc:
That would be the correct process here

raveit65 (changes requested):
1. This breaks CI building with debian! https://app.travis-ci.com/github/mate-desktop/caja/jobs/538805035
2. Column order of creation date / access date / change date are sorted in the same way like nemo/nautilus or mozilla thunderbird did.
Looks like you misunderstood the meaning of the symbol.
In result there is no issue and no need for a code change.
Better close this bogus pull request.

zhuyaliang:
Keep the old code and don't need to make changes

PR #1567, opened 2021-10-20
caja-dnd: wrong type field in printf format string

rbuj:
```
caja-dnd.c:613:40: warning: format specifies type 'unsigned short' but the argument has type 'int' [-Wformat]
                            uri, x, y, w, h);
                                       ^
caja-dnd.c:613:43: warning: format specifies type 'unsigned short' but the argument has type 'int' [-Wformat]
                            uri, x, y, w, h);
                                          ^
```


mbkma (approved):
LGTM

PR #1568, opened 2021-10-21
Accessibility improvements: keyboard navigation and labelling

vvillenave:


raveit65:
Any news from this?

src/file-manager/fm-properties-window.c
```diff
@@ -564,6 +565,16 @@ create_image_widget (FMPropertiesWindow *window,
 				  G_CALLBACK (fm_properties_window_drag_data_received), NULL);
 		g_signal_connect (button, "clicked",
 				  G_CALLBACK (select_image_button_callback), window);
+
+		AtkObject * button_acc = gtk_widget_get_accessible (GTK_WIDGET (button));
+		gchar *icon_msg;
```
rbuj, line 570: The variable declaration should be at the top, just before `button = gtk_button_new ();`

rbuj:
I think we should wait to see what they say about the [issue opened at orca](https://gitlab.gnome.org/GNOME/orca/-/issues/218), otherwise we are adding a useless code (adding a blank space at the end of the string).

About using the identifier `icon_name` as a description, I don't like it, but it should be reviewed by someone who is more familiar with accessibility.

About adding a function that removes the underscores... i don't think it's really necessary, but I'm not against it, maybe there's a pango function that allows you to get the text rendered without mnemonics, or any other formatting text.

rbuj:
I don't think it's a good solution adding a blank widget to allow keyboard navigation.

src/file-manager/fm-properties-window.c
vvillenave: Done.

raveit65:
Any news from this?
Or is this ready to go?

rbuj:
Almost all the proposed changes are related to bugs which come from the third party software called orca. If we considered them useful, then we should be able to remove them when they are fixed in orca, so they need to be labeled with labels such as FIXME including the uri of the external bug. In addition you shoudn't split commits which make use of the new method called caja_remove_mnemonics.

At first glance I haven't seen any change since my last review, the improper use of the file name as a image description, adding a blank widget to allow navigation, adding a blank space at the end because of the recent changes in orca...

lukefromdc:
What is the current status of this? Have the problems been fixed in Orca? If not, do we want to mark all these changes w FIXME's to remove them if Orca gets updates to do the same job, close this, or something else?

lukefromdc:
Should we close this since it relates to trying to work around bugs in an old version of Orca?

lukefromdc:
Any updates on this?

lukefromdc:
Closing as it was replaced by https://github.com/mate-desktop/caja/pull/1810

PR #1569, opened 2021-10-21
caja-mime-actions: fix memory leak

rbuj:


mbkma (approved):
LGTM

lukefromdc (approved):
Builds fine, this file builds with no warnings at all. Runs fine, and clicking both on files that can and cannot be opened from caja works same as before. Easy way to test "There is no application installed..." is to click on a shared object-or a binary mistaken for one due to the known (and not from our packages) mimetype issue with PIE executables

Issue #1442, opened 2020-08-04
bad height size in sidebar popup menu


sc0w:
#### Expected behaviour

the size must be ok in sidebar popup menu

#### Actual behaviour

bad height size in sidebar popup menu

#### Steps to reproduce the behaviour

right click in one item (for example, cd)

![2020-08-05_00-42](https://user-images.githubusercontent.com/7734191/89352795-50b9cd00-d6b5-11ea-9071-8626f97841b5.png)

right click in other item (for example, file system)

![2020-08-05_00-43](https://user-images.githubusercontent.com/7734191/89352849-662ef700-d6b5-11ea-94f6-389ad71668b3.png)

the height is wrong, is too big, and it seems to be the same height as before (previous capture)

right click again in the same item (file system in this case)

![2020-08-05_00-43_1](https://user-images.githubusercontent.com/7734191/89353138-169cfb00-d6b6-11ea-92b6-005a14746167.png)

now the size is ok

so, the size in every popup menu is the right size of previous popup menu

#### MATE general version

1.24

#### Package version

1.24.0-1

#### Linux Distribution

debian bullseye/sid

lukefromdc:
Looks like this is same issue as
https://github.com/mate-desktop/caja/issues/1473
thus the same fix

PR #1571, opened 2021-10-22
caja-places-sidebar: fix right-click menu height

### LINKED ISSUE #1473: Size of 'Side Pane' pop-up (right click on item) is not correct
#### Expected behaviour
The pop-up should fit all the items and do not show scrollbar buttons at the top and bottom.

#### Actual behaviour
The pop-up sometimes has the correct size and sometimes not.

popup_size

#### MATE general version
1.24.1

#### Package version
Caja 1.24.0

#### Linux Distribution
Tested on Manjaro and Ubuntu MATE 20.10

#### Link to bugreport of your Distribution
Reproduced on 2 OSes so not distro-specific

Rakksor:
Prevents the places sidebar right-click menu from having an incorrect height when selecting an item with a different amount of menu entries. This should fix #1473 / #1442.

Note that this also changes the behavior of right-clicking on a different sidebar item when the menu is already open: Instead of only closing the menu, the menu of the new item is opened immediately. I think I prefer it this way but I don't know if this is desired.

### DIFF
--- a/src/caja-places-sidebar.c
+++ b/src/caja-places-sidebar.c
@@ -2892,7 +2892,7 @@ bookmarks_button_release_event_cb (GtkWidget *widget,
             return FALSE;
         }
 
-    	model = gtk_tree_view_get_model (tree_view);
+        model = gtk_tree_view_get_model (tree_view);
 
         gtk_tree_view_get_path_at_pos (tree_view, (int) event->x, (int) event->y,
                                        &path, NULL, NULL, NULL);
--- a/src/caja-places-sidebar.c
+++ b/src/caja-places-sidebar.c
@@ -2884,15 +2884,15 @@ bookmarks_button_release_event_cb (GtkWidget *widget,
     }
 
     tree_view = GTK_TREE_VIEW (widget);
-    model = gtk_tree_view_get_model (tree_view);
 
     if (event->button == 1)
     {
-
         if (event->window != gtk_tree_view_get_bin_window (tree_view))
         {
             return FALSE;
         }
+
+    	model = gtk_tree_view_get_model (tree_view);
 
         gtk_tree_view_get_path_at_pos (tree_view, (int) event->x, (int) event->y,
                                        &path, NULL, NULL, NULL);
@@ -2903,15 +2903,6 @@ bookmarks_button_release_event_cb (GtkWidget *widget,
     }
     else if (event->button == 3)
     {
-        GtkTreeModel *model;
-        GtkTreePath *path;
-        GtkTreeView *tree_view;
-
-        tree_view = GTK_TREE_VIEW (widget);
-        g_assert (tree_view == sidebar->tree_view);
-
-        model = gtk_tree_view_get_model (tree_view);
-
         gtk_tree_view_get_path_at_pos (tree_view, (int) event->x, (int) event->y,
                                        &path, NULL, NULL, NULL);
 
--- a/src/caja-places-sidebar.c
+++ b/src/caja-places-sidebar.c
@@ -2906,12 +2906,13 @@ bookmarks_button_release_event_cb (GtkWidget *widget,
         gtk_tree_view_get_path_at_pos (tree_view, (int) event->x, (int) event->y,
                                        &path, NULL, NULL, NULL);
 
-        gtk_tree_view_set_cursor(tree_view, path, NULL, FALSE);
-        bookmarks_popup_menu (sidebar, event);
-
         if (path != NULL)
         {
+            gtk_tree_view_set_cursor(tree_view, path, NULL, FALSE);
             gtk_tree_path_free (path);
+
+            bookmarks_popup_menu (sidebar, event);
+
             return TRUE;
         }
     }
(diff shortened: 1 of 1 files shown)


lukefromdc (commented):
This works, and entirely suppressed scrollbars in sidebar right-click menus as I iterated through every item trying to invoke the scrollbars/undersized menu. Exactly once, as an encrypted flash drive mounted, I was able to get a menu larger than needed, but that space was needed once the rest of the menuitems were present on reopening that menu.

One problem did come up: I got this build warning for an unused variable:

```
caja-places-sidebar.c: In function ‘bookmarks_button_release_event_cb’:
caja-places-sidebar.c:2906:23: warning: variable ‘model’ set but not used [-Wunused-but-set-variable]
```

src/caja-places-sidebar.c
```diff
@@ -2901,6 +2901,29 @@ bookmarks_button_release_event_cb (GtkWidget *widget,
 
         gtk_tree_path_free (path);
     }
+    else if (event->button == 3)
+    {
+        GtkTreeModel *model;
+        GtkTreePath *path;
+        GtkTreeView *tree_view;
```
rbuj, line 2908: These variables shadows others, which also already have the same value.

rbuj:
```
caja-places-sidebar.c: In function ‘bookmarks_button_release_event_cb’:
caja-places-sidebar.c:2906:23: warning: declaration of ‘model’ shadows a previous local [-Wshadow]
 2906 |         GtkTreeModel *model;
      |                       ^~~~~
caja-places-sidebar.c:2869:19: note: shadowed declaration is here
 2869 |     GtkTreeModel *model;
      |                   ^~~~~
caja-places-sidebar.c:2907:22: warning: declaration of ‘path’ shadows a previous local [-Wshadow]
 2907 |         GtkTreePath *path;
      |                      ^~~~
caja-places-sidebar.c:2868:18: note: shadowed declaration is here
 2868 |     GtkTreePath *path;
      |                  ^~~~
caja-places-sidebar.c:2908:22: warning: declaration of ‘tree_view’ shadows a previous local [-Wshadow]
 2908 |         GtkTreeView *tree_view;
      |                      ^~~~~~~~~
caja-places-sidebar.c:2870:18: note: shadowed declaration is here
 2870 |     GtkTreeView *tree_view;
      |                  ^~~~~~~~~
```

Rakksor:
Wow yes, I don't know what I was doing there, sorry for the mess. I added a second commit that should fix it.

src/caja-places-sidebar.c
```diff
@@ -2884,23 +2884,37 @@ bookmarks_button_release_event_cb (GtkWidget *widget,
...
+                                       &path, NULL, NULL, NULL);
+
+        gtk_tree_view_set_cursor(tree_view, path, NULL, FALSE);
+        bookmarks_popup_menu (sidebar, event);
+
+        if (path != NULL)
+        {
+            gtk_tree_path_free (path);
+            return TRUE;
+        }
```
rbuj, line 2916: Redundant NULL check before freeing the variable
```C
/**
 * gtk_tree_path_free:
 * @path: (allow-none): a #GtkTreePath-struct
 *
 * Frees @path. If @path is %NULL, it simply returns.
 */
void
gtk_tree_path_free (GtkTreePath *path)
{
  if (!path)
    return;

  g_free (path->indices);
  g_slice_free (GtkTreePath, path);
}
```

rbuj:
changing the desktop theme will trigger the issue to occur, so I think he doesn't fix it not fully

src/caja-places-sidebar.c
Rakksor: This was copied from the code handling middle-clicks in the `bookmarks_button_press_event_cb` callback. I assumed this is necessary to return `TRUE` in cases when no `GtkTreePath` can be constructed due to right-clicking on an empty area of the sidebar.

But I just noticed right-clicking on an empty sidebar area gives:
```
(caja:26121): Gtk-CRITICAL **: 13:26:08.438: gtk_tree_view_set_cursor_on_cell: assertion 'path != NULL' failed
```

This is fixed by putting `gtk_tree_view_set_cursor` inside this if clause, so I think the NULL check should remain. I'll update the pull request.

Rakksor:
I think the menu height after changing the desktop theme is a separate issue as it for example also happens with the menubar menus of every GTK3 application I tested.

rbuj (approved):
LGTM

src/caja-places-sidebar.c
```diff
@@ -2884,23 +2884,38 @@ bookmarks_button_release_event_cb (GtkWidget *widget,
...
 
     if (event->button == 1)
     {
-
         if (event->window != gtk_tree_view_get_bin_window (tree_view))
         {
             return FALSE;
         }
 
+        model = gtk_tree_view_get_model (tree_view);
```
mbkma, line 2895: I got the warning: `space before tab in indent` in this line. Could you fix this please?

Rakksor: This should already be fixed by the third commit (079102fcb47fad3d34ad98be9136863d0400b166)

raveit65: Seems OK.

raveit65 (approved):
LGTM

PR #1575, opened 2021-10-26
Remove whitespaces before #define, #if, and #endif

rbuj:


mbkma (approved):
It is okay but as I said, maybe we should run `clang-format` on all the code to get proper formatting in one single PR...

raveit65 (approved):
LGTM

PR #1576, opened 2021-10-26
caja-file: Remove unreachable code

rbuj:
reported by scan-build

### DIFF
--- a/libcaja-private/caja-file.c
+++ b/libcaja-private/caja-file.c
@@ -3922,7 +3922,6 @@ caja_file_get_boolean_metadata (CajaFile *file,
 		result = FALSE;
 	} else {
 		g_error ("boolean metadata with value other than true or false");
-		result = default_metadata;
 	}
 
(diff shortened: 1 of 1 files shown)


libcaja-private/caja-file.c
```diff
@@ -3922,7 +3922,6 @@ caja_file_get_boolean_metadata (CajaFile *file,
 		result = FALSE;
 	} else {
 		g_error ("boolean metadata with value other than true or false");
```
mbkma, line 3924: It is not obvious that the `else` never is reached. Anyway, this PR does not solve the issue, because it just lets the return value uninitialised...

rbuj: g_error calls to abort() for terminating the application, and it will result in a core dump, so the instruction below is never run

mbkma: Thanks, I never knew that.

mbkma (approved):
LGTM

PR #1585, opened 2021-11-15
Show non-generic mime type icon where applicable

### LINKED ISSUE #795: Wrong mime type icon in Icon View and Compact View
#### Expected behaviour
My python scripts in my folder should show the set `~/.icons/sable-ultra-flat-icons/mimetypes/scalable/gnome-mime-text-x-python.svg` symbol in Icon View.

#### Actual behaviour
Shows some standard icon which is also used by normal text files. Might be "text-x-generic.svg" "text-plain.svg" "text.svg" "text-x.svg" or maybe even some other one which I overlooked in my icon folder.

Open folder in Icon View

#### MATE general version
Not using Mate, using XFCE 4.12.1. Caja 1.16.2

#### Package version
Caja 1.16.2

#### Linux Distribution
Debian 4.5.1-1

#### Note
The correct icon is shown only in `List View`. This suggests the icon path or Mime-Type configuration is not at fault here.

monnerat:
When the target mime type icon is in 'hicolor', themed generic icon is
currently preferred.

The reason is: the gThemedIcon for the file mentions the generic icon
name as the last alternative. Themed icon name selection prefers it over
the others because it is always in the theme, while other are not.

The solution implemented here is to reduce the gThemedIcon names to the
first usable one only, in hope the generic icon name will be dropped. If
no usable icon is found, the original gThemedIcon is kept unchanged.

Fixes #795

Thanks for considering it.

### DIFF
--- a/libcaja-private/caja-file.c
+++ b/libcaja-private/caja-file.c
@@ -2431,6 +2431,25 @@ update_info_internal (CajaFile *file,
 	}
 
 	icon = g_file_info_get_icon (info);
+	g_object_ref (icon);
+	if (G_IS_THEMED_ICON (icon)) {
+		/* Only keep the first icon name found in theme, in hope to
+		   drop the generic icon name. */
+		GtkIconTheme *icon_theme = gtk_icon_theme_get_default ();
+		const gchar * const *names = g_themed_icon_get_names (G_THEMED_ICON (icon));
+		int i;
+
+		if (names) {
+			for (i = 0; names[i]; i++) {
+				if (gtk_icon_theme_has_icon (icon_theme, names[i])) {
+					GIcon *new_icon = G_ICON (g_themed_icon_new_from_names ((char **) &names[i], 1));
+					g_object_unref (icon);
+					icon = new_icon;
+					break;
+				}
+			}
+		}
+	}
 	if (!g_icon_equal (icon, file->details->icon)) {
 		changed = TRUE;
 
@@ -2439,6 +2458,7 @@ update_info_internal (CajaFile *file,
 		}
 		file->details->icon = g_object_ref (icon);
 	}
+	g_object_unref (icon);
 
 	thumbnail_path =  g_file_info_get_attribute_byte_string (info, G_FILE_ATTRIBUTE_THUMBNAIL_PATH);
 	if (eel_strcmp (file->details->thumbnail_path, thumbnail_path) != 0) {


rbuj:
At first glance, you are loading the first match without knowing if it's the valid one by using your custom icon resolver, that is, if you have the icon of any subtype for any generic icon, the subtype icon is used even if it doesn't match with the file content.

monnerat:
> At first glance, you are loading the first match without knowing if it's the valid one by using your custom icon resolver, that is, if you have the icon of any subtype for any generic icon, the subtype icon is used even if it doesn't match with the file content.

First, all names are supposed to be valid, in terms of file representation. The only thing that may later 'invalidate' a name is its 
absence within the theme (note that, in this context, `hicolor` icons are implicitly part of the theme).

Next, `gThemedIcon` names are listed in preference order: generic and symbolic icons being at end of list. The problem is: the gtk selection algorihm does not respect this order if some trailing name can be found in the theme icon **directory** while leading ones only stand in `hicolor` directory.
The filter implemented by this patch only limits the name list to the first (always valid) usable name, that's all. In particular, it does not affect the way the file type is determined.
The file icon is later processed in `caja_file_get_gicon`() before actual display.

As this is already the case, the file content is not used to determine the mime type (and thus the icon) if the file name extension already selects an existing mime type. I tried to set a mime type for extension `.aaa` with its own icon in `hicolor` and store C code in file `test.aaa`: unpatched Caja shows it with a text icon while patched Caja shows the aaa icon. If I remove `aaa` mime type definition, both patched and unpatched Caja show it with a C code icon.

Of course, you can still override a particular file's icon with a custom one!

I hope this helps you understand this commit. If you have some counterexample that misses the proper icon with the patched Caja only, please post here. Thanks.

lukefromdc (commented):
Builds fine and runs on my system without visible changes, presumably given my icon set. How should I properly test this?

monnerat:
> How should I properly test this?

I guess it depends on your distro and packages you've installed.
Maybe you can try follow https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/using_the_desktop_environment_in_rhel_8/configuring-file-associations_using-the-desktop-environment-in-rhel-8#adding-a-custom-mime-type-for-all-users_configuring-file-associations then create icon file `/usr/share/icons/hicolor/32x32/mimetypes/application-x-newtype.png` with the image of your choice: a non-empty file named `foo.xyz` should then appear with your icon. even in Caja icon view.

cwendling:
IIUC what you're trying to do is rather use a specific icon outside the theme (well, inside the hierarchy, but possibly outside the final theme) over a more generic icon but from the specific theme, right?  I can see upsides and downsides from both approaches: the hicolor icon might be specific but look overly out of place in the theme, while the specific icon might be better at identifying the type than the generic one.
My balance doesn't lean either way really, but it might be worth having the same logic everywhere, including in Caja, file chooser and other places using the generic API.  I don't know what the spec says about this, but I feel like this change should rather go higher in the stack, e.g. in GTK and/or GIO.

This said, IIUC it's always possible to work around either direction by adding the specific icon in the specific theme: if you symlink your specific MIME icon in the current theme it'll be used just fine, and it should as well be if linked under `~/.icons/`, shouldn't it?

monnerat:
> it's always possible to work around either direction by adding the specific icon in the specific theme: if you symlink your specific MIME icon in the current theme it'll be used just fine, and it should as well be if linked under ~/.icons/, shouldn't it?

Yes, but it should not be needed as `hicolor` is a standard feature.
On a practical point of view, a typical application package does not know what themes are installed on your system and thus has no other choice than leaving its icons in `hicolor`. The solution you propose requires a snippet to be executed at application install/uninstall time to handle these symlinks in each theme present. Even in this case, a theme installed **after** the application will not have these icons :-(

If this tuning is manual, it must be done by an administrator aware of the newly installed application package contents. This process is hard to generalize and almost impossible to automate. In addition, this is out of normal user control and requires skills/knowledge that most users and/or administrators don't have.

> My balance doesn't lean either way really, but it might be worth having the same logic everywhere, including in Caja, file chooser and other places using the generic API. I don't know what the spec says about this, but I feel like this change should rather go higher in the stack, e.g. in GTK and/or GIO.

The real problem comes from GIO function `g_content_type_get_icon`() that is responsible to gather possible names from the mime type definition, but also always includes the generic icon name. This is probably legitimate, as it is listed last.
However the GTK selection algorithm puts more priority for presence in theme directory than for name position within the gThemedIcon names list, resulting in what we have today, breaking the standard `hicolor` feature. I haven't found any doc at freedesktop.org about alternate icon names: it seems to be a GLib-specific non-standard feature.

My feeling is Gtk/Glib maintainers won't change anything about this. What I can tell you is: Qt-based browsers (tried with Dolphin) do show the type icon, while Gio-based ones (Caja, Nautilus, Nemo) show the generic icon. The exception is Thunar that implements the same kind of filtering as done by the current PR.

monnerat:
Note also that some application(s) have found a workaround: declare the generic icon in mime-info file with the name of the non-generic icon. This works but breaks the standards, as the mime-info DTD limits the generic-icon name to a list of possible values (xmllint will fail):
```
<!ELEMENT generic-icon EMPTY>
<!ATTLIST generic-icon name (application-x-executable | audio-x-generic | folder | font-x-generic | image-x-generic | package-x-generic | text-html | text-x-generic | text-x-generic-template | text-x-script | video-x-generic | x-office-address-book | x-office-calendar | x-office-document | x-office-presentation | x-office-spreadsheet) #REQUIRED>
```

For this reason it should not be used.

See https://github.com/OpenShot/openshot-qt/blob/develop/xdg/org.openshot.OpenShot.xml

lukefromdc:
This is a bit out of my understanding, so I will have to leave that for someone else
to test. Note that I didn't get any problems testing this on a running system, but 
of course that's a snapshot of one possible system

rbuj:
### Steeps to reproduce
Create new mime-type:
```
cat <<EOF >>shinythings-shiny.xml
<?xml version="1.0"?>
<mime-info xmlns='http://www.freedesktop.org/standards/shared-mime-info'>
  <mime-type type="text/x-shiny">
    <comment>Shiny new file type</comment>
    <glob pattern="*.shiny"/>
    <glob pattern="*.shi"/>
  </mime-type>
</mime-info>
EOF
$ sudo xdg-mime install shinythings-shiny.xml
```
Install the icons for the new mime-type on the current icon theme:
```
$ gsettings get org.mate.interface icon-theme
'mate'
sudo ln -s /usr/share/icons/hicolor/24x24/apps/mate-panel-fish.png /usr/share/icons/mate/24x24/mimetypes/text-x-shiny.png
sudo ln -s /usr/share/icons/hicolor/32x32/apps/mate-panel-fish.png /usr/share/icons/mate/32x32/mimetypes/text-x-shiny.png
sudo ln -s /usr/share/icons/hicolor/48x48/apps/mate-panel-fish.png /usr/share/icons/mate/48x48/mimetypes/text-x-shiny.png
```
Create a test file:
```
echo "Hello world" > test.shiny
```
The icon is displayed correctly in the list view, but it isn't displayed in the icon view.

monnerat:
> The icon is displayed correctly in the list view, but it isn't displayed in the icon view.

You may need to restart Caja to see the results.
With the unpatched Caja, you get a generic icon in icon view. With the patched Caja, the fish icon appears.
This occurs although icons are in theme directory, not in `hicolor`: I guess this time it is an icon size priority problem, but its only a supposition. Fortunately this is also fixed by the patch.

rbuj:
@monnerat I only describe briefly the issue you are trying to fix and the steps to reproduce it.

I will try debug the code, perhaps between today and tomorrow, both caja and glib, to see where the problem comes from and if your pull request fixes the issue and it doesn't create new ones.

Note when you add the new iteration for solving the icon name, you are increasing the computational complexity of the search. Before that cost could be constant O(1) if a hash table is used, O(n) if you use another LUT using iterative search. For this reason I think we should try to keep complexity low. Besides, the mime type obtained in the following lines can be used for getting the icon instead of doing an iterative search of the mime types associated with the icon, which in the case of text files is very high.

monnerat:
> I only describe briefly the issue you are trying to fix and the steps to reproduce it.

This is what I've understand and I thank you for that. My comment was just a complement of information.

> Note when you add the new iteration for solving the icon name, you are increasing the computational complexity of the search. 

This is done very early in the process and reduces the name count to a single one, effectively reducing the complexity of all posterior uses of the gicon associated with the file.

> the mime type obtained in the following lines can be used for getting the icon instead of doing an iterative search of the mime types associated with the icon, which in the case of text files is very high.

There's no additional mime-type determination added by this patch: only cache table lookups in `gtk_icon_theme_has_icon`(). The mime type and the icon are already set in the `gFileInfo` upon `update_info_internal`() entry. The latter only uses them without recomputing.

rbuj:
It displays the icons correctly. (test: open a folder with some C source files)

Maybe it would be nice to use `gtk_icon_theme_lookup_icon_for_scale` since we use the flag GTK_ICON_LOOKUP_FORCE_SIZE, or make the selection of the available/preferred themed icon in libcaja-private/caja-icon-info.c, I do not know if it could be done in caja_icon_info_lookup.

monnerat:
> Maybe it would be nice to use gtk_icon_theme_lookup_icon_for_scale since we use the flag GTK_ICON_LOOKUP_FORCE_SIZE

I don't think it's a good idea, or at last feasible: neither the size nor the scale is known at this time. In addition, the gicon stored here is reused at many places to generate icons of different sizes (list view, icon view, compact view, file properties and maybe others if some).

> make the selection of the available/preferred themed icon in libcaja-private/caja-icon-info.c, I do not know if it could be done in caja_icon_info_lookup.

This means you will reexecute the filter each time you perform a lookup, which seems useless. Doing it upstream is consistent as we NEVER want to have the generic name in the gThemedIcon for files that do have a specific icon: generic names are handled by Caja itself (maybe a bit simplistic, but that's another problem), thus we do not need to consider them twice. In addition, `caja_icon_info_lookup()` can be called for non-file icons.

rbuj:
The themed icons are displayed as expected in other places, so the problem isn't in the modified function. Replacing the icon list for a themed icon with just one item is not a good idea.

Captura de pantalla 2021-11-18 a les 10 39 36

rbuj:
All themed icons are displayed as expected after disabling the preview for text files, setting `Show text in icons` to `Never`.

Captura de pantalla 2021-11-18 a les 13 48 09

monnerat:
@rbuj : I guess your tests resulting in the previous 2 comments were performed with icons in `mate` theme, not `hicolor`. I tried this here (with `hicolor`) without success.

>  Replacing the icon list for a themed icon with just one item is not a good idea.

So what would you reommend to achieve the same result with `hicolor` ?
The only alternative I can imagine is:
- Get the generic icon name separately (this adds some overhead: mime info query on the known content-type) and store it in the CajaFile. This step is needed because there is no way to determine from a name if it is generic or not +
- Filter out only the generic name in the file icon +
- Use the stored generic name as a fallback icon.

... but this has a (somehow low) performance cost.

rbuj:
I think you will need to check how the icon resolution method works and whether you can change the resolution order. I think it first does a search in the current icon-theme and then in hicolor, for this reason it solves the generic since it exists in the current icon-theme so it doesn't actually search in hicolor. However, I think this is a different issue to the linked one.

lukefromdc:
What is the status of this? Should it be closed or are folks still working on it?

monnerat:
> What is the status of this?

Stalled.

I give up: in fact there are two problems:

1. Text icons: they're intended to be filled with the file text itself, but nothing appears and the icon stays blank. I would have preferred to use the real application's icon and a tooltip for the text rather than a unreadable preview, if it worked.
2. Theme takes precedence over icon being generic or not: this results in having almost only generic icons in caja if the corresponding applications do not provide a specific one for the active theme (which is the majority) ---> hicolor is masked off by all theme's generic icons. This is primarily a gtk caveat and a caja-specific workaround without stripping the generic icon would be really cumbersome. It should be noted that neither Qt-based file browsers nor Thunar (it strips the generic name) have this problem.

In addition it seems there is a strong opposition in finding a reasonable solution. I would even say nobody cares of this problem.

rbuj:
This pull request doesn't fix it doesn't fix any real issues #795. The author just disagrees on icon search priority, which wants to discard the icon theming or give it a lower priority.

rbuj (changes requested):
The theme folder should take precedence over hicolor folder. It doesn't fix the issue #795, as the icons for the user-defined mime types work without any issues, see  https://github.com/mate-desktop/caja/pull/1585#issuecomment-970697871

PR #1591, opened 2021-12-17
Use gresources for ui files

rbuj:


lukefromdc (approved):
Builds and runs fine, preferences and bookmarks windows work same as before. This makes caja theoretically faster to load, since loading these a gresources means they are compiled into the binary rather than read separately from disk. The slower the disk, the bigger the advantage but everyone benefits.

lukefromdc:
Merging as we've never had a problem with the gresource updates and we have a LOT of open PR's

PR #1593, opened 2021-12-20
Fix some -Wsign-compare warnings

rbuj:


mbkma (approved):
LGTM. Conversion fixes looks good and the removed check
```
    if (xml_len == -1)
    {
        xml_len = strlen (xml);
    }
```
can be safely removed because `xml_len` is of type `gsize` aka `unsigned long`, hence guaranteed to be >0.

lukefromdc (approved):
Builds and runs fine, note that we still get quite a few warnings about comparing signed to unsigned variables but fixing some of them is a good start

Issue #689, opened 2016-12-02
Segfault when browsing Windows Network if a server has an empty name


dougg3:
I have tested on Ubuntu MATE 16.04 with caja 1.12.7 as well as Ubuntu MATE 16.10 with caja 1.16.0.

If a Samba server on the local network has an empty name, caja will segfault in eel_ref_str_ref. To reproduce:

1. On a different machine, install samba and manually set its NetBIOS name to an empty string by editing /etc/samba/smb.conf and adding this line under the [global] section:
    * `netbios name = ""`
2. On the other machine, restart nmbd:
    * `sudo service nmbd restart`
3. On your computer, run `smbtree -N` to verify that the other machine has an empty name. You should see a line like this, with emptiness after the \\\\:
    * `\\ ubuntu-mate-2 server (Samba, Ubuntu)`
4. Now, on your computer, browse to the Windows network in caja by going to Places -> Network and opening the "Windows Network" folder.
5. Open the folder corresponding to the workgroup that the other machine belongs to.
6. caja will segfault.

Here is the top of the stack trace:
    eel_ref_str_ref ()
    ?? ()
    caja_file_new_from_info ()
    ?? ()
    ?? ()

It's not necessarily uncommon for a server to have an empty name; [nmbd used to have a bug](https://bugzilla.samba.org/show_bug.cgi?id=10896) that would set the default provided NetBIOS name to an empty string if the computer's hostname was too long.

This exact same problem used to occur in Nautilus until it was fixed last year. [See this bug report and patch](https://bugzilla.gnome.org/show_bug.cgi?id=700507). It appears that the crash in caja manifests itself in a slightly different way, though, because naively applying the analogous fix to caja didn't seem to fix the problem when I tried it.

kn00tcn:
i was thinking 'if empty string, insert name as 'untitled server' for example, but... what happens when multiple servers have the same name or same no name?

other DEs or distros or file managers or even browsers handle this fine?

dougg3:
Based on my limited testing, if you have two servers with the same name, only one will appear in the list. This is repeatable in Mac OS X, Windows, and caja's network browsers, as well as smbtree -N.

[...]

PR #1600, opened 2022-01-16
Fix crash while browsing Windows Network

dougg3:
This commit fixes two problems that each cause a crash while browsing the Windows Network if a computer on the network has an empty name. See #689 for more detail about how the problem happens. It sounds pretty weird for a computer to have an empty name, but it's definitely possible -- [nmbd used to have a bug](https://bugzilla.samba.org/show_bug.cgi?id=10896) that caused the autogenerated NetBIOS name to be empty if the computer's hostname was too long.

Both of these changes are basically just bringing in small fixes that have already been in Nautilus for years.

### DIFF
--- a/libcaja-private/caja-file.c
+++ b/libcaja-private/caja-file.c
@@ -2534,7 +2534,8 @@ update_info_internal (CajaFile *file,
 				(file->details->directory, file);
 
 			g_clear_pointer (&file->details->name, g_ref_string_release);
-			if (eel_strcmp (file->details->display_name, name) == 0) {
+			if (file->details->display_name &&
+			    strcmp (file->details->display_name, name) == 0) {
 				file->details->name = g_ref_string_acquire (file->details->display_name);
 			} else {
 				file->details->name = g_ref_string_new (name);
@@ -4099,7 +4100,7 @@ caja_file_peek_display_name (CajaFile *file)
 		}
 	}
 
-	return file->details->display_name;
+	return file->details->display_name ? file->details->display_name : "";
 }
 
 char *


lukefromdc:
Someone who has access to a Windows machine will need to test this, I don't have one

dougg3:
All the testing I did was without Windows -- I just pulled up a few Ubuntu MATE VMs and installed Samba on one of them. The easiest way to make an empty server name show up is to set up Samba with `netbios name = ""` in /etc/samba/smb.conf.

lukefromdc:
OK, that lets more folks here work on it, though in my case I also do everything on bare metal and have never learned to work with VM's. I have never used Samba so not familiar with it

dougg3:
Makes sense -- I will post detailed test instructions here for reproducing the problem.

dougg3:
Here are instructions to reproduce the crash. Unfortunately, Windows networking can be extremely finicky and its behavior often depends on what other devices are on your LAN, so it's best to test this with a pair of VMware VMs set up with NAT networking so they're isolated from the rest of your network but can still talk to each other. Note that NAT won't work with VirtualBox, they are kept isolated in VirtualBox. If you're a VirtualBox expert, you may be able to figure out how to do it here.

One VM will be the server, configured to output an empty server name. The other one will be the client that crashes.

### Set up the server

- Create a VMware virtual machine to boot up an [Ubuntu MATE 16.04 ISO](https://cdimage.ubuntu.com/ubuntu-mate/releases/16.04/release/ubuntu-mate-16.04.6-desktop-amd64.iso). I had trouble reproducing the problem if I used a newer version of Ubuntu as my Samba server -- I think things have changed with Samba more recently. This VM will act as a Samba server that has an empty name. Set it for NAT networking.
- Just boot it up in "Try Ubuntu MATE" mode.
- In the VM, run the following command: `sudo apt update ; sudo apt install samba`
- In the VM, edit /etc/samba/smb.conf. Just under [global], add the following lines, which set the name to be empty and force it to be the master browser:
  - `netbios name = ""`
  - `local master = yes`
  - `os level = 100`
  - `domain master = no`
  - `preferred master = yes`
- In the VM, run the following commands:
  - `sudo service smbd restart`
  - `sudo service nmbd restart`

The server is now ready to go. Keep it running.

### Set up the client

- Create another virtual machine to boot an Ubuntu MATE live CD (more modern version such as [21.10](https://cdimage.ubuntu.com/ubuntu-mate/releases/21.10/release/ubuntu-mate-21.10-desktop-amd64.iso)). This VM will be the client that demonstrates the Caja crash, so the newer the better. Also set this VM for NAT networking.
- Boot up this VM in "Try Ubuntu MATE" mode as well.
- In this VM, open up Caja. Go to the "Go" menu and choose Network.
- Usually if it's going to work, you'll see two items: smb-server- and Windows Network. If they both don't show up, restart the VM and try again (yes, seriously, that is how messed up Windows networking is...)
- Open up Windows Network. If you're lucky, it'll actually open.
- There should be an item called WORKGROUP inside of there. Open it, wait a sec, and Caja should crash.
- If you can't get Windows Network to show up, try rebooting this VM and trying again. Like I said, Windows networking is really finicky.

If you apply the fix in this pull request and run the patched version of Caja in the VM, it will no longer crash. The setup for this bug might sound crazy and convoluted, but this situation does happen. I experienced it myself while at work.

lukefromdc:
I am indeed going to have to leave this for others on the team: this is completely out of my league and with no landline I cannot even download the disc images.

raveit65:
Which samba version does old  Ubuntu MATE 16.04 use?
Looks like 16.04 is EOL.
https://news.itsfoss.com/ubuntu-16-04-reaches-end-of-life/
I am using modern fedora but i will try reproduce the issue with an EOL version of fedora.

rbuj:
I'm not sure the NetBIOS name can be set to "". In such case, the NetBIOS name comes from the hostname, which cannot be null. Anyway, it looks like a user mistake or, a name resolution error. Did you fill a bug report on samba?
https://www.oreilly.com/library/view/using-samba/1565924495/ch01s03.html#ch01-SECT-3.2

dougg3:
Yeah, 16.04 is pretty old. It's using Samba 4.3.11. FWIW, an 18.04 VM as the server (Samba 4.7.6) also reproduces the problem.

I can't reproduce the problem with an Ubuntu MATE 20.04 VM as the server - it seems like on newer servers, the "WORKGROUP" icon never shows up inside of Windows Network. I'm not sure why.

However, the underlying problem isn't necessarily just a Samba thing though -- this is simply the only way I know of to cause a NULL details->display_name to happen.

dougg3:
If you look at the link I posted in the original message, there was a problem on older versions of Samba that caused the NetBIOS name to be empty if the computer's hostname was too long. https://bugzilla.samba.org/show_bug.cgi?id=10896

That bug is how I originally ran into this problem a few years ago.

rbuj:
There is not segfault on Fedora 34. Therefore, it isn't reproducible anymore.
The samba bug about the long hostnames was fixed in 2014.

Screenshot at 2022-01-31 09-26-12

```
[robert@localhost ~]$ sudo cat /etc/samba/smb.conf
# See smb.conf.example for a more detailed config file or
# read the smb.conf manpage.
# Run 'testparm' to verify the config is correct after
# you modified it.
#
# Note:
# SMB1 is disabled by default. This means clients without support for SMB2 or
# SMB3 are no longer able to connect to smbd (by default).

[global]
	workgroup = SAMBA
	security = user

	passdb backend = tdbsam

	netbios name = ""

	printing = cups
	printcap name = cups
	load printers = yes
	cups options = raw

[homes]
	comment = Home Directories
	valid users = %S, %D%w%S
	browseable = No
	read only = No
	inherit acls = Yes

[printers]
	comment = All Printers
	path = /var/tmp
	printable = Yes
	create mask = 0600
	browseable = No

[print$]
	comment = Printer Drivers
	path = /var/lib/samba/drivers
	write list = @printadmin root
	force group = @printadmin
	create mask = 0664
	directory mask = 0775

[share]
        comment = My Share
        path = /home/robert/share
        writeable = yes
        browseable = yes
        public = yes
        create mask = 0644
        directory mask = 0755
        write list = user
```
The name can be resolved on the local server, but it cannot be resolved on the client.
From my point of view, setting an empty NetBIOS name is a user mistake. What's more, no double quotes are used to define the **netbios name** in smb.conf.

libcaja-private/caja-file.c
```diff
@@ -2534,7 +2534,7 @@ update_info_internal (CajaFile *file,
 				(file->details->directory, file);
 
 			g_clear_pointer (&file->details->name, g_ref_string_release);
-			if (eel_strcmp (file->details->display_name, name) == 0) {
+			if (g_strcmp0 (file->details->display_name, name) == 0) {
```
cwendling, line 2537: Why this change?  I'm not against it as `g_strcmp0()` is more standard, but both work on `NULL`, yet they have slightly different behavior: `eel_strcmp()` considers `NULL`s equal to `""` while `g_strcmp0()` sorts `NULL` before anything else.

If the goal is not to use a `NULL` `file->details->display_name`, maybe explicit that?

cwendling (commented):
Looks good to me if that's the only problem with `NULL`s.  And it seems a good idea to consolidate this area.  Even if it's caused by a buggy Samba server, it should not crash Caja.

libcaja-private/caja-file.c
dougg3: I was wondering if this would be a problem, because there are a bunch of other usages of `eel_strcmp()` that I didn't touch and it felt weird to only change that one spot. Nautilus already changed across the board to `g_strcmp0()`, so it wasn't affected by this part of the issue. You're absolutely right, the goal is to ensure we don't end up with a `NULL` `display_name`, because that results in an assertion failure later. I will change to explicitly check for NULL instead, and leave `eel_strcmp()` alone. Thanks for the review!

dougg3:
I agree. Yes, it's a weird setup, and it was probably a bigger problem 5 years ago than it is now, but it's still conceivable that I could plug my computer into an old Samba network and Caja would crash. I'm sorry that the test setup is so convoluted, but Samba networking in general is complicated. If it helps at all, I have attached videos showing Caja's behavior before and after this fix.

**Before:**
https://user-images.githubusercontent.com/1565758/152090612-5322fd81-5208-4235-bd64-bed773eef5ed.mp4

**After:**
https://user-images.githubusercontent.com/1565758/152090618-9062c73f-d526-4717-b40e-bb8e59f8596e.mp4

libcaja-private/caja-file.c
```diff
@@ -2534,7 +2534,8 @@ update_info_internal (CajaFile *file,
 				(file->details->directory, file);
 
 			g_clear_pointer (&file->details->name, g_ref_string_release);
-			if (eel_strcmp (file->details->display_name, name) == 0) {
+			if (file->details->display_name &&
+			    eel_strcmp (file->details->display_name, name) == 0) {
```
rbuj, line 2538: `name` isn't NULL, since the previous line `strcmp (file->details->name, name) != 0) {` doesn't trigger a SEGV, so you can use `strcmp` instead of `eel_strcmp`.

Please note that I cannot approve this change until I have verified that it solves an actual issue on a supported os, otherwise you may introduce a software regression or an incompatibility with current libs..

dougg3: Excellent point, I changed to just using strcmp.

dougg3:
I'm not really sure what else to say or do at this point. If I'm on the same Windows network as someone who has a broken Samba server with an empty name, Caja will crash when I browse the workgroup because the name ends up being NULL instead of an empty string. In my opinion, crashes are bugs that should be fixed. I've provided details instructions for how to simulate such a broken Windows network, and a video of the crash in action.

I can understand concerns about introducing regressions, but the change I'm proposing is essentially just guarding against NULL pointers in two places.

dougg3:
Closing because the crash that this pull request solves is no longer reproducible on modern Samba networks (see discussion in #689).

PR #1601, opened 2022-01-16
eel-editable-label: drop _clear_layout, _get_font_description

rbuj:


lukefromdc (approved):
eel-editable-label.lo builds with no warnings at all with this applied and my default build options. Build finished fine, and on installing and restarting caja everything seems to work same as before. Renaming files works as expected, and fonts looked the same as before, at least with what I have installed

raveit65 (approved):
LGTM

PR #1615, opened 2022-02-10
caja-bookmark: drop create_image_cairo_for_bookmark

rbuj:


lukefromdc (approved):
caja-bookmark.lo built clean without warnings with this. On installation of the finished build, adding, removing, and opening bookmarks from the "bookmarks" menu item or the side pane worked as expected

vkareh:
works for me

PR #1622, opened 2022-03-15
caja-application: Variable is assigned a value that is never used

rbuj:


mbkma (approved):
LGTM, `exit_with_last_window`  is a local variable and reducing the scope is a good thing.

lukefromdc (approved):
This should have absolutely NO effect on functionality, but prevent a value from being stored that is never used, thus saving the CPU cycles necessary to copy it into RAM.

At any rate, caja works normally with this so far as I can tell without trashing my session and starting a new one. Setting exit_witth_last_window does work, note that you have to turn icons on the desktop OFF for this to work. Years ago, this setting could cause the session manager to repeatedly restart caja over and over again, I have forgotten if a fix for that loop was ever found. Either way this PR would not change that behavior so no reason not to merge it

Issue #1630, opened 2022-05-18
(Hidpi) Crash w segfault on confirmation in file conflict dialog


lukefromdc:
#### Expected behaviour
When replacing a file by moving or copying from another directory with another file of the same name, thus invoking the file conflict dialog,, clicking "replace" should replace the file and caja should keep running normally

#### Actual behaviour
Clicking "replace" causes caja to crash instantly, then restart as normal in a MATE session. On opening the file, it is clear the file replacement has succeeded. 

#### Steps to reproduce the behaviour
Copy any short file (short is all I tested) to another directory, then move or copy it back to the original directory

#### MATE general version
1.26

#### Package version
locally compiled git master current as of March 22 (no change since then according to git pull)

#### Linux Distribution
Debian Unstable updated on or after May 15 2022

#### Link to bugreport of your Distribution (requirement)
None as this is a locally compiled package

lukefromdc:
I got this backtrace by turning off caja in session>required_components and running caja under gdb in a new session:
```

Thread 1 "caja" received signal SIGSEGV, Segmentation fault.
0x00007ffff787cc96 in gtk_widget_get_scale_factor () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
(gdb) thread apply all bt full

Thread 63 (Thread 0x7fffcf7fe640 (LWP 108856) "pool-caja"):
#0  syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
#1  0x00007ffff6f35322 in g_cond_wait_until () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#2  0x00007ffff6eb3671 in  () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#3  0x00007ffff6eb3c52 in g_async_queue_timeout_pop () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#4  0x00007ffff6f0d669 in  () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#5  0x00007ffff6f0cd0d in  () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#6  0x00007ffff7f52d80 in start_thread (arg=0x7fffcf7fe640) at pthread_create.c:481
        ret = <optimized out>
        pd = 0x7fffcf7fe640
[...]

PR #1633, opened 2022-06-01
Properly initialize IconKey::scale

cwendling:
This value is used to look up icons in the cache, but somehow was not properly initialized.  It was introduced in cfd502157420406aa56a6b5319f0bd8dc90df8e4 but lost in refactoring in 6c423bc27d2803f3259bcca9f518a1c22b2b5fd1.

Might or might not be related to #1630.


rbuj (approved):
LGTM

PR #1635, opened 2022-06-06
Document and fix default-sort-order preference

thesquash:
This patch resolves three related issues:

The first issue was that the GSettings schema for Caja did not include an entry to sort by the "btime", or creation date, of files.  If the user chose such an option in the Caja Preferences, GSettings would produce a warning (often out-of-sight, as it was usually redirected into the user's .xsession-errors file), and Caja would not actually change the default sort-order of files.  This patch adds the btime as a valid setting in the schema.

The second issue was that, because of the above (an entry in the settings schema was missing), some of the alternative sort orders listed in the schema (everything after and including "atime") were not assigned the same numbers as the sort orders listed elsewhere in the Caja source code.  Specifically, in icon- and compact-views, if the default sort-order was "emblems", the observed / actual sort-order would be the entry before "emblems", namely "atime" -- so instead of sorting by the names of associated emblems, Caja would sort by each file's access time.  An array in the code for the list-view also was missing many values and included some values out of order, so the default sort-order setting affected directories viewed in list-view mode seemingly randomly.  The former is taken care of using the fix described in the above paragraph; the latter is fixed in this patch by adding /
reorganizing the array for the list-view sort-orders appropriately.

The third issue (admittedly, a lesser issue) was that the documentation for the default-sort-order setting was lacking -- it at least did not list all the possible values that the setting could accept.  In this patch, I resolve this issue by listing all values in the setting's description, and also go into more detail about what each value does.  (However, perhaps I included a little too much detail.  It'll only benefit [supposed power users who may already know this stuff] who use GSettings or DConf directly, and it'll certainly be a headache for translators.  I'll admit that.)

### DIFF
--- a/libcaja-private/org.mate.caja.gschema.xml
+++ b/libcaja-private/org.mate.caja.gschema.xml
@@ -35,11 +35,12 @@
     <value nick="size" value="3"/>
     <value nick="type" value="4"/>
     <value nick="mtime" value="5"/>
-    <value nick="atime" value="6"/>
-    <value nick="emblems" value="7"/>
-    <value nick="trash-time" value="8"/>
-    <value nick="size_on_disk" value="9"/>
-    <value nick="extension" value="10"/>
+    <value nick="btime" value="6"/>
+    <value nick="atime" value="7"/>
+    <value nick="emblems" value="8"/>
+    <value nick="trash-time" value="9"/>
+    <value nick="size_on_disk" value="10"/>
+    <value nick="extension" value="11"/>
   </enum>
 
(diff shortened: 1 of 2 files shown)


lukefromdc:
I don't have enough understanding of all the sorting options (I only use a very few) to functionally evaluate this. Should I run a basic build and run test if merge if no issues, or wait for someone else to test this more thouroughly?

lukefromdc (approved):
Different sorting options (tested by trying the newly documented options) seem to work fine with this. No other issues came up in a test build and run

lukefromdc:
We have another approving review so will merge this

PR #1639, opened 2022-06-07
Fix file conflict dialog crashes

### LINKED ISSUE #1630: (Hidpi) Crash w segfault on confirmation in file conflict dialog

cwendling:
Fix #1630.

This contains 3 separate but fairly closely related changes (see each commit for details):

1) Fix reference to objects by releasing them in `dispose()` as we ought to instead of waiting for `finalize()`.  This is the real fix, and is the right thing to do in the GObject world: no object is supposed to hold a reference to any other object after `dispose()`.
2) Simplification of the icon update code, which optimizes it as well and allows for improving safety in 3 -- thanks @rbuj
3) thanks to 2, make use of `g_signal_connect_object()` to improve safety of the handler by making sure all parameters are valid at signal dispatch time.

1 alone is enough of a fix, but 2 is a nice improvement of the code, and 3 makes things safer.  Actually, 2+3 alone are also a fix for #1630, although not as thorough.

### DIFF
--- a/libcaja-private/caja-file-conflict-dialog.c
+++ b/libcaja-private/caja-file-conflict-dialog.c
@@ -672,6 +672,39 @@ caja_file_conflict_dialog_init (CajaFileConflictDialog *fcd)
 }
 
 static void
+do_dispose (GObject *self)
+{
+    CajaFileConflictDialogPrivate *details =
+        CAJA_FILE_CONFLICT_DIALOG (self)->details;
+
+    if (details->handle != NULL)
+    {
+        caja_file_list_cancel_call_when_ready (details->handle);
+        details->handle = NULL;
+    }
+
+    if (details->src_handler_id)
+    {
+        g_signal_handler_disconnect (details->source, details->src_handler_id);
+        caja_file_monitor_remove (details->source, self);
+        details->src_handler_id = 0;
+    }
+
+    if (details->dest_handler_id)
+    {
+        g_signal_handler_disconnect (details->destination, details->dest_handler_id);
+        caja_file_monitor_remove (details->destination, self);
+        details->dest_handler_id = 0;
+    }
+
+    g_clear_pointer (&details->source, caja_file_unref);
+    g_clear_pointer (&details->destination, caja_file_unref);
+    g_clear_pointer (&details->dest_dir, caja_file_unref);
+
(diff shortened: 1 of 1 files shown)


cwendling:
For anybody who'd merge this, please do not squash the commits together, it'd make more sense to have the split as is (and they all leave the code in an acceptable state, although c81cc650a7d3307feed134df719722444db1769c alone is not a fix.

rbuj (commented):
missing `g_signal_handler_disconnect` in the dispose method:
```C
    if (details->src_handler_id)
    {
        g_signal_handler_disconnect (details->source, details->src_handler_id);
        details->src_handler_id = 0;
    }

    if (details->dest_handler_id)
    {
        g_signal_handler_disconnect (details->destination, details->dest_handler_id);
        details->dest_handler_id = 0;
    }
```

rbuj (approved):
LGTM, I didn't see the last commit.

lukefromdc (approved):
Looks good on my end, some of these were the changes that worked before. Builds fine, runs with no new problems on a quick test-and stops the crashes. Will merge this and NOT squash as per request

lukefromdc:
Github just did something ugly: merged only the LAST commit on a "rebase and merge" will emergency revert and do a manual merge

lukefromdc:
Revert PR showed all three commits, looks like git worked and github just didn't show it all. Running git log or looking at the history on a local pull of master shows all three commits separately. Thus we are good to go here

cwendling:
@lukefromdc thanks, and yep master looks fine

lukefromdc:
Cherrypicked to 1.26 after testing on that branch

PR #1655, opened 2022-07-29
caja-bookmark: 'caja_bookmark_get_uri (CajaBookmark *)' may return NULL

rbuj:
When the uri is changed, caja automatically checks whether the uri exists in the sidebar to select it. It always checks in  places, bookmarks, ... . If the bookmarks list is empty, then some critical error messages are obtained such as:
```
GLib-GIO-CRITICAL **: 08:40:56.086: g_file_get_uri: assertion 'G_IS_FILE (file)' failed
```
When the bookmarks list is empty, the variable bookmark is NULL:
https://github.com/mate-desktop/caja/blob/141440bf89af9df9b64f77ac3a1c0fc63b768989/src/caja-bookmarks-sidebar.c#L332-L343

### DIFF
--- a/libcaja-private/caja-bookmark.c
+++ b/libcaja-private/caja-bookmark.c
@@ -274,11 +274,13 @@ char *
 caja_bookmark_get_uri (CajaBookmark *bookmark)
 {
     GFile *file;
-    char *uri;
+    char *uri = NULL;
 
-    file = caja_bookmark_get_location (bookmark);
-    uri = g_file_get_uri (file);
-    g_object_unref (file);
+    if ((file = caja_bookmark_get_location (bookmark)) != NULL)
+    {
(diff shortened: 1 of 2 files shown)


mbkma (approved):
builds fine, code looks good

lukefromdc:
OK, I see how this works: if` bookmark` is NULL, return NULL for` URI` as well, so as to avoid the 
` g_file_get_uri: assertion 'G_IS_FILE (file)' failed` 
testing this now w an emptied bookmarks list

lukefromdc:
Error messages confirmed

lukefromdc (approved):
Confirm this STOPS the error messages with empty bookmarks, will merge. Now to replace my bookmarks...

PR #1657, opened 2022-07-30
image properties: Handle rotated images

rbuj:
test on Fedora: file https://www.gstatic.com/webp/gallery/1.jpg
- rotate the image
```
$  sudo dnf install -y perl-Image-ExifTool
$  exiftool -Orientation=8 -n 1.jpg
```
- check the values height and width are swapped:
Screenshot at 2022-07-30 17-11-02


mbkma:
Does this only affect images rotated with exiftools? Because when rotating images with eom, then caja already shows the correct swapped values.

rbuj:
I think exiftool keeps the data chunk and only sets the exif metadata. eom modifies only the data chunk.

PR #1658, opened 2022-07-31
image properties: Improve the layout

rbuj:
**Before**
Screenshot at 2022-07-31 11-39-22

**After**
Screenshot at 2022-07-31 11-30-14

### DIFF
--- a/src/caja-image-properties-page.c
+++ b/src/caja-image-properties-page.c
@@ -53,12 +53,14 @@ struct _CajaImagePropertiesPagePrivate
     GCancellable *cancellable;
     GtkWidget *vbox;
     GtkWidget *loading_label;
+    GtkWidget *grid;
     GdkPixbufLoader *loader;
     gboolean got_size;
     gboolean pixbuf_still_loading;
     char buffer[LOAD_BUFFER_SIZE];
     int width;
     int height;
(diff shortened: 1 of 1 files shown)


mbkma (approved):
builds and runs fine, layout looks way better now, code looks good on first sight

lukefromdc:
Layouts of the image properties tab in both images looks identical to me.Is this an effort to improve the layout of the UI or of the code.  Note that we add an additional widget (the grid) here as well.

rbuj:
Yes, you are right. It adds an additional grid to align the columns instead of stacking a single label for each property. atril also has a grid. I think you mean whether to use bold or italics for the property names... I think the two property pages layouts look pretty similar.

Captura de pantalla 2022-08-02 a les 20 24 31

There is another PR that adds the properties page for audio and video files,, but I haven't modified the original layout.

Captura de pantalla 2022-08-02 a les 20 46 45

Image properties can be grouped (General, Exif and Exempi) as in the property pages above.

mbkma:
@lukefromdc before this PR the right column is not aligned properly.

lukefromdc (approved):
Works fine, and I didn't find any problems with the new A/V  properties tab either.

PR #1661, opened 2022-08-01
image properties: Look for supported mime types at runtime

rbuj:
See https://gitlab.gnome.org/GNOME/nautilus/-/commit/7aef6be


lukefromdc (approved):
This works, and it does show the properties tab for webp images as well as older filetypes. I don't have a lot of files in newer formats but I do have a  few of those kept around for test purposes

lukefromdc:
Merged as this worked and was being ignored due to volume

PR #1676, opened 2022-11-09
icon view: Refresh icon positions for manual layout on zoom change

cwendling:
When zoom changes on a manual layout icon view, the available area changes and can lead to some icons to either overflow or be able to go back to their actually saved position.

This is done correctly when the view is reloaded entirely, but not in response to zoom change, leading to disappearing icons (when zoom increases) or unexpected empty space (when zoom decreases).

Fix this by re-computing actual positions based on saved positions when zoom changes, to match what would actually happen when the view gets loaded.

---

To test this:
1. Make sure you show desktop icons `gsettings set org.mate.background show-desktop-icons true`
2. Put some icons to the far bottom and/or right edges of the desktop
3. Play with the icon view default zoom `gsettings set org.mate.caja.icon-view default-zoom-level large` (`smaller`, `small`, `standard`, `large`, `laregr`, etc.).  You can also use *Edit → Preferences → Views → Defaults → Icon View → Default zoom level*.

Expected results (with this PR): when the zoom changes, all icons remain visible (e.g. don't overflow)

Actual results (without this PR): when zoom grows, some icons overflow and disappear.  If you refresh the desktop (<kbd>F5</kbd> on it) or restart Caja, they appear properly.

Known issues: given the bigger the icons the least amount can fit on the desktop, at some point when you increase the zoom factor, some icons might overlap.  That didn't change, and I don't know of a real theoretical solution as there always gonna be so much space to use anyway.

Basically, this PR should make the state of the desktop after a zoom change match the one after a zoom change followed by a reload/restart.

### DIFF
--- a/libcaja-private/caja-icon-container.c
+++ b/libcaja-private/caja-icon-container.c
@@ -296,14 +296,15 @@ icon_is_positioned (const CajaIcon *icon)
 
 /* x, y are the top-left coordinates of the icon. */
 static void
-icon_set_position (CajaIcon *icon,
-                   double x, double y)
+icon_set_position_full (CajaIcon *icon,
+                        double x, double y,
+                        gboolean force)
 {
     CajaIconContainer *container;
     int x1, x2, y1, y2;
     EelDRect icon_bounds;
 
-    if (icon->x == x && icon->y == y)
+    if (!force && icon->x == x && icon->y == y)
     {
         return;
     }
@@ -389,6 +390,13 @@ icon_set_position (CajaIcon *icon,
 
     icon->x = x;
     icon->y = y;
+}
+
+static void
+icon_set_position (CajaIcon *icon,
+                   double x, double y)
+{
+    icon_set_position_full (icon, x, y, FALSE);
 }
 
 static void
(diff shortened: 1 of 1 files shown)


lukefromdc (commented):
This will prevent icons from overflowing when zooming on on the desktop but not on a navigation window. They return to original positions without issue when zooming out. When zooming further out, the entire layout is zoomed out, showing extra space to the bottom and right same as before and same as in a navigation window. Not sure if that is the intended behavior but it is not objectionable. 

Navigation windows w manual icon arrangement seemed unaffected when zooming in: the whole arrangment zoomed in on,icons overflowing offscreen to bottom and right

cwendling:
> When zooming further out, the entire layout is zoomed out, showing extra space to the bottom and right same as before and same as in a navigation window. Not sure if that is the intended behavior but it is not objectionable.

I think it is intended: the view keeps the stored locations and clamps them to the view area, but they are still relative to the top left corner, not "at bottom right", or "20% from the top".  That could possibly be a nice feature (although I'm not sure how useful it would be, or how well it would actually play with the icons growing), but that's orthogonal to the issue here I think.

> Navigation windows w manual icon arrangement seemed unaffected when zooming in: the whole arrangment zoomed in on,icons overflowing offscreen to bottom and right

I think that's expected, as they gain a scrollbar then (which the desktop doesn't, and wouldn't expect).  And you can see that if you refresh the view, they still overflow, so at least it would be a *different* problem.

PR #1678, opened 2022-11-23
icon view: Fix positioning overflowing icons on the desktop

cwendling:
Properly update the icon data before placing the icon, because positioning might depend on full icon contents on the desktop, whereas updating contents don't care about position.

When an icon position overflows the desktop area, it is clamped to stay in the visible area, but this computation depends on accurate icon and label sizes, which is only available when the icon is fully loaded.

Fix the code to first load the contents and then position instead of the other way around, which was actually trivial.

Note that visible positions were most often correct anyway for two reasons:

1. Most of the time icons do not overflow, as they are positioned on the final desktop size anyway.  It however can easily happen reducing monitor resolution or increasing desktop view zoom. (see also #1676)
2. A second layout pass happens most of the time (I'm not yet sure why and when though), but not when an update is triggered before the previous one terminated (e.g. quickly hitting F5 twice).

### DIFF
--- a/libcaja-private/caja-icon-container.c
+++ b/libcaja-private/caja-icon-container.c
@@ -7687,19 +7687,6 @@ assign_icon_position (CajaIconContainer *container,
 }
 
 static void
-finish_adding_icon (CajaIconContainer *container,
-                    CajaIcon *icon)
-{
-    caja_icon_container_update_icon (container, icon);
-    eel_canvas_item_show (EEL_CANVAS_ITEM (icon->item));
-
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
Didn't get any visible changes in behavior on changing desktop zoom with this installed (didn't get the issue just before it) but also had no problems. No changes in behavior when dropping an icon onto the desktop, adding a new icon,or removing an icon either.

Not sure why this was ever a separate function unless it was once used elsewhere in the file

lukefromdc:
Again calling for another review if we can get one

PR #1680, opened 2022-11-23
Remove dead code found by cppcheck

cwendling:
This removes various unused code, based on a [cppcheck report](https://caja.mate-desktop.dev/2022-11-23-174623-5790-cppcheck@ae663c369cf2_desktop-no-overflow/).  I carefully checked everything the best I could (not simply trusting the reports), but most of it was accurate.  Most stuff predates MATE repository so I can only guess they are leftovers from some old refactoring.

### DIFF
--- a/libcaja-private/caja-icon-dnd.c
+++ b/libcaja-private/caja-icon-dnd.c
@@ -620,8 +620,6 @@ caja_icon_container_selection_items_local (CajaIconContainer *container,
     /* must have at least one item */
     g_assert (items);
 
-    result = FALSE;
-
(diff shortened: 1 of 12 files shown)


lukefromdc (approved):
Works fine here, and changing view options (involving changes to the code discussed earlier in this pr) works fine. Cppcheck has its limits but seems to do quite well at finding these sorts of issues. Never had a problem resulting from its use.

lukefromdc:
Calling for another review if we can get it

lukefromdc:
Looks like we're not going to get another review, proceeding to merge

PR #1684, opened 2022-11-27
directory: Include hidden files in properties window totals

### LINKED ISSUE #527: Caja doesn't respect hidden files in properties window
My Caja doesn't show hidden files as default (otherwise it works). When I have hidden file or directory then it is not respected in property window, so the calculated size and object count is incorrect.

Tested with 1.12.6 and master: ce7cc9580809d4017c74b0128f7e82f94eb173d9

Steps to reproduce:
1. Don't show hidden files/directories as default:

2. Create two files inside a directory.
3. Hide one of them (add dot "." before the name).
4. Press F5 (it hides the file) and optionally press Ctrl+H to show hidden files.
5. Open property for the directory:

Expected results: Caja shows real directory object count and real directory size even if there are hidden files/directories.
Otherwise we can see e.g. that the directory have 1MiB and it real contents will have e.g. 100GiB. It is misleading...
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/32228619-caja-doesn-t-respect-hidden-files-in-properties-window?utm_campaign=plugin&utm_content=tracker%2F651521&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F651521&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #738: Caja displays wrong properties info
Caja displays wrong properties info as you can see from the link below.

https://drive.google.com/open?id=0B3iAYb-9aeiHXzhwTzFIRzAwRVk

The folder is about 180GB but Caja only says that is 7.5GB. 
In addition, Caja wrongly reports the numbers of items and the free space.

Please let me know if you need any more info.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/41973514-caja-displays-wrong-properties-info?utm_campaign=plugin&utm_content=tracker%2F651521&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F651521&utm_medium=issues&utm_source=github).
</bountysource-plugin>

wesinator:
This change will make the directory properties info always count "hidden" dotfiles in the count and size.
based on https://gitlab.gnome.org/GNOME/nautilus/-/merge_requests/821/diffs

fixes #527, fixes #738

### DIFF
--- a/libcaja-private/caja-directory-async.c
+++ b/libcaja-private/caja-directory-async.c
@@ -2823,11 +2823,6 @@ deep_count_one (DeepCountState *state,
     CajaFile *file;
     gboolean is_seen_inode;
 
-    if (should_skip_file (NULL, info))
-    {
-        return;
-    }
-
     is_seen_inode = seen_inode (state, info);
     if (!is_seen_inode)
     {


cwendling:
Not tested, but I just got extremely surprised and worried that folder size doesn't take hidden files into account, so I'm all for a fix.

This said, this topic is apparently somewhat controversial on the Nautilus side, see https://gitlab.gnome.org/GNOME/nautilus/-/issues/638

wesinator:
> This said, this topic is apparently somewhat controversial on the Nautilus side, see https://gitlab.gnome.org/GNOME/nautilus/-/issues/638

Colomban, thanks for taking a look.
Yes, I saw that. 
I opted for a quick fix to get some traction on this,and I think this proposed change is better than the current state. We can fix the 'bug' and then work out optimal implementation. 
https://github.com/mate-desktop/caja/issues/738#issuecomment-285326038 has some good suggestions

lukefromdc (approved):
This works as expected, and the code change is simple here

PR #1688, opened 2022-12-11
Enable showing in Budgie Desktop and XFCE

JoshStrobl:
This PR enables the displaying of Caja (as a file manager) for use in Budgie Desktop and XFCE. Prior to this commit, it would only be able accessible in MATE or with downstreams like Fedora patching in the support. The addition of these is important as file managers such as Nautilus adopt libadwaita and begin to directly come into conflict with user theming choices.

### DIFF
--- a/data/caja-browser.desktop.in.in
+++ b/data/caja-browser.desktop.in.in
@@ -12,4 +12,4 @@ Type=Application
 Categories=GTK;System;Core;
 # Translators: Search terms to find this application. Do NOT translate or localize the semicolons! The list MUST also end with a semicolon!
 Keywords=files;browser;manager;MATE;
-OnlyShowIn=MATE;
+OnlyShowIn=Budgie;MATE;XFCE;


lukefromdc:
I do not have either of those DE's installed so cannot directly test function of this, but I see 
no reason it would create any problems for MATE. The devs of XFCE and Budgie probably
should weigh in on this as well.

BTW, there is a way to defeat Libadwaita's "don't theme my app" behavior: setting the desired
theme as an environmental variable prior to launching the app. That's how I force-themed
metadata-cleaner to match the rest of my DE. Libadwaita however is likely to greatly reduce
use of GNOME apps in other desktop environments.

Also a distro-or a skilled end user-could edit any desired theme right into libadwaita and recompile.
Distros could easily do this, so GNOME's original objective of stopping distros from changing
themes won't work. With any open source program, any hardcoded variable can be changed by
a distro patch, no known exceptions.

JoshStrobl:
> The devs of XFCE and Budgie probably should weigh in on this as well.

I am the lead developer of Budgie.

lukefromdc:
Than we are just waiting to hear back from XFCE.

lukefromdc (approved):
Didn't get any issues running this in a MATE session (note same session not  a restarted session) but caja appears in my panel menu as before and works normally as I absolutely expected but test installed to be sure

lukefromdc:
New session caja works same as before too. All good to go on the MATE side

lukefromdc:
Any updates on this? We need to know what XFCE devs think of this, especially given XFCE defaults to Thunar as file manager. Keep in mind that caja can be run from terminal or by any other means of invoking a command without regard for the .desktop file, this should only affect showing up in menus as we handle autostarting caja by code in the session manager itself, which is in turn user-selectable via dconf-editor.

Not sure how XFCE handles autostarting Thunar, we do NOT want XFCE users with caja installed to end up with caja and thunar both starting at once. At least they should not fight over the desktop, as Caja will not manage the desktop outside a mate session unless invoked with `--force-desktop`

JoshStrobl:
> Any updates on this? We need to know what XFCE devs think of this

No, there's no update on this. The XFCE change was because this already existed in Fedora's caja package, see commit below. All I did was add Budgie in addition to that since I need it for the Fedora Budgie Spin, and said "hey, I may as well try getting this upstreamed while I'm at it". I personally have no interest in XFCE, I just didn't want to remove what is likely an intentional change on the Fedora side.

- https://src.fedoraproject.org/rpms/caja/pull-request/4#request_diff

PR #1722, opened 2023-06-17
wayland: initial support for showing the desktop

lukefromdc:
*Uses gtk-layer-shell to keep the desktop always on the bottom *Uses gtk-layer-shell to keep the desktop on all workpaces *Depends on two changes in mate-desktop package
*We can later make wayland support optional here due to new dependency on gtk-layer-shell

*Depends on https://github.com/mate-desktop/mate-desktop/pull/558

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2,7 +2,7 @@ AC_PREREQ(2.60)
 
 dnl ===========================================================================
 
-m4_define(gdkpixbuf_minver,           2.36.5)
+m4_define(gdkpixbuf_minver,            2.36.5)
 m4_define(glib_minver,                 2.58.1)
 m4_define(gio_minver,                  2.50.0)
 m4_define(mate_desktop_minver,         1.17.3)
@@ -13,6 +13,7 @@ m4_define(exif_minver,                 0.6.14)
 m4_define(exempi_minver,               1.99.5)
 m4_define(gail_minver,                 3.0.0)
 m4_define(notify_minver,               0.7.0)
+m4_define(gtk-layer-shell_minver,      0.8.0)
 
 
 dnl 1. If the library code has changed at all since last release, then increment revision.
@@ -177,7 +178,33 @@ if test "x$enable_xmp" != "xno"; then
     AC_SUBST(EXEMPI_LIBS)
 fi
 
-dnl ==========================================================================
+dnl **************************************************************************
+dnl Whether or not to build with wayland support and check for gtk-layer-shell
+dnl **************************************************************************
+
+
+AC_ARG_ENABLE(wayland,
+              [AS_HELP_STRING([--enable-wayland],
+                              [Explicitly enable or disable Wayland support
+                              (default is to enable only if Wayland client development library is detected)])],
+              [enable_wayland=$enableval],
+              [enable_wayland=auto])
+
+# Check if we have gtk-layer-shell installed, and thus should build with Wayland support
+have_wayland=no
+if test "x$enable_wayland" != "xno"; then
+  PKG_CHECK_MODULES(WAYLAND, gtk-layer-shell-0 >= gtk-layer-shell_minver wayland-client, have_wayland=yes, [
+        if test "x$enable_wayland" = "xyes"; then
+          AC_MSG_ERROR([Wayland enabled but GTK Layer Shell library not found])
+        fi
+    ])
+fi
+
+AM_CONDITIONAL(ENABLE_WAYLAND, [test "x$have_wayland" = "xyes"])
+
+if test "x$have_wayland" = "xyes"; then
+  AC_DEFINE(HAVE_WAYLAND, 1, [Have the Wayland development library])
+fi
 
 dnl ****************************
 dnl *** Check for libselinux ***
@@ -240,7 +267,7 @@ AC_SUBST(LIBCAJA_EXTENSION_LIBS)
 
 dnl core caja
 PKG_CHECK_MODULES(GMODULE, gmodule-2.0, [GMODULE_ADD="gmodule-2.0"],[GMODULE_ADD=""])
-CORE_MODULES="glib-2.0 $GMODULE_ADD mate-desktop-2.0 gthread-2.0 gio-2.0 gio-unix-2.0 gail-3.0 libxml-2.0 $EXTRA_CORE_MODULES gtk+-3.0 x11"
+CORE_MODULES="glib-2.0 $GMODULE_ADD mate-desktop-2.0 gthread-2.0 gio-2.0 gio-unix-2.0 gail-3.0 libxml-2.0 gtk+-3.0 x11"
 CORE_CFLAGS="`$PKG_CONFIG --cflags $CORE_MODULES`"
 AC_SUBST(CORE_CFLAGS)
 CORE_LIBS="`$PKG_CONFIG --libs $CORE_MODULES`"
@@ -323,6 +350,7 @@ Configure summary:
     PackageKit support:           $msg_packagekit
     Native Language support:      $USE_NLS
     Self check:                   $msg_self_check
+    Wayland support:              ${have_wayland}
 
     caja-extension documentation: ${enable_gtk_doc}
     caja-extension introspection: ${found_introspection}
--- a/configure.ac
+++ b/configure.ac
@@ -13,7 +13,7 @@ m4_define(exif_minver,                 0.6.14)
 m4_define(exempi_minver,               1.99.5)
 m4_define(gail_minver,                 3.0.0)
 m4_define(notify_minver,               0.7.0)
-m4_define(gtk-layer-shell_minver,      0.8.0)
+m4_define(gtk_layer_shell_minver,      0.8.0)
 
 
 dnl 1. If the library code has changed at all since last release, then increment revision.
@@ -192,13 +192,11 @@ AC_ARG_ENABLE(wayland,
 
 # Check if we have gtk-layer-shell installed, and thus should build with Wayland support
 have_wayland=no
-if test "x$enable_wayland" != "xno"; then
-  PKG_CHECK_MODULES(WAYLAND, gtk-layer-shell-0 >= gtk-layer-shell_minver wayland-client, have_wayland=yes, [
-        if test "x$enable_wayland" = "xyes"; then
-          AC_MSG_ERROR([Wayland enabled but GTK Layer Shell library not found])
-        fi
-    ])
-fi
+m4_define([WAYLAND_DEPS], [gtk-layer-shell-0 >= gtk_layer_shell_minver wayland-client])
+AS_IF([test "x$enable_wayland" = "xyes"],
+      [PKG_CHECK_MODULES([WAYLAND], [WAYLAND_DEPS], [have_wayland=yes])],
+      [test "x$enable_wayland" != "xno"],
+      [PKG_CHECK_MODULES([WAYLAND], [WAYLAND_DEPS], [have_wayland=yes], [have_wayland=no])])
 
 AM_CONDITIONAL(ENABLE_WAYLAND, [test "x$have_wayland" = "xyes"])
 
--- a/eel/eel-background.c
+++ b/eel/eel-background.c
@@ -105,7 +105,7 @@ free_background_surface (EelBackground *self)
            it will live forever, so we need to kill it manually.
            If set as root background it will be killed next time the
            background is changed. */
-        if (self->details->unset_root_surface)
+        if (GDK_IS_X11_DISPLAY (gdk_display_get_default()) && (self->details->unset_root_surface))
         {
             XKillClient (cairo_xlib_surface_get_display (surface),
                          cairo_xlib_surface_get_drawable (surface));
@@ -312,7 +312,7 @@ drawable_get_adjusted_size (EelBackground *self,
     {
         GdkScreen *screen = gtk_widget_get_screen (self->details->widget);
         GdkDisplay *display = gdk_screen_get_display (screen);
-        if  (GDK_IS_X11_DISPLAY (display))
+        if (GDK_IS_X11_DISPLAY (display))
         {
                gint scale = gtk_widget_get_scale_factor (self->details->widget);
                *width = WidthOfScreen (gdk_x11_screen_get_xscreen (screen)) / scale;
@@ -433,8 +433,8 @@ set_root_surface (EelBackground *self,
         self->details->bg_surface = mate_bg_create_surface (self->details->bg, window,
         						    width, height, TRUE);
     }
-
-    if (self->details->bg_surface != NULL)
+    GdkDisplay *display = gdk_screen_get_display (screen);
+    if ((GDK_IS_X11_DISPLAY (display)) && (self->details->bg_surface != NULL))
         mate_bg_set_surface_as_root (screen, self->details->bg_surface);
 }
 
@@ -649,7 +649,7 @@ widget_realized_setup (GtkWidget     *widget,
     }
 
     GdkScreen *screen = gtk_widget_get_screen (widget);
-    GdkWindow *window = gdk_screen_get_root_window (screen);
+    GdkDisplay *display = gdk_screen_get_display (screen);
 
     if (self->details->screen_size_handler > 0)
     {
@@ -665,8 +665,16 @@ widget_realized_setup (GtkWidget     *widget,
     self->details->screen_monitors_handler =
         g_signal_connect (screen, "monitors-changed", G_CALLBACK (screen_size_changed), self);
 
-    self->details->use_common_surface =
-        (gdk_window_get_visual (window) == gtk_widget_get_visual (widget)) ? TRUE : FALSE;
+    if (GDK_IS_X11_DISPLAY (display))
+    {
+        GdkWindow *window = gdk_screen_get_root_window (screen);
+        self->details->use_common_surface =
+            (gdk_window_get_visual (window) == gtk_widget_get_visual (widget)) ? TRUE : FALSE;
+    }
+    else    /*Wayland is always composited*/
+    {
+        self->details->use_common_surface = TRUE;
+    }
 
     init_fade (self);
 }
--- a/eel/eel-editable-label.c
+++ b/eel/eel-editable-label.c
@@ -1028,7 +1028,6 @@ eel_editable_label_ensure_layout (EelEditableLabel *label,
             else
             {
                 gint wrap_width;
-                gint scale;
 
                 pango_layout_set_width (label->layout, -1);
                 pango_layout_get_extents (label->layout, NULL, &logical_rect);
@@ -1039,7 +1038,6 @@ eel_editable_label_ensure_layout (EelEditableLabel *label,
                 longest_paragraph = width;
 
                 wrap_width = get_label_wrap_width (label);
-                scale = gtk_widget_get_scale_factor (widget);
                 width = MIN (width, wrap_width);
                 width = MIN (width,
                 PANGO_SCALE * (gdk_screen_width () + 1) / 2);
--- a/src/caja-application.c
+++ b/src/caja-application.c
@@ -634,19 +634,15 @@ selection_get_cb (GtkWidget          *widget,
 static GtkWidget *
 get_desktop_manager_selection (GdkDisplay *display)
 {
-	/*FIXME: for Wayland we need a new desktop canvas anyway
-    *so when that is ready we will need to add a wayland code
-    *path here to support it
-    */
-    if  (GDK_IS_X11_DISPLAY (display)){
-        char selection_name[32];
-        GdkAtom selection_atom;
-        Window selection_owner;
-        GtkWidget *selection_widget;
+    char selection_name[32];
+    GdkAtom selection_atom;
+    Window selection_owner;
+    GtkWidget *selection_widget;
 
-        g_snprintf (selection_name, sizeof (selection_name), "_NET_DESKTOP_MANAGER_S0");
-        selection_atom = gdk_atom_intern (selection_name, FALSE);
-
+    g_snprintf (selection_name, sizeof (selection_name), "_NET_DESKTOP_MANAGER_S0");
+    selection_atom = gdk_atom_intern (selection_name, FALSE);
+    if (GDK_IS_X11_DISPLAY (display))
+    {
         selection_owner = XGetSelectionOwner (GDK_DISPLAY_XDISPLAY (display),
                                           gdk_x11_atom_to_xatom_for_display (display,
                                                   selection_atom));
@@ -654,24 +650,31 @@ get_desktop_manager_selection (GdkDisplay *display)
         {
         return NULL;
         }
+    }
+    selection_widget = gtk_invisible_new_for_screen (gdk_display_get_default_screen (display));
+    /* We need this for gdk_x11_get_server_time() */
+    gtk_widget_add_events (selection_widget, GDK_PROPERTY_CHANGE_MASK);
 
-        selection_widget = gtk_invisible_new_for_screen (gdk_display_get_default_screen (display));
-        /* We need this for gdk_x11_get_server_time() */
-        gtk_widget_add_events (selection_widget, GDK_PROPERTY_CHANGE_MASK);
-
+    if (GDK_IS_X11_DISPLAY (display))
+    {
         if (gtk_selection_owner_set_for_display (display,
             selection_widget,
             selection_atom,
             gdk_x11_get_server_time (gtk_widget_get_window (selection_widget))))
         {
-
             g_signal_connect (selection_widget, "selection_get",
-                          G_CALLBACK (selection_get_cb), NULL);
+                              G_CALLBACK (selection_get_cb), NULL);
             return selection_widget;
         }
-
-        gtk_widget_destroy (selection_widget);
     }
+    else
+    {
+        g_signal_connect (selection_widget, "selection_get",
+                          G_CALLBACK (selection_get_cb), NULL);
+        return selection_widget;
+    }
+
+    gtk_widget_destroy (selection_widget);
 
     return NULL;
 }
--- a/src/caja-desktop-window.c
+++ b/src/caja-desktop-window.c
@@ -171,8 +171,8 @@ caja_desktop_window_screen_size_changed (GdkScreen             *screen,
         /*No root window or primary monitor in wayland unless compositors add it back*/
         GdkRectangle geometry = {0};
         GdkMonitor *monitor;
-        monitor = gdk_display_get_monitor(display, 0);
-        gdk_monitor_get_geometry(monitor,&geometry);
+        monitor = gdk_display_get_monitor (display, 0);
+        gdk_monitor_get_geometry (monitor, &geometry);
         width_request = geometry.width;
         height_request = geometry.height;
     }
@@ -207,8 +207,8 @@ caja_desktop_window_new (CajaApplication *application,
         */
         GdkRectangle geometry = {0};
         GdkMonitor *monitor;
-        monitor = gdk_display_get_monitor(display, 0);
-        gdk_monitor_get_geometry(monitor,&geometry);
+        monitor = gdk_display_get_monitor (display, 0);
+        gdk_monitor_get_geometry (monitor, &geometry);
         width_request = geometry.width;
         height_request = geometry.height;
     }
@@ -380,10 +380,10 @@ realize (GtkWidget *widget)
     /* This is the new way to set up the desktop window in x11 but not for wayland */
     display = gtk_widget_get_display (widget);
     if (GDK_IS_X11_DISPLAY (display))
-        {
-            set_wmspec_desktop_hint (gtk_widget_get_window (widget));
-            set_desktop_window_id (window, gtk_widget_get_window (widget));
-        }
+    {
+        set_wmspec_desktop_hint (gtk_widget_get_window (widget));
+        set_desktop_window_id (window, gtk_widget_get_window (widget));
+    }
 
     details->size_changed_id =
         g_signal_connect (gtk_window_get_screen (GTK_WINDOW (window)), "size_changed",
--- a/src/caja-desktop-window.c
+++ b/src/caja-desktop-window.c
@@ -29,7 +29,10 @@
 #include <gtk/gtk.h>
 #include <gio/gio.h>
 #include <glib/gi18n.h>
-
+#ifdef HAVE_WAYLAND
+#include <gdk/gdkwayland.h>
+#include <gtk-layer-shell/gtk-layer-shell.h>
+#endif
 #include <eel/eel-background.h>
 #include <eel/eel-vfs-extensions.h>
 
@@ -156,9 +159,23 @@ caja_desktop_window_screen_size_changed (GdkScreen             *screen,
 {
     int width_request, height_request;
 
-    GdkWindow *root_window;
-    root_window = gdk_screen_get_root_window (screen);
-    gdk_window_get_geometry (root_window, NULL, NULL, &width_request, &height_request);
+    GdkDisplay *display = gdk_screen_get_display (screen);
+    if (GDK_IS_X11_DISPLAY (display))
+    {
+        GdkWindow *root_window;
+        root_window = gdk_screen_get_root_window (screen);
+        gdk_window_get_geometry (root_window, NULL, NULL, &width_request, &height_request);
+    }
+    else
+    {
+        /*No root window or primary monitor in wayland unless compositors add it back*/
+        GdkRectangle geometry = {0};
+        GdkMonitor *monitor;
+        monitor = gdk_display_get_monitor(display, 0);
+        gdk_monitor_get_geometry(monitor,&geometry);
+        width_request = geometry.width;
+        height_request = geometry.height;
+    }
 
     g_object_set (window,
                   "width_request", width_request,
@@ -174,10 +191,27 @@ caja_desktop_window_new (CajaApplication *application,
     int width_request, height_request;
     int scale;
 
-    scale = gdk_window_get_scale_factor (gdk_screen_get_root_window (screen));
-
-    width_request = WidthOfScreen (gdk_x11_screen_get_xscreen (screen)) / scale;
-    height_request = HeightOfScreen (gdk_x11_screen_get_xscreen (screen)) / scale;
+    GdkDisplay *display = gdk_screen_get_display (screen);
+    if (GDK_IS_X11_DISPLAY (display))
+    {
+        scale = gdk_window_get_scale_factor (gdk_screen_get_root_window (screen));
+        width_request = WidthOfScreen (gdk_x11_screen_get_xscreen (screen)) / scale;
+        height_request = HeightOfScreen (gdk_x11_screen_get_xscreen (screen)) / scale;
+    }
+    else
+    {
+        /*FIXME: There is no primary monitor in wayland itself
+        *compositors can implement this but as this is written
+        *only a few wayland compositors allow setting a primary monitor
+        *and they all do it differently. For now, use the first monitor
+        */
+        GdkRectangle geometry = {0};
+        GdkMonitor *monitor;
+        monitor = gdk_display_get_monitor(display, 0);
+        gdk_monitor_get_geometry(monitor,&geometry);
+        width_request = geometry.width;
+        height_request = geometry.height;
+    }
 
     window = CAJA_DESKTOP_WINDOW
              (gtk_widget_new (caja_desktop_window_get_type(),
@@ -191,11 +225,47 @@ caja_desktop_window_new (CajaApplication *application,
     /* will cause the desktop window to open at the wrong size in gtk 3.20 */
     gtk_window_set_default_size (GTK_WINDOW (window), -1, -1);
 
+    /*For wayland only
+     *Code taken from gtk-layer-shell simple-example.c
+     */
+#ifdef HAVE_WAYLAND
+    if (GDK_IS_WAYLAND_DISPLAY (display))
+    {
+        GtkWindow *gtkwin;
+        gtkwin = (GTK_WINDOW(window));
+
+        /* Before the window is first realized, set it up to be a layer surface */
+        gtk_layer_init_for_window (gtkwin);
+
+        /* Order below normal windows */
+        gtk_layer_set_layer (gtkwin, GTK_LAYER_SHELL_LAYER_BOTTOM);
+
+        gtk_layer_set_namespace (gtkwin, "desktop");
+
+        /*Anchor the desktop to all four corners
+         *This is much simpler than on x11 and
+         *should always render the desktop across
+         *all of the screen
+         */
+        gtk_layer_set_anchor (gtkwin, GTK_LAYER_SHELL_EDGE_TOP, TRUE);
+        gtk_layer_set_anchor (gtkwin, GTK_LAYER_SHELL_EDGE_BOTTOM, TRUE);
+        gtk_layer_set_anchor (gtkwin, GTK_LAYER_SHELL_EDGE_LEFT, TRUE);
+        gtk_layer_set_anchor (gtkwin, GTK_LAYER_SHELL_EDGE_RIGHT, TRUE);
+
+        /*Enable keyboard use on the desktop*/
+        gtk_layer_set_keyboard_mode (gtkwin, GTK_LAYER_SHELL_KEYBOARD_MODE_ON_DEMAND);
+    }
+#endif
     /* Special sawmill setting*/
     GdkWindow *gdkwin;
-    gtk_widget_realize (GTK_WIDGET (window));
+    if ((GDK_IS_X11_DISPLAY (display)))
+        gtk_widget_realize (GTK_WIDGET (window));
+    else
+        gtk_widget_show (GTK_WIDGET (window));
+
     gdkwin = gtk_widget_get_window (GTK_WIDGET (window));
-    if (gdk_window_ensure_native (gdkwin)) {
+    if ((GDK_IS_X11_DISPLAY (display)) && (gdk_window_ensure_native (gdkwin)))
+    {
         Display *disp = GDK_DISPLAY_XDISPLAY (gdk_window_get_display (gdkwin));
         XClassHint *xch = XAllocClassHint ();
         xch->res_name = "desktop_window";
@@ -230,15 +300,21 @@ unrealize (GtkWidget *widget)
     CajaDesktopWindow *window;
     CajaDesktopWindowPrivate *details;
     GdkWindow *root_window;
+    GdkDisplay *display;
 
     window = CAJA_DESKTOP_WINDOW (widget);
     details = window->details;
+    display = gtk_widget_get_display (widget);
 
-    root_window = gdk_screen_get_root_window (
+    /*Avoid root window on wayland-it's not supposed to work*/
+    if (GDK_IS_X11_DISPLAY (display))
+    {
+        root_window = gdk_screen_get_root_window (
                       gtk_window_get_screen (GTK_WINDOW (window)));
 
-    gdk_property_delete (root_window,
-                         gdk_atom_intern ("CAJA_DESKTOP_WINDOW_ID", TRUE));
+        gdk_property_delete (root_window,
+                             gdk_atom_intern ("CAJA_DESKTOP_WINDOW_ID", TRUE));
+    }
 
     if (details->size_changed_id != 0) {
         g_signal_handler_disconnect (gtk_window_get_screen (GTK_WINDOW (window)),
@@ -249,6 +325,7 @@ unrealize (GtkWidget *widget)
     GTK_WIDGET_CLASS (caja_desktop_window_parent_class)->unrealize (widget);
 }
 
+/*This should only be reached in x11*/
 static void
 set_wmspec_desktop_hint (GdkWindow *window)
 {
@@ -262,18 +339,18 @@ set_wmspec_desktop_hint (GdkWindow *window)
                          GDK_PROP_MODE_REPLACE, (guchar *) &atom, 1);
 }
 
+/*This should only be reached in x11*/
 static void
 set_desktop_window_id (CajaDesktopWindow *window,
                        GdkWindow             *gdkwindow)
 {
-    /* Tuck the desktop windows xid in the root to indicate we own the desktop.
+    /* Tuck the desktop windows xid in the root to indicate we own the desktop in on x11
      */
     Window window_xid;
     GdkWindow *root_window;
 
     root_window = gdk_screen_get_root_window (
                       gtk_window_get_screen (GTK_WINDOW (window)));
-
     window_xid = GDK_WINDOW_XID (gdkwindow);
 
     gdk_property_change (root_window,
@@ -289,17 +366,24 @@ realize (GtkWidget *widget)
     CajaDesktopWindowPrivate *details;
     window = CAJA_DESKTOP_WINDOW (widget);
     details = window->details;
+    GdkDisplay *display;
 
     /* Make sure we get keyboard events */
-    gtk_widget_set_events (widget, gtk_widget_get_events (widget)
-                           | GDK_KEY_PRESS_MASK | GDK_KEY_RELEASE_MASK);
-    /* Do the work of realizing. */
+    display = gtk_widget_get_display (widget);
+    if (GDK_IS_X11_DISPLAY (display))
+        gtk_widget_set_events (widget, gtk_widget_get_events (widget)
+                               | GDK_KEY_PRESS_MASK | GDK_KEY_RELEASE_MASK);
+
+    /*Do the work of realizing. */
     GTK_WIDGET_CLASS (caja_desktop_window_parent_class)->realize (widget);
 
-    /* This is the new way to set up the desktop window */
-    set_wmspec_desktop_hint (gtk_widget_get_window (widget));
-
-    set_desktop_window_id (window, gtk_widget_get_window (widget));
+    /* This is the new way to set up the desktop window in x11 but not for wayland */
+    display = gtk_widget_get_display (widget);
+    if (GDK_IS_X11_DISPLAY (display))
+        {
+            set_wmspec_desktop_hint (gtk_widget_get_window (widget));
+            set_desktop_window_id (window, gtk_widget_get_window (widget));
+        }
 
     details->size_changed_id =
         g_signal_connect (gtk_window_get_screen (GTK_WINDOW (window)), "size_changed",
@@ -311,7 +395,6 @@ draw (GtkWidget *widget,
       cairo_t   *cr)
 {
     eel_background_draw (widget, cr);
-
     return GTK_WIDGET_CLASS (caja_desktop_window_parent_class)->draw (widget, cr);
 }
 
@@ -332,7 +415,8 @@ caja_desktop_window_class_init (CajaDesktopWindowClass *klass)
     wclass->realize = realize;
     wclass->unrealize = unrealize;
     wclass->map = map;
-    wclass->draw = draw;
+    if (GDK_IS_X11_DISPLAY (gdk_display_get_default()))
+        wclass->draw = draw;
 
     gtk_widget_class_set_accessible_type (wclass, CAJA_TYPE_DESKTOP_WINDOW_ACCESSIBLE);
 
--- a/src/file-manager/fm-desktop-icon-view.c
+++ b/src/file-manager/fm-desktop-icon-view.c
@@ -136,8 +136,8 @@ icon_container_set_workarea (CajaIconContainer *icon_container,
         scale = 1; /*wayland handles this for us*/
         GdkRectangle geometry = {0};
         GdkMonitor *monitor;
-        monitor = gdk_display_get_monitor(display, 0);
-        gdk_monitor_get_geometry(monitor,&geometry);
+        monitor = gdk_display_get_monitor (display, 0);
+        gdk_monitor_get_geometry (monitor, &geometry);
         screen_width = geometry.width;
         screen_height = geometry.height;
     }
@@ -434,8 +434,8 @@ realized_callback (GtkWidget *widget, FMDesktopIconView *desktop_icon_view)
         /*No real root window or primary monitor in wayland unless compositors add it back*/
         GdkRectangle geometry = {0};
         GdkMonitor *monitor;
-        monitor = gdk_display_get_monitor(display, 0);
-        gdk_monitor_get_geometry(monitor,&geometry);
+        monitor = gdk_display_get_monitor (display, 0);
+        gdk_monitor_get_geometry (monitor, &geometry);
         allocation.width = geometry.width;
         allocation.height = geometry.height;
     }
--- a/src/file-manager/fm-desktop-icon-view.c
+++ b/src/file-manager/fm-desktop-icon-view.c
@@ -120,13 +120,27 @@ icon_container_set_workarea (CajaIconContainer *icon_container,
     int screen_width, screen_height;
     int scale;
     int i;
+    GdkDisplay *display;
 
     left = right = top = bottom = 0;
 
-    scale = gdk_window_get_scale_factor (gdk_screen_get_root_window (screen));
-
-    screen_width  = WidthOfScreen (gdk_x11_screen_get_xscreen (screen)) / scale;
-    screen_height = HeightOfScreen (gdk_x11_screen_get_xscreen (screen)) / scale;
+    display = gdk_screen_get_display (screen);
+    if (GDK_IS_X11_DISPLAY (display))
+    {
+       scale = gdk_window_get_scale_factor (gdk_screen_get_root_window (screen));
+       screen_width = WidthOfScreen (gdk_x11_screen_get_xscreen (screen)) / scale;
+       screen_height = HeightOfScreen (gdk_x11_screen_get_xscreen (screen)) / scale;
+    }
+    else
+    {
+        scale = 1; /*wayland handles this for us*/
+        GdkRectangle geometry = {0};
+        GdkMonitor *monitor;
+        monitor = gdk_display_get_monitor(display, 0);
+        gdk_monitor_get_geometry(monitor,&geometry);
+        screen_width = geometry.width;
+        screen_height = geometry.height;
+    }
 
     for (i = 0; i < n_items; i += 4)
     {
@@ -148,6 +162,7 @@ icon_container_set_workarea (CajaIconContainer *icon_container,
                                      left, right, top, bottom);
 }
 
+/*This code is only reached when running in x11*/
 static void
 net_workarea_changed (FMDesktopIconView *icon_view,
                       GdkWindow         *window)
@@ -239,6 +254,7 @@ net_workarea_changed (FMDesktopIconView *icon_view,
         g_free (workareas);
 }
 
+/*This code is reached only in x11*/
 static GdkFilterReturn
 desktop_icon_view_property_filter (GdkXEvent *gdk_xevent,
                                    GdkEvent *event,
@@ -258,7 +274,6 @@ desktop_icon_view_property_filter (GdkXEvent *gdk_xevent,
     default:
         break;
     }
-
     return GDK_FILTER_CONTINUE;
 }
 
@@ -317,6 +332,7 @@ fm_desktop_icon_view_class_init (FMDesktopIconViewClass *class)
     FM_ICON_VIEW_CLASS (class)->supports_labels_beside_icons = real_supports_labels_beside_icons;
 }
 
+/*This code is only reached when running on x11*/
 static void
 fm_desktop_icon_view_handle_middle_click (CajaIconContainer *icon_container,
         GdkEventButton *event,
@@ -373,6 +389,9 @@ fm_desktop_icon_view_handle_middle_click (CajaIconContainer *icon_container,
 static void
 unrealized_callback (GtkWidget *widget, FMDesktopIconView *desktop_icon_view)
 {
+    if (!GDK_IS_X11_DISPLAY (gdk_display_get_default()))
+        return;
+
     g_return_if_fail (desktop_icon_view->priv->root_window != NULL);
 
     /* Remove the property filter */
@@ -387,13 +406,13 @@ realized_callback (GtkWidget *widget, FMDesktopIconView *desktop_icon_view)
 {
     GdkWindow *root_window;
     GdkScreen *screen;
+    GdkDisplay *display;
     GtkAllocation allocation;
-    gint scale;
 
     g_return_if_fail (desktop_icon_view->priv->root_window == NULL);
 
     screen = gtk_widget_get_screen (widget);
-    scale = gtk_widget_get_scale_factor (widget);
+    display = gdk_screen_get_display (screen);
 
     /* Ugly HACK for the problem that the views realize at the
      * wrong size and then get resized. (This is a problem with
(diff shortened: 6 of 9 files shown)


lukefromdc:
Out of time tonight but took a copy of the travis errors. Builds finished here fine but travis found an issue in eel-canvas.c that should predate this PR as that file was not modified.

Note that with this and it's matching mate-desktop PR, plus current panel and the applets I am working on, I now have a mate-wayfire session that is becoming increasingly difficult to tell from my x11 session! This was supposed to be the hardest part, and I have working desktop icons in wayland, with the desktop window staying under the others and rendering on all workspaces. Interaction with it seems normal.

This will need to be checked over to make sure NO changes to x11 behavior occur and I've made as few x11 codepath changes as possible for that reason. I have to move a function though to avoid a segfault on trying to detect the monitor too soon. The desktop code is a bit racy, and that caused problems in two areas

lukefromdc:
travis build failures came from an inabiity to find gtk-layer-shell.
If this is a problem in distros wayland support will have to be made optional for those who have this available

lukefromdc:
Mate-Wayland_6--20-2023

This is caja managing a wayland desktop. Note that mate-panel can be made the DEFAULT panel in wayfire

raveit65:
Build dependency gtk-layer-shell is missing in CI.
https://app.travis-ci.com/github/mate-desktop/caja/jobs/604390722#L951
```
configure: error: Package requirements (

    gdk-pixbuf-2.0 >= 2.36.5

    glib-2.0 >= 2.58.1

    mate-desktop-2.0 >= 1.17.3

    gthread-2.0

    gio-unix-2.0

    gio-2.0 >= 2.50.0

    pango >= 1.1.2

    gtk+-3.0 >= 3.22.0

    gtk-layer-shell-0 >= gtk-layer-shell-0_minver

    libnotify

    libxml-2.0 >= 2.4.7

    gail-3.0 >= 3.0.0

) were not met:

Package 'gtk-layer-shell-0', required by 'virtual:world', not found
```
Take a look a mate-panel how it looks. For example debian
https://github.com/mate-desktop/mate-panel/blob/master/.build.yml#L44

lukefromdc:
I just noticed gtk3-devel also missing from Fedora build of caja but present in mate-panel. Will add that too, we can remove it if it's not in fact necessary

lukefromdc:
Looks like all the Travis builds now worked

lukefromdc:
Just added code to use gtk-layer-shell to always anchor the desktop to all four edges, a simple approach to ensuring the desktop always follows any screen resolution changes. 

Also make wayland support optional, so as not to force a gtk-layer-shell dependency on builds not intended for use with wayland. Note that the earlier code to put "x-isms" behind x11-only selectors allows navigation windows to run in wayland without any wayland libs, and using the desktop on wayland (as it is outside a mate-session) requires passing 
`--force-desktop` to caja on invocation

lukefromdc:
Just ran some tests with wlr-randr which does in wlroots based compositors what xrandr does in x11. Always the desktop spanned the monitor, and even fading to the new and correct background image size worked.

lukefromdc:
Just rebased after https://github.com/mate-desktop/caja/pull/1724 merged

lukefromdc:
I had to do some fiddling to get a rational multimonitor setup in wayfire but once I did, caja would render a desktop on the leftmost monitor (my largest and set with wlr-randr to position 0,0) just fine

lukefromdc:
Rebased after https://github.com/mate-desktop/caja/commit/126a4a2a3ad3d93088644406936a2499f7be18be

configure.ac
```diff
@@ -78,7 +79,7 @@ PKG_CHECK_MODULES(ALL, [
     gio-unix-2.0
     gio-2.0 >= gio_minver
     pango >= pango_minver
-    gtk+-3.0 >= gtk_minver
+    gtk-layer-shell-0 >= gtk-layer-shell-0_minver
```
raveit65, line 82: Why gtk+-3.0 dependency is removed?

src/caja-desktop-window.c
```diff
@@ -28,8 +28,10 @@
 #include <gdk/gdkx.h>
 #include <gtk/gtk.h>
 #include <gio/gio.h>
-#include <glib/gi18n.h>
```
raveit65, line 31: Is it correct to remove glib?

lukefromdc:
No it is NOT,, that looks like it wil cause
translation problems!

lukefromdc:
Probably bad cut and paste while bringing the
code over from mate-panel thanks for catching it.

lukefromdc:
brought back
`#include <glib/gi18n.h>`

raveit65:
With a quick test this works for x11 and wayland.
My wayfire desktop:
screenshot-2023-07-02-20:05:37
Seems caja don't handle the desktop from my second monitor.

Sadly, i lost always my scaling factor1.7  when i change something in wayfire.ini or with wayfire-config-tool, which is annoying. Also with session start the scaling factor will be ignored.

But for a first raw implementation it looks good.
Code review i will do the next week.
@mate-desktop/core-team 
Your reviews are wanted!

lukefromdc:
I'm not getting icons on a second monitor with any recent version of caja, including from before I started the wayland work.

raveit65:
Really, this was always working with x11 and marco.

lukefromdc:
Might be a compiz issue in my case, I will check that out when I get home. Also I have  a ton of old caja debs and might be able to find out when it started acting up if it's not compiz in some way. Background renders just fine, I just can't move icons to the 2nd window x11 or Wayland and rolling back to an early 2023 version didn't fix it

eel/eel-background.c
```diff
@@ -433,8 +433,8 @@ set_root_surface (EelBackground *self,
         self->details->bg_surface = mate_bg_create_surface (self->details->bg, window,
         						    width, height, TRUE);
     }
-
-    if (self->details->bg_surface != NULL)
+    GdkDisplay *display = gdk_screen_get_display (screen);
```
raveit65, line 436: indentation for the if clause is misleading, GdkDisplay should start 4 space to the right, exactly under the `s` from line above.

eel/eel-background.c
```diff
@@ -665,8 +665,15 @@ widget_realized_setup (GtkWidget     *widget,
     self->details->screen_monitors_handler =
         g_signal_connect (screen, "monitors-changed", G_CALLBACK (screen_size_changed), self);
 
-    self->details->use_common_surface =
-        (gdk_window_get_visual (window) == gtk_widget_get_visual (widget)) ? TRUE : FALSE;
+    if  (GDK_IS_X11_DISPLAY (display))
```
raveit65, line 668: One space after `if` should be sufficient.

eel/eel-background.c
```diff
@@ -665,8 +665,15 @@ widget_realized_setup (GtkWidget     *widget,
...
+    {
+        GdkWindow *window;
+        window = gdk_screen_get_root_window (screen);
+        self->details->use_common_surface =
+            (gdk_window_get_visual (window) == gtk_widget_get_visual (widget)) ? TRUE : FALSE;
+    }
+    else /*Wayland is always composited*/
+         self->details->use_common_surface = TRUE;
```
raveit65, line 676: Indentation, 4 spaces for the else clause.

src/caja-application.c
```diff
@@ -634,44 +634,48 @@ selection_get_cb (GtkWidget          *widget,
...
+    g_snprintf (selection_name, sizeof (selection_name), "_NET_DESKTOP_MANAGER_S0");
+    selection_atom = gdk_atom_intern (selection_name, FALSE);
+    if  (GDK_IS_X11_DISPLAY (display))
+    {
         selection_owner = XGetSelectionOwner (GDK_DISPLAY_XDISPLAY (display),
                                           gdk_x11_atom_to_xatom_for_display (display,
                                                   selection_atom));
-        if (selection_owner != None)
+
+        if  (selection_owner != None)
```
raveit65, line 650: One space after `if`.

src/caja-application.c
```diff
@@ -634,44 +634,48 @@ selection_get_cb (GtkWidget          *widget,
...
+    {
         if (gtk_selection_owner_set_for_display (display,
             selection_widget,
             selection_atom,
             gdk_x11_get_server_time (gtk_widget_get_window (selection_widget))))
         {
-
             g_signal_connect (selection_widget, "selection_get",
-                          G_CALLBACK (selection_get_cb), NULL);
+                      G_CALLBACK (selection_get_cb), NULL);
```
raveit65, line 667: I would start with G_CALLBACK under the `s` from selection above.

src/caja-desktop-window.c
```diff
@@ -174,10 +191,27 @@ caja_desktop_window_new (CajaApplication *application,
     int width_request, height_request;
     int scale;
 
-    scale = gdk_window_get_scale_factor (gdk_screen_get_root_window (screen));
-
-    width_request = WidthOfScreen (gdk_x11_screen_get_xscreen (screen)) / scale;
-    height_request = HeightOfScreen (gdk_x11_screen_get_xscreen (screen)) / scale;
+    GdkDisplay *display = gdk_screen_get_display (screen);
+    if  (GDK_IS_X11_DISPLAY (display))
```
raveit65, line 195: Too much spaces after the if.

src/caja-desktop-window.c
```diff
@@ -191,11 +225,47 @@ caja_desktop_window_new (CajaApplication *application,
...
+        GtkWindow *gtkwin;
+        gtkwin = (GTK_WINDOW(window));
+
+        /* Before the window is first realized, set it up to be a layer surface */
+        gtk_layer_init_for_window (gtkwin);
+
+        /* Order below normal windows */
+        gtk_layer_set_layer (gtkwin, GTK_LAYER_SHELL_LAYER_BOTTOM);
+
+	    gtk_layer_set_namespace (gtkwin, "desktop");
```
raveit65, line 243: Indentation

src/caja-desktop-window.c
```diff
@@ -191,11 +225,47 @@ caja_desktop_window_new (CajaApplication *application,
...
     GdkWindow *gdkwin;
-    gtk_widget_realize (GTK_WIDGET (window));
+    if  ((GDK_IS_X11_DISPLAY (display)))
+        gtk_widget_realize(GTK_WIDGET (window));
+    else
+        gtk_widget_show_all (GTK_WIDGET (window));
+
     gdkwin = gtk_widget_get_window (GTK_WIDGET (window));
-    if (gdk_window_ensure_native (gdkwin)) {
+    if  ((GDK_IS_X11_DISPLAY (display)) && (gdk_window_ensure_native (gdkwin)))
```
raveit65, line 267: Too much spaces after the if.

src/caja-desktop-window.c
```diff
@@ -230,15 +303,21 @@ unrealize (GtkWidget *widget)
...
 
-    root_window = gdk_screen_get_root_window (
+    /*Avoid root window on wayland-it's not supposed to work*/
+    if  (GDK_IS_X11_DISPLAY (display))
+    {
+        root_window = gdk_screen_get_root_window (
                       gtk_window_get_screen (GTK_WINDOW (window)));
 
-    gdk_property_delete (root_window,
+        gdk_property_delete (root_window,
```
raveit65, line 318: Indentation

src/caja-desktop-window.c
```diff
@@ -252,34 +331,44 @@ unrealize (GtkWidget *widget)
 static void
 set_wmspec_desktop_hint (GdkWindow *window)
 {
-    GdkAtom atom;
+    GdkDisplay *display;
+    display = gdk_screen_get_display (gdk_screen_get_default());
+    if  (GDK_IS_X11_DISPLAY (display))
```
raveit65, line 336: again, too much spaces after the if.

configure.ac
lukefromdc: That's probably a bad cut and paste, will bring it back now

eel/eel-background.c
lukefromdc: look more closely:
`    if (self->details->bg_surface != NULL)`
was removed and replaced by
`    if ((GDK_IS_X11_DISPLAY (display)) && (self->details->bg_surface != NULL))`
in the same position.
It is followed by
`            mate_bg_set_surface_as_root (screen, self->details->bg_surface)`
with the correct indent. A misleading indent btw brings a build warning, though that can easily get buried under all the old warnings in a full build of caja

eel/eel-background.c
lukefromdc: Found that, fixed it. Will push after the rest of these

eel/eel-background.c
lukefromdc: Four spaces between else and the comment is indeed used elsewhere in the same file, adding it here

As for the indentation of the following line, here is an example of the following line from fr-command-unarchiver.c in engrampa, where the code uses tabs but includes another single-line else clause:
```
	if (overwrite)
		fr_process_add_arg (comm->process, "-f");
	else
		fr_process_add_arg (comm->process, "-s");
```
However, in this case that does in fact look odd, but indenting only four spaces might be misleading. Adding brackets for that, but can do something different if you prefer

src/caja-application.c
lukefromdc: Found that and removed the extra line too, that entire section should now be unchanged from master after I push this round of edits

src/caja-application.c
```diff
@@ -634,44 +634,48 @@ selection_get_cb (GtkWidget          *widget,
...
+    }
+    selection_widget = gtk_invisible_new_for_screen (gdk_display_get_default_screen (display));
+    /* We need this for gdk_x11_get_server_time() */
+    gtk_widget_add_events (selection_widget, GDK_PROPERTY_CHANGE_MASK);
 
-        selection_widget = gtk_invisible_new_for_screen (gdk_display_get_default_screen (display));
-        /* We need this for gdk_x11_get_server_time() */
-        gtk_widget_add_events (selection_widget, GDK_PROPERTY_CHANGE_MASK);
-
+    if  (GDK_IS_X11_DISPLAY (display))
```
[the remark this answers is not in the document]
lukefromdc: Found that in TWO places (copied no doubt), fixed them both

src/caja-application.c
lukefromdc: Done and matches next occurance of same

src/caja-desktop-window.c
lukefromdc: There and multiple other places in same file, all found and fixed

src/caja-desktop-window.c
lukefromdc: Found a tab there and fixed it

src/caja-desktop-window.c
```diff
@@ -191,11 +225,47 @@ caja_desktop_window_new (CajaApplication *application,
...
+
+        /*Anchor the desktop to all four corners
+         *This is much simpler than on x11 and
+         *should always render the desktop across
+         *all of the screen
+         */
+         gtk_layer_set_anchor(gtkwin,GTK_LAYER_SHELL_EDGE_TOP, TRUE);
+         gtk_layer_set_anchor(gtkwin,GTK_LAYER_SHELL_EDGE_BOTTOM, TRUE);
+         gtk_layer_set_anchor(gtkwin,GTK_LAYER_SHELL_EDGE_LEFT, TRUE);
+         gtk_layer_set_anchor(gtkwin,GTK_LAYER_SHELL_EDGE_RIGHT, TRUE);
```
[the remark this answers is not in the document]
lukefromdc: These need 8 spaces (they are inside` an if (GDK_IS_WAYLAND_DISPLAY (display))`) clause-but were indented 9 spaces instead of 8, an easy screwup to make after a multiline comment. Fixed again

src/caja-desktop-window.c
lukefromdc: That one moves into an   `  if (GDK_IS_X11_DISPLAY (display))` clause, thus the indent change to match the rest of that clause

src/caja-desktop-window.c
```diff
@@ -252,34 +331,44 @@ unrealize (GtkWidget *widget)
...
+        GdkAtom atom;
 
-    atom = gdk_atom_intern ("_NET_WM_WINDOW_TYPE_DESKTOP", FALSE);
+        atom = gdk_atom_intern ("_NET_WM_WINDOW_TYPE_DESKTOP", FALSE);
 
-    gdk_property_change (window,
+        gdk_property_change (window,
                          gdk_atom_intern ("_NET_WM_WINDOW_TYPE", FALSE),
                          gdk_x11_xatom_to_atom (XA_ATOM), 32,
                          GDK_PROP_MODE_REPLACE, (guchar *) &atom, 1);
```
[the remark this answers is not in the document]
lukefromdc: Found more double-spaced instances of` if (GDK_IS_X11_DISPLAY (display))` no doubt did that once and then copied it all over the files

src/caja-desktop-window.c
```diff
@@ -174,10 +191,27 @@ caja_desktop_window_new (CajaApplication *application,
...
+        *compositors can implement this but as this is written
+        *only a few wayland compositors allow setting a primary monitor
+        *and they all do it differently. For now, use the first monitor
+        */
+        GdkRectangle geometry = {0};
+        GdkMonitor *monitor;
+        monitor = gdk_display_get_monitor(display, 0);
+        gdk_monitor_get_geometry(monitor,&geometry);
+        width_request = geometry.width;
+        height_request = geometry.height;
```
cwendling, line 213: Shouldn't this use *all* monitors?  in X11, a screen is not a monitor, and nowadays usually there's only one screen no matter the number of monitors.
Ideally I guess there should be one desktop window per monitor to fit any kind of layout, but that's probably a bigger change.

src/caja-desktop-window.c
```diff
@@ -191,11 +225,47 @@ caja_desktop_window_new (CajaApplication *application,
...
+        gtk_layer_set_anchor (gtkwin,GTK_LAYER_SHELL_EDGE_RIGHT, TRUE);
+
+        /*Enable keyboard use on the desktop*/
+        gtk_layer_set_keyboard_mode (gtkwin, GTK_LAYER_SHELL_KEYBOARD_MODE_ON_DEMAND);
+    }
+#endif
     /* Special sawmill setting*/
     GdkWindow *gdkwin;
-    gtk_widget_realize (GTK_WIDGET (window));
+    if  ((GDK_IS_X11_DISPLAY (display)))
```
cwendling, line 261: only one space after `if`, and no need to double parentheses around the condition
```suggestion
    if (GDK_IS_X11_DISPLAY (display))
```

src/caja-desktop-window.c
```diff
@@ -191,11 +225,47 @@ caja_desktop_window_new (CajaApplication *application,
...
+        gtk_widget_realize(GTK_WIDGET (window));
```
cwendling, line 262: missing space before parenthesis
```suggestion
        gtk_widget_realize (GTK_WIDGET (window));
```

src/caja-desktop-window.c
```diff
@@ -204,6 +274,9 @@ caja_desktop_window_new (CajaApplication *application,
         XFree(xch);
     }
 
+    gtk_widget_hide (CAJA_WINDOW (window)->details->menubar);
+    gtk_widget_hide (CAJA_WINDOW (window)->details->statusbar);
+
```
cwendling, line 279: this is not needed in the X11 case (that doesn't have the `show_all()`), is it?  Also, shouldn't you use `gtk_widget_show()` above instead?

src/caja-desktop-window.c
```diff
@@ -230,15 +303,21 @@ unrealize (GtkWidget *widget)
     CajaDesktopWindow *window;
     CajaDesktopWindowPrivate *details;
     GdkWindow *root_window;
+    GdkDisplay *display;
 
     window = CAJA_DESKTOP_WINDOW (widget);
     details = window->details;
+    display = gdk_screen_get_display (gdk_screen_get_default());
```
cwendling, line 310:
```suggestion
    display = gtk_widget_get_display (widget);
```
maybe?

src/caja-desktop-window.c
```diff
@@ -230,15 +303,21 @@ unrealize (GtkWidget *widget)
...
-    root_window = gdk_screen_get_root_window (
+    /*Avoid root window on wayland-it's not supposed to work*/
+    if (GDK_IS_X11_DISPLAY (display))
+    {
+        root_window = gdk_screen_get_root_window (
                       gtk_window_get_screen (GTK_WINDOW (window)));
 
-    gdk_property_delete (root_window,
+        gdk_property_delete (root_window,
                          gdk_atom_intern ("CAJA_DESKTOP_WINDOW_ID", TRUE));
```
cwendling, line 319: should align with the opening parenthesis from the line above

src/caja-desktop-window.c
```diff
@@ -252,34 +331,44 @@ unrealize (GtkWidget *widget)
 static void
 set_wmspec_desktop_hint (GdkWindow *window)
 {
-    GdkAtom atom;
+    GdkDisplay *display;
+    display = gdk_screen_get_display (gdk_screen_get_default());
+    if (GDK_IS_X11_DISPLAY (display))
```
cwendling, line 336: maybe
```suggestion
    if (GDK_IS_X11_WINDOW (window))
```

cwendling: Actually the whole function only does X11 stuff, so maybe only use it there?  or return early instead of adding an additional indentation layer?

src/caja-desktop-window.c
```diff
@@ -252,34 +331,44 @@ unrealize (GtkWidget *widget)
...
 set_desktop_window_id (CajaDesktopWindow *window,
                        GdkWindow             *gdkwindow)
 {
-    /* Tuck the desktop windows xid in the root to indicate we own the desktop.
+    /* Tuck the desktop windows xid in the root to indicate we own the desktop in on x11
      */
-    Window window_xid;
-    GdkWindow *root_window;
+    GdkDisplay *display = gdk_screen_get_display (gdk_screen_get_default());
+    if (GDK_IS_X11_DISPLAY (display))
```
cwendling, line 357: Again, probably worth using
```suggestion
    if (GDK_IS_X11_WINDOW (gdkwindow))
```

src/caja-desktop-window.c
```diff
@@ -252,34 +331,44 @@ unrealize (GtkWidget *widget)
...
 
-    window_xid = GDK_WINDOW_XID (gdkwindow);
-
-    gdk_property_change (root_window,
+        gdk_property_change (root_window,
                          gdk_atom_intern ("CAJA_DESKTOP_WINDOW_ID", FALSE),
                          gdk_x11_xatom_to_atom (XA_WINDOW), 32,
                          GDK_PROP_MODE_REPLACE, (guchar *) &window_xid, 1);
+    }
+
```
cwendling, line 371: useless blank line

src/caja-desktop-window.c
```diff
@@ -291,9 +380,12 @@ realize (GtkWidget *widget)
     details = window->details;
 
     /* Make sure we get keyboard events */
-    gtk_widget_set_events (widget, gtk_widget_get_events (widget)
+    GdkDisplay *display = gdk_screen_get_display (gdk_screen_get_default());
+    if (GDK_IS_X11_DISPLAY (display))
+        gtk_widget_set_events (widget, gtk_widget_get_events (widget)
                            | GDK_KEY_PRESS_MASK | GDK_KEY_RELEASE_MASK);
```
cwendling, line 386: add 4 spaces to align the start of the line with the opening parenthesis:
```suggestion
        gtk_widget_set_events (widget, gtk_widget_get_events (widget)
                               | GDK_KEY_PRESS_MASK | GDK_KEY_RELEASE_MASK);
```

src/file-manager/fm-desktop-icon-view.c
```diff
@@ -244,21 +260,24 @@ desktop_icon_view_property_filter (GdkXEvent *gdk_xevent,
                                    GdkEvent *event,
                                    gpointer data)
 {
-    XEvent *xevent = gdk_xevent;
-    FMDesktopIconView *icon_view;
+    GdkDisplay *display = gdk_display_get_default();
+    if (GDK_IS_X11_DISPLAY (display))
```
cwendling, line 264: this again doesn't make much sens, it's an X11-specific function it seems, just call it there only?

src/file-manager/fm-desktop-icon-view.c
```diff
@@ -510,6 +553,13 @@ delayed_init (FMDesktopIconView *desktop_icon_view)
                              "done_loading",
                              G_CALLBACK (done_loading), desktop_icon_view, 0);
 
+    /*Handle the middle click on x11 */
+    if (GDK_IS_X11_DISPLAY (gdk_display_get_default))
+    {
+        g_signal_connect_object (desktop_icon_view, "middle_click",
+                             G_CALLBACK (fm_desktop_icon_view_handle_middle_click), desktop_icon_view, 0);
```
cwendling, line 560:
```suggestion
        g_signal_connect_object (desktop_icon_view, "middle_click",
                                 G_CALLBACK (fm_desktop_icon_view_handle_middle_click), desktop_icon_view, 0);
```

src/file-manager/fm-desktop-icon-view.c
```diff
@@ -510,6 +553,13 @@ delayed_init (FMDesktopIconView *desktop_icon_view)
                              "done_loading",
                              G_CALLBACK (done_loading), desktop_icon_view, 0);
 
+    /*Handle the middle click on x11 */
+    if (GDK_IS_X11_DISPLAY (gdk_display_get_default))
```
cwendling, line 557: This is dangerously wrong :)  And breaks the X11 feature I guess
```suggestion
    if (GDK_IS_X11_DISPLAY (gdk_display_get_default ()))
```

src/file-manager/fm-desktop-icon-view.c
```diff
@@ -510,6 +553,13 @@ delayed_init (FMDesktopIconView *desktop_icon_view)
                              "done_loading",
                              G_CALLBACK (done_loading), desktop_icon_view, 0);
 
+    /*Handle the middle click on x11 */
```
cwendling, line 556: is it OK to move this here from `fm_desktop_icon_view_init()`? It probably is, but given the test below was totally wrong it hasn't been tested

cwendling (changes requested):

eel/eel-background.c
raveit65: opps, my fault. The line starting with `if` was removed. sorry for that noise.

src/caja-desktop-window.c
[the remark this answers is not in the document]
raveit65: I counted the spaces from the point where `if` begins.
Any way, you fixed it, thx.
Can you please add a space between `gtkwin,` and `GTK_Layer_SHELL_.....`?
Like this for example:
`gtk_layer_set_anchor (gtkwin, GTK_LAYER_SHELL_EDGE_TOP, TRUE);`

src/caja-desktop-window.c
raveit65: Sorry, my fault. But the next Line should be moved too or not.... :)

raveit65:
I will check again code-format when all code changes are done.
During testing i found an issue with caja. `open-in-terminal` command from content menu crashes caja in wayfire. From desktop or inside the browser window.

src/caja-desktop-window.c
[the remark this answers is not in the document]
raveit65: Move the marked lines too, should look like this.
```
        gdk_property_change (root_window,
                             gdk_atom_intern ("CAJA_DESKTOP_WINDOW_ID", FALSE),
                             gdk_x11_xatom_to_atom (XA_WINDOW), 32,
                             GDK_PROP_MODE_REPLACE, (guchar *) &window_xid, 1);
```

src/file-manager/fm-desktop-icon-view.c
lukefromdc: I have no idea and have tried to avoid deliberate changes in the x11 codepath. That would be for another PR is my guess, in case it ever had to be reverted. Any one commt in this would be combined with the rest in a squash and merge

lukefromdc: I will check to ensure that wasn't somehow moved by mistake and will move it back if it was

lukefromdc: Here's what I found: for some reason, running`  if (GDK_IS_X11_DISPLAY (gdk_display_get_default))`  in `fm_desktop_icon_view_init()` for some reason could not access the display and segfaulted. I returned that code to it's original location (so as to restore the old x11 order of things) and moved the x11 check to the middle click handling function itself and that works

lukefromdc: Moving the code back and putting the x11 check in fm_desktop_icon_view_handle_middle_click generated this warning when running in Wayland:

`sys:1: Warning: ../gobject/gsignal.c:2517: signal 'middle_click' is invalid for instance '0x5614b513e5c0' of type 'FMDesktopIconView'
`

lukefromdc:
I was able to successfully open terminals using the open-in-terminal menu item

src/file-manager/fm-desktop-icon-view.c
lukefromdc: It is indeed called only when in x11, so marked it as such with a comment and removed the check

src/file-manager/fm-desktop-icon-view.c
lukefromdc: applied

src/file-manager/fm-desktop-icon-view.c
lukefromdc: That was a cut and paste from elsewhere, maybe you just found my problem running this earlier

lukefromdc: YES! That was it-now I can move this back to it's original place

src/file-manager/fm-desktop-icon-view.c
lukefromdc: As described above, the problem that originally forced this to move was accidently using
` if (GDK_IS_X11_DISPLAY (gdk_display_get_default))`
instead of the intended
` if (GDK_IS_X11_DISPLAY (gdk_display_get_default()))`
and for some reason the compiler didn't stop the build and return an error on that

configure.ac
```diff
@@ -13,6 +13,7 @@ m4_define(exif_minver,                 0.6.14)
 m4_define(exempi_minver,               1.99.5)
 m4_define(gail_minver,                 3.0.0)
 m4_define(notify_minver,               0.7.0)
+m4_define(gtk-layer-shell-0_minver,      0.8.0)
```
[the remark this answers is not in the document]
lukefromdc: That works so long as gtk-layer-shell-0 is specified in the other places

configure.ac
```diff
@@ -79,6 +80,7 @@ PKG_CHECK_MODULES(ALL, [
     gio-2.0 >= gio_minver
     pango >= pango_minver
     gtk+-3.0 >= gtk_minver
+    gtk-layer-shell-0 >= gtk-layer-shell-0_minver
```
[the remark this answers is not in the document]
lukefromdc: Moving that to  `PKG_CHECK_MODULES(WAYLAND` ... works fine

configure.ac
```diff
@@ -177,7 +179,37 @@ if test "x$enable_xmp" != "xno"; then
...
+              [AS_HELP_STRING([--enable-wayland],
+                              [Explicitly enable or disable Wayland support
+                              (default is to enable only if Wayland client development library is detected)])],
+              [enable_wayland=$enableval],
+              [enable_wayland=auto])
+
+# Check if we have gtk-layer-shell installed, and thus should build with Wayland support
+have_wayland=no
+if test "x$enable_wayland" != "xno"; then
+  PKG_CHECK_MODULES(WAYLAND, gtk-layer-shell-0 wayland-client, have_wayland=yes, [
```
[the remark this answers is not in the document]
lukefromdc: Just added that with the prior move to stop an unconditional dependency on gtk-layer-shell

configure.ac
```diff
@@ -177,7 +179,37 @@ if test "x$enable_xmp" != "xno"; then
...
+        if test "x$enable_wayland" = "xyes"; then
+          AC_MSG_ERROR([Wayland enabled but GTK Layer Shell library not found])
```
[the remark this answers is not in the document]
lukefromdc: This is the part of the code I know the least about: the build system

configure.ac
```diff
@@ -177,7 +179,37 @@ if test "x$enable_xmp" != "xno"; then
...
+fi
+
+AM_CONDITIONAL(ENABLE_WAYLAND, [test "x$have_wayland" = "xyes"])
+
+if test "x$have_wayland" = "xyes"; then
+  AC_DEFINE(HAVE_WAYLAND, 1, [Have the Wayland development library])
+fi
+
+AC_SUBST(WAYLAND_CFLAGS)
+AC_SUBST(WAYLAND_LIBS)
```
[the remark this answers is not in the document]
lukefromdc: Indeed, a configure and build without them worked

configure.ac
```diff
@@ -240,7 +272,7 @@ AC_SUBST(LIBCAJA_EXTENSION_LIBS)
 
 dnl core caja
 PKG_CHECK_MODULES(GMODULE, gmodule-2.0, [GMODULE_ADD="gmodule-2.0"],[GMODULE_ADD=""])
-CORE_MODULES="glib-2.0 $GMODULE_ADD mate-desktop-2.0 gthread-2.0 gio-2.0 gio-unix-2.0 gail-3.0 libxml-2.0 $EXTRA_CORE_MODULES gtk+-3.0 x11"
+CORE_MODULES="glib-2.0 $GMODULE_ADD mate-desktop-2.0 gthread-2.0 gio-2.0 gio-unix-2.0 gail-3.0 libxml-2.0 gtk-layer-shell-0 $EXTRA_CORE_MODULES gtk+-3.0 x11"
```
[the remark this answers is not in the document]
lukefromdc: Build failure without it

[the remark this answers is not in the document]
lukefromdc: `EXTRA_CORE_MODULES` turned out to be unused, found nowhere else in the entire source directory nor in configure.ac
I found that as things stood, non-wayland builds were broken. Looks like `EXTRA_CORE_MODULES` was keeping gtk from being added to `CORE_MODULES` when gtk-layer-shell wasn't also part of it, so checked to see if anything used it, and removed it on finding no other references to it. Then had to add wayland libs in src/Makefile.am since they were no longer included in core libs. That worked

Tested three build cases after all this: 
1: no gtk-layer-shell, attempt to enable wayland. Produced the error stating that gtk-layer-shell is required for wayland
2: no gtk-layer-shell, build with --disable-wayland, successful configuration and build 
3: gtk-layer-shell installed, build with --enable-wayland, successful configuration and build

eel/eel-background.c
```diff
@@ -649,7 +649,7 @@ widget_realized_setup (GtkWidget     *widget,
     }
 
     GdkScreen *screen = gtk_widget_get_screen (widget);
-    GdkWindow *window = gdk_screen_get_root_window (screen);
+    GdkDisplay *display = gdk_screen_get_display (gdk_screen_get_default());
```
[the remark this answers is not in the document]
lukefromdc: Fixed

eel/eel-background.c
```diff
@@ -665,8 +665,17 @@ widget_realized_setup (GtkWidget     *widget,
     self->details->screen_monitors_handler =
         g_signal_connect (screen, "monitors-changed", G_CALLBACK (screen_size_changed), self);
 
-    self->details->use_common_surface =
-        (gdk_window_get_visual (window) == gtk_widget_get_visual (widget)) ? TRUE : FALSE;
+    if (GDK_IS_X11_DISPLAY (display))
+    {
+        GdkWindow *window;
+        window = gdk_screen_get_root_window (screen);
```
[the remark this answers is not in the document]
lukefromdc: Changed. I take it this is consided the preferable format to  declare a variable inside this smaller scope, well down in the function?

src/caja-application.c
```diff
@@ -634,44 +634,47 @@ selection_get_cb (GtkWidget          *widget,
...
-
             g_signal_connect (selection_widget, "selection_get",
-                          G_CALLBACK (selection_get_cb), NULL);
+                              G_CALLBACK (selection_get_cb), NULL);
             return selection_widget;
         }
-
-        gtk_widget_destroy (selection_widget);
     }
+    else if (!GDK_IS_X11_DISPLAY (display))
```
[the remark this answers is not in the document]
lukefromdc: Removed

src/caja-desktop-window.c
```diff
@@ -156,9 +159,23 @@ caja_desktop_window_screen_size_changed (GdkScreen             *screen,
...
+    GdkDisplay *display = gdk_screen_get_display (screen);
+    if (GDK_IS_X11_DISPLAY (display))
+    {
+        GdkWindow *root_window;
+        root_window = gdk_screen_get_root_window (screen);
+        gdk_window_get_geometry (root_window, NULL, NULL, &width_request, &height_request);
+    }
+    else
+    {
+        /*No root window or primary monitor in wayland unless compositors add it back*/
```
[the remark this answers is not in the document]
lukefromdc: In wayland, there is no defined primary monitor. Even if this works it's not documented as working. At it stands now, in a multimonitor setup (as I've been testing) caja's desktop will open on the same screen the terminal you are calling it with --force-desktop is on if called from terminal. If called by wayfire's config.ini file it will come up on the leftmost monitor, the same one mate-panel comes up on if also called from wayfire.ini.

If we were to look for a particular monitor in both cases, not only wayfire but all other supported compositors would have to have a standard way to pass that to caja and mate panel.

It seems that the leftmost (position 0,0) becomes the default primary monitor in wayfire. Only way to change this I have seen so far  is to open a terminal, kill the panel and caja, move the terminal to the other screen,and either open a second terminal to run` mate-panel `and `caja --force desktop ` each from one of them, or to use that terminal to run a script to open both while the terminal is on the desired screen

src/caja-desktop-window.c
lukefromdc: The background is in fact drawn (by wayfire itself) across all the monitors. You can only put icons on one of them, and for some reason, recent (pre-wayland) versions of caja seem to have an issue with trying to move icons to a second monitor even in x11.

GTK has deprecated all of the ways of getting a screen's geometry and in fact has removed GtkScreen from GTK4 it seems, thus the deprecations in GTK 3. In wayland I am getting some inconsistant behavior with monitors as well:
if I have two monitors active and call caja with something like `--geometry 4000x4000` it will extend partially to the second monitor and partially overflow the other side of the first. I do NOT know how to fix this, but this is not a common use case I would imagine.

If someone knows a better way I am open to it, but we have  a consensus against adding new deprecated code.

This can wait for another PR if necessary.

src/caja-desktop-window.c
lukefromdc: found and fixed

src/caja-desktop-window.c
lukefromdc: Found and fixed

src/caja-desktop-window.c
lukefromdc: gtk_widget_show() works as suggested, and keeps the statusbar off the desktop.

src/caja-desktop-window.c
lukefromdc: Didn't know we had that, will try it

lukefromdc: And in fact that works, saving us one variable (no matter how the source code calls it, the compiler still has to allocate it)

src/caja-desktop-window.c
lukefromdc: Using  `if (GDK_IS_X11_WINDOW (window)) ` let the build finish without a warning, but caused one of those functions not to run in x11, which allowed the desktop to be rendered over other windows, no longer always on the bottom. Probably not getting the actual window

src/caja-desktop-window.c
lukefromdc: Removing the check here, moving it to where this is called

src/caja-desktop-window.c
lukefromdc: In most of the existing code, first letter of second line of something like this is under the first letter after the (  and I've been advised elsewhere to align it that way, so doing that here too

src/caja-desktop-window.c
lukefromdc: removed

src/caja-desktop-window.c
lukefromdc: Found and fixed

lukefromdc:
Changes pushed and ready for testing

src/caja-desktop-window.c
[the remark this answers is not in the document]
lukefromdc: Done

src/caja-desktop-window.c
lukefromdc: First letter of 2nd line under first letter of 1st line is how that is aligned now, and how similar things elsewhere in the code are formatted

src/caja-desktop-window.c
lukefromdc: set_desktop_window_id  and set_wmspec_desktop_hint are now called only from inside  a 
`    if (GDK_IS_X11_DISPLAY (display))`
block in` realize`
unfortunately we had to use display there to get it to work, otherwise the x11 desktop sometimes rendered on top of other windows

src/caja-desktop-window.c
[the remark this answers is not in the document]
lukefromdc: That's what we now have

src/caja-desktop-window.c
cwendling: > Using `if (GDK_IS_X11_WINDOW (window)) ` let the build finish without a warning, but caused one of those functions not to run in x11, which allowed the desktop to be rendered over other windows, no longer always on the bottom. Probably not getting the actual window.

Okay, so that wasn't a so great idea in the end.  Weird, but I trust the test.

src/caja-desktop-window.c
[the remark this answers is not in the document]
cwendling: OK… well, supporting multi-monitor desktop can be a later improvement anyway

src/caja-desktop-window.c
cwendling: It's bizarre it wouldn't work on X11, but if it's not this PR I'm OK to fix this later

src/caja-desktop-window.c
```diff
@@ -204,6 +274,8 @@ caja_desktop_window_new (CajaApplication *application,
         XFree(xch);
     }
 
+    gtk_widget_hide (CAJA_WINDOW (window)->details->menubar);
+
```
cwendling, line 278: Is this still needed?

src/caja-desktop-window.c
cwendling: That looks good to me now

src/caja-desktop-window.c
```diff
@@ -289,17 +368,23 @@ realize (GtkWidget *widget)
...
     window = CAJA_DESKTOP_WINDOW (widget);
     details = window->details;
+    GdkDisplay *display;
 
     /* Make sure we get keyboard events */
-    gtk_widget_set_events (widget, gtk_widget_get_events (widget)
-                           | GDK_KEY_PRESS_MASK | GDK_KEY_RELEASE_MASK);
-    /* Do the work of realizing. */
-    GTK_WIDGET_CLASS (caja_desktop_window_parent_class)->realize (widget);
+    if (GDK_IS_X11_WINDOW (GDK_WINDOW(widget)))
```
cwendling, line 374: This is not right, `widget` is not a `GdkWindow`.  And as you use the display a little bit down, just use that here as well

src/file-manager/fm-desktop-icon-view.c
```diff
@@ -169,6 +183,8 @@ net_workarea_changed (FMDesktopIconView *icon_view,
      * elements in the workareas array describing
      * x,y,width,height) */
     display = gtk_widget_get_display (GTK_WIDGET (icon_view));
+    if  (!GDK_IS_X11_DISPLAY (display))
+       return;
```
cwendling, line 187: This whole function does nothing but on X11, so you should remove this and rather make the function itself (`net_workarea_changed()`) only called on X11

eel/eel-background.c
raveit65: I do not understand why using brackets for the else clause wit one line, but ok.
Indentation is fixed.

src/file-manager/fm-desktop-icon-view.c
```diff
@@ -239,11 +255,14 @@ net_workarea_changed (FMDesktopIconView *icon_view,
...
 }
 
+/*This code is reached only in x11*/
 static GdkFilterReturn
 desktop_icon_view_property_filter (GdkXEvent *gdk_xevent,
                                    GdkEvent *event,
                                    gpointer data)
 {
+    GdkDisplay *display = gdk_display_get_default();
+
```
cwendling, line 265: Useless now, just remove those two lines

src/file-manager/fm-desktop-icon-view.c
```diff
@@ -373,6 +392,9 @@ fm_desktop_icon_view_handle_middle_click (CajaIconContainer *icon_container,
 static void
 unrealized_callback (GtkWidget *widget, FMDesktopIconView *desktop_icon_view)
 {
+    if (!GDK_IS_X11_DISPLAY (gdk_display_get_default))
+        return;
+
```
cwendling, line 397: The test is wrong (should again be `gdk_display_get_default ()`), but more importantly the whole function should only be called on X11 if it's not used anywhere else.

Or if we want to keep the realize/unrealize pair still, maybe wrap this in an `if () { }` instead to suggest there could be potential?  Either way is fine though, but you need to fix the test.

src/file-manager/fm-desktop-icon-view.c
```diff
@@ -387,13 +409,21 @@ realized_callback (GtkWidget *widget, FMDesktopIconView *desktop_icon_view)
...
-    gint scale;
 
     g_return_if_fail (desktop_icon_view->priv->root_window == NULL);
 
     screen = gtk_widget_get_screen (widget);
-    scale = gtk_widget_get_scale_factor (widget);
+    display = gdk_screen_get_display (screen);
+    if (GDK_IS_X11_DISPLAY (display))
+        root_window = gdk_screen_get_root_window (screen);
+
```
cwendling, line 421: remove this empty line

src/file-manager/fm-desktop-icon-view.c
```diff
@@ -387,13 +409,21 @@ realized_callback (GtkWidget *widget, FMDesktopIconView *desktop_icon_view)
...
+    else
+    {
+        /*Do not use the unavailable root window*/
+        desktop_icon_view->priv->root_window = NULL;
+    }
```
cwendling, line 426: This is redundant with what you have below L457+, you can just remove it I think (and maybe move the Wayland comment down there)

src/file-manager/fm-desktop-icon-view.c
```diff
@@ -610,8 +662,13 @@ fm_desktop_icon_view_init (FMDesktopIconView *desktop_icon_view)
                                          CAJA_ICON_LAYOUT_T_B_R_L :
                                          CAJA_ICON_LAYOUT_T_B_L_R);
 
-    g_signal_connect_object (icon_container, "middle_click",
-                             G_CALLBACK (fm_desktop_icon_view_handle_middle_click), desktop_icon_view, 0);
+    /*Handle the middle click on x11 */
+    if (GDK_IS_X11_DISPLAY (gdk_display_get_default()))
```
cwendling, line 666:
```suggestion
    if (GDK_IS_X11_DISPLAY (gdk_display_get_default ()))
```

cwendling (changes requested):

configure.ac
```diff
@@ -13,6 +13,7 @@ m4_define(exif_minver,                 0.6.14)
 m4_define(exempi_minver,               1.99.5)
 m4_define(gail_minver,                 3.0.0)
 m4_define(notify_minver,               0.7.0)
+m4_define(gtk-layer-shell_minver,      0.8.0)
```
cwendling, line 16: I checked and this actually doesn't work as expected.  See https://github.com/cwendling/caja/commit/8e65f5e489fd57c5f9736274d47808ce52131806 for a fix (which is in branch https://github.com/cwendling/caja/commits/wayland-desktop-followup)

configure.ac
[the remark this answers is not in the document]
cwendling: I suggest you use the two additional commits I just pushed to https://github.com/cwendling/caja/commits/wayland-desktop-followup

raveit65:
offtop on
Caja crashes for me when using open-terminal extension in wayland.
I mean this menu entry. (sorry german,but it should be clear, screenshot from X11)
caja_open_terminal
Still wondering that it works for you.
But i have a stacktrace, starts with libxkb !
https://www.dropbox.com/s/z089gb5kadj0cni/backtrace_caja_wayland_open_terminal?dl=0
Maybe you guys could take a look at it?
Should i open a report for it?
offtopic off

Sunderland93:
> Should i open a report for it?

I can confirm same crash on Debian 12 on Wayland (caja built from current git master)

cwendling:
Looks a bit like libstartup-notification is choosing an X backend and crashes somehow…  I can't tell much more like that, but it's a bit weird… not saying it should work, but XWayland should probably kick in and pretend everything's alright?

cwendling:
And yeah, that should probably be another report, it looks like an issue either in libstartup-notification or in libmate-desktop, but doesn't seem like a bug in Caja itself

lukefromdc:
At leaat in Debian, wayfire is being shipped with xwayland disabled for some reason.To switch my whole session over will require a local builf if this does not change, so as to allow kdenlive and audacious to run.

A dev build of GIMP using gtk3 would presumably work

src/caja-desktop-window.c
lukefromdc: Looks like it is not: we have
```
    gtk_widget_hide (CAJA_WINDOW (window)->details->statusbar);
    gtk_widget_hide (CAJA_WINDOW (window)->details->menubar);
```
in caja_desktop_window_init and that works unless `gtk_widget_show_all` is called somewhere. The x11 code uses `gtk_widget_realize()` and due to a problem with that in wayland we use g`tk_widget_show` as using `gtk_widget_realize()` in wayland with and only with gtk-layer-shell gets us this error and blocks showing the desktop:
`
(caja:30733): Gdk-WARNING **: 13:52:09.061: (../gdk/wayland/gdkwindow-wayland.c:915):buffer_release_callback: runtime check failed: (impl->staging_cairo_surface != cairo_surface)`

With gtk_widget_show we avoid that, but if we use  gtk_widget_show_all() we have to re-hide the menubar and statusbar. 

Thus it seems safe to remove the second instance of 
`gtk_widget_hide (CAJA_WINDOW (window)->details->menubar);`
from the x11 code as well. Can put it back if others want me to so as to avoid any change in the x11 code

src/caja-desktop-window.c
lukefromdc: Tossing that, reverting to the display check

src/file-manager/fm-desktop-icon-view.c
lukefromdc: Turns out I can remove line 186 and 187 w no other changes

This is called in threee places total,  all already behind x11-only conditionals. One is in
`desktop_icon_view_property_filter()` which in turn is called only from inside an x11-only block in 
`realized_callback () `and another x11-only block in `unrealized_callback ()`

It is also called directly from the same x11-only block in 
`realized_callback () ` in which `desktop_icon_view_property_filter()` is called
which means when running x11 it gets called twice, only five non-blank lines apart
and in wayland is not reached at all. The x11 codepath is same as before, just put
into the x11 only conditional.

Asserts could be used to catch future programming errors calling the x11-only code
from wayland, the comments would tell what functions to put them in. That's still more code to run however

src/file-manager/fm-desktop-icon-view.c
lukefromdc: Done

src/file-manager/fm-desktop-icon-view.c
lukefromdc: Just fixed that one, it was a cut and paste from the same problem elsewhere fixed yesterday.

src/file-manager/fm-desktop-icon-view.c
lukefromdc: Build went OK with that empty line removed, I had been under the impression that it was required after a one line `if()` block not enclosed in brackets

src/file-manager/fm-desktop-icon-view.c
lukefromdc: Just saw that and removed the redundent block

src/file-manager/fm-desktop-icon-view.c
lukefromdc: Done

src/file-manager/fm-desktop-icon-view.c
lukefromdc: Fixing this one also stopped the segfault on calling `caja --quit` in wayland with the desktop showing, which was exactly the condition this was intended to stop but it didn't work when I screwed up the conditional

configure.ac
lukefromdc: Switched over to your latest version of configure.ac, will simply apply those two commits as one by hand. Here is the error message yielded by passing --enable-wayland without gtk-layer-shell-dev (in Debian) installed:
```

Package 'gtk-layer-shell-0', required by 'virtual:world', not found

Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.

Alternatively, you may set the environment variables WAYLAND_CFLAGS
and WAYLAND_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.
```
Both wayland and non-wayland builds went fine

src/file-manager/fm-desktop-icon-view.c
cwendling: It's not, C doesn't care about blanks but for separating tokens. You could write all your C file in one line, that'd be a mess but it'd work just fine -- assuming you could debug it though :D

configure.ac
cwendling: Cool :+1:
That's exactly what I'm talking about with the error message: it actually tells you how you could fix it if it's not merely a missing dependency, but that pkg-config cannot find it.

src/file-manager/fm-desktop-icon-view.c
lukefromdc: Thanks

configure.ac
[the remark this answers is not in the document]
lukefromdc: Done, thanks for the assist on this

src/caja-desktop-window.c
cwendling: What do you mean?  It wasn't in the original X11 code path, was it?

The reason why `show_all()` caused a problem is that what it does is recursively call `show()` on itself and all public children, which we don't want here.

lukefromdc:
Next round of changes pushed and ready for testing

eel/eel-editable-label.c
```diff
@@ -1039,7 +1038,6 @@ eel_editable_label_ensure_layout (EelEditableLabel *label,
                 longest_paragraph = width;
 
                 wrap_width = get_label_wrap_width (label);
-                scale = gtk_widget_get_scale_factor (widget);
```
cwendling, line 1042: this has nothing to do with this changeset, but looks unused indeed.

src/caja-desktop-window.c
```diff
@@ -289,17 +366,24 @@ realize (GtkWidget *widget)
...
+    GTK_WIDGET_CLASS (caja_desktop_window_parent_class)->realize (widget);
 
-    set_desktop_window_id (window, gtk_widget_get_window (widget));
+    /* This is the new way to set up the desktop window in x11 but not for wayland */
+    display = gtk_widget_get_display (widget);
+    if (GDK_IS_X11_DISPLAY (display))
+        {
+            set_wmspec_desktop_hint (gtk_widget_get_window (widget));
+            set_desktop_window_id (window, gtk_widget_get_window (widget));
+        }
```
cwendling, line 386: One level of indentation too much
```suggestion
    if (GDK_IS_X11_DISPLAY (display))
    {
        set_wmspec_desktop_hint (gtk_widget_get_window (widget));
        set_desktop_window_id (window, gtk_widget_get_window (widget));
    }
```

cwendling (commented):
Apart from the couple of style issue that could be improved, code looks good.  Not tested though.

cwendling (approved):

src/caja-desktop-window.c
lukefromdc: Found and fixed, thanks

eel/eel-editable-label.c
lukefromdc: If we need to I can split that out as another PR, but as the build warning kept popping up I figured best to catch it now, as there's no possible harm in removing a variable that is in fact unused

cwendling: I think it can stay; I usually don't like mixing unrelated stuff in commits & PRs, but that one looks really iniquitous.

> there's no possible harm in removing a variable that is in fact unused

Although it's most likely true here, that's not necessarily always the case. The biggest risk is actually removing the function call that could have side effects you didn't expect.  This said, doesn't matter here.

src/caja-desktop-window.c
lukefromdc: On x11 calling caja oversize such as` --geometry 4000x4000` gives an undecorated fullscreen window exactly filing the primary monitor, it's only on wayland  and then only with a second monitor running that it comes up oversize. Calling caja from a terminal on a secondary monitor in x11 works ONLY if called oversize, and this is with efforts to avoid changing the x11 codepath.

On wayland, caja invoked from terminal appears on the monitor the terminal is on,but can be rendered oversize if and only if a second monitor is active.

raveit65:
Did test it again today and crash in X11 is gone.
I suggest to wait some days before merging, so that PR can be better tested in daily use on my system or by other testers.

In general i don't think that wayfire compositor is a stable platform for MATE on wayland. There config system based on a text file doesn't work very well here. I don't think that all settings in `wayfire.ini` will be applied with starting the session. For example the scaling factor for HIDPI is still ignored here or i did mistakes in my setup.
Also, i don't think that all there effects like the ugly wobbly windows or the cube are really needed for a clean and silent production system ;)
When i found some time i will take a look at mir or sway.
Anyway, nice job with this PR
It's a good starting point for caja on wayland.

lukefromdc:
Thanks, I think wayfire will be rather like compiz was in the 2008 era:  not the baseline compositor for the reasons you describe but one that can be substituted.

Note that the wlr-randr package can change the monitor resolution at runtime, and it would not need to autostart anything started by a MATE-wayland session, so it should be drop-in compatable with any other wlroots-based compositor. GNOME 2 and MATE therafter on x11 have always been modular in this way.

No problem with running this a while before merge, so as to find hidden problems

zhuyaliang (commented):
```
monitor = gdk_display_get_monitor(display, 0);
gdk_monitor_get_geometry(monitor,&geometry);
```
Unified code format, spaces need to be added

lukefromdc:
Found and fixed that last style issue, thanks

zhuyaliang (approved):
Good code format

raveit65:
> Looks a bit like libstartup-notification is choosing an X backend and crashes somehow… I can't tell much more like that, but it's a bit weird… not saying it should work, but XWayland should probably kick in and pretend everything's alright?

Open report https://github.com/mate-desktop/caja/issues/1731

lukefromdc:
For some reason at least Debian is shipping wayfire without xwayland support

L-U-T-i:
PR #1730 cannot be applied after this PR - the problematic part of this PR is:.
```
From 2d1c397a4ccd161b255282a908b2aa0dbb9d4317 Mon Sep 17 00:00:00 2001
From: lukefromdc <lukefromdc@hushmail.com>
Date: Wed, 5 Jul 2023 14:53:08 -0400
Subject: [PATCH 2/4] eel-editable-label-c: remove unused variable

This was leftover from earlier code compatable only with x11
---
 eel/eel-editable-label.c | 2 --
 1 file changed, 2 deletions(-)

diff --git a/eel/eel-editable-label.c b/eel/eel-editable-label.c
index 92ce70154..7c4bcb38f 100644
--- a/eel/eel-editable-label.c
+++ b/eel/eel-editable-label.c
@@ -1028,7 +1028,6 @@ eel_editable_label_ensure_layout (EelEditableLabel *label,
             else
             {
                 gint wrap_width;
-                gint scale;
 
                 pango_layout_set_width (label->layout, -1);
                 pango_layout_get_extents (label->layout, NULL, &logical_rect);
@@ -1039,7 +1038,6 @@ eel_editable_label_ensure_layout (EelEditableLabel *label,
                 longest_paragraph = width;
 
                 wrap_width = get_label_wrap_width (label);
-                scale = gtk_widget_get_scale_factor (widget);
                 width = MIN (width, wrap_width);
                 width = MIN (width,
                 PANGO_SCALE * (gdk_screen_width () + 1) / 2);
```

Is it OK that we remove 'scale' here? I believe 'scale' is still needed, considering we use it in X11 code...

lukefromdc:
I can manually remake https://github.com/mate-desktop/caja/pull/1730 after this, no problem

L-U-T-i:
But if you apply #1730 after that, it needs scale (so it is not unused anymore...).

L-U-T-i:
I've manually modified this patch to keep scale (so not to modify eel/eel-editable-label.c at all) and still had to modify #1730 as well (adapt for a modified eel/eel-gtk-extensions.c).

raveit65:
GDK knows about scaling. Only using Width/HeightOfScreen, which are xlib macros, needs the scale variable.
The goal should be to avoid re-adding Width/HeightOfScreen and replacing gdk stuff, which can be use for x11 and wayland.
So current state of https://github.com/mate-desktop/caja/pull/1730 is a bit weird, ihmo.

lukefromdc:
Should I close https://github.com/mate-desktop/caja/pull/1730 ?

lukefromdc:
Closed https://github.com/mate-desktop/caja/pull/1730 based on discussions elsewhere-and the likelihood that the windowing system agnostic gdk code may be better maintained going forward

Sunderland93:
After applying this patch and run Caja with `caja --force-desktop -n` wallpaper is turned black. When I tried to set another desktop wallpaper, it would appear for a couple of seconds and turn back to black. "Change desktop background" option in right-clicking context menu also doesn't work. Am I miss something?

raveit65:
Actual only the wayfire background modul is working. In wayfire.ini you need:
`background = wf-background`
My autostart section:
```
[autostart]
a1 = /usr/libexec/mate-settings-daemon
a6 = caja --force-desktop -n
a7 = nm-applet --indicator
autostart_wf_shell = false
background = wf-background
idle = swayidle before-sleep swaylock
panel = mate-panel
portal = /usr/libexec/xdg-desktop-portal
```
The background modul of mate-settings-daemon needs a complete re-work for wayland, and......and......a lot of ands..... :)

Sunderland93:
Ah, I see. Well, I use LabWC, so my option is swaybg :) Anyway, thank you! Everything else works well.

Sunderland93:
> [autostart]
a1 = /usr/libexec/mate-settings-daemon

On my setup mate-settings-daemon doesn't work at all, it crashes after startup with

```
invalid cast from 'GdkWaylandDisplay' to 'GdkX11Display'
Segmentation fault
```

raveit65:
Only s few modules of m-s-d do work. 3 or 4..... You can disable the plugins which do not work with dconf-editor.
First disable all and than enable each plugin and see if it crash or not.

Sunderland93:
Thanks, I've disabled modules that causes crash. Now it works.

lukefromdc:
Is this good to go for merging to master (which seems to get more 3ed party testing) or do you
all want to wait longer for anyone else on the team to review it?

PR #1725, opened 2023-06-29
Revert "caja-desktop-window: Fix the xrandr error"

raveit65:
Reverts mate-desktop/caja#1724

### DIFF
--- a/src/caja-desktop-window.c
+++ b/src/caja-desktop-window.c
@@ -155,9 +155,12 @@ caja_desktop_window_screen_size_changed (GdkScreen             *screen,
         CajaDesktopWindow *window)
 {
     int width_request, height_request;
+    int scale;
 
-    width_request = gdk_screen_get_width(screen);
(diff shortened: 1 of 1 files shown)


lukefromdc:
Did you find an issue with this beyond the deprecation? Reverting will bring back the situation where some users must restart caja after changing screen orientation, which is admittedly an unusual case.

I am OK with this either way, what does everyone else think?

zhuyaliang:
@raveit65 @lukefromdc I will handle this problem and find a better way to solve it

lukefromdc:
See
https://github.com/mate-desktop/caja/pull/1727
if you have something better feel free to submit an alternate PR and I will subject it to my full battery of tests

https://github.com/mate-desktop/caja/pull/1727 anchors the x11 desktop window to the root window, and passed all my tests: rotate, change resolution, change scaling, change number of monitors. That's still one machine, one GTK version, one underlyng OS however

PR #1729, opened 2023-06-30
ensure all of the desktop is usable in x11

lukefromdc:
Fix https://github.com/mate-desktop/caja/issues/1728 (area near bottom of screen unusable for icon placement)

*revert x11 icon position handling to the older and well tested code 
*use the newer code only in wayland where all the screen can already be used

Issue #1728, opened 2023-06-30
Desktop icons moved from the bottom edge after commit #5688856


L-U-T-i:
#### Expected behaviour
Desktop icons are aligned down by the desktop bottom edge

#### Actual behaviour
Desktop icons are about 50px up from the desktop bottom edge

#### Steps to reproduce the behaviour
Install caja build after git commit #5688856

#### MATE general version
1.27.0

#### Package version
Custom build

#### Linux Distribution
Rocky Linux 9.2

#### Link to bugreport of your Distribution (requirement)
Bug not reported as it is a custom build

If I apply the following patch (partially reverted commit #5688856):
```
diff -upr a/libcaja-private/caja-icon-container.c b/libcaja-private/caja-icon-container.c
--- a/libcaja-private/caja-icon-container.c	2023-05-25 20:56:56.000000000 +0200
+++ b/libcaja-private/caja-icon-container.c	2023-06-29 13:43:06.144776032 +0200
@@ -327,6 +327,7 @@ icon_set_position_full (CajaIcon *icon,
         GdkWindow *window;
         GdkMonitor *monitor;
         GdkRectangle workarea = {0};
+        int scale;
 
         /*  FIXME: This should be:
 
@@ -345,12 +343,13 @@ icon_set_position_full (CajaIcon *icon,
         window  = gtk_widget_get_window (GTK_WIDGET(container));
         monitor = gdk_display_get_monitor_at_window (gdk_display_get_default(), window);
         gdk_monitor_get_workarea(monitor, &workarea);
+        scale = gtk_widget_get_scale_factor (GTK_WIDGET (container));
         container_x = 0;
         container_y = 0;
-        container_width = workarea.width  - container_x
+        container_width = WidthOfScreen (gdk_x11_screen_get_xscreen (gdk_screen_get_default ())) / scale - container_x
                           - container->details->left_margin
                           - container->details->right_margin;
-        container_height = workarea.height - container_y
+        container_height = HeightOfScreen (gdk_x11_screen_get_xscreen (gdk_screen_get_default ())) / scale - container_y
                            - container->details->top_margin
                            - container->details->bottom_margin;
         pixels_per_unit = EEL_CANVAS (container)->pixels_per_unit;
```
[...]

PR #1730, opened 2023-07-02
always use same methods to compute desktop area

lukefromdc:
*Use the older gdk_x11 functions in x11
*Use the monitor dimensions in wayland for consistancy

### DIFF
--- a/eel/eel-gtk-extensions.c
+++ b/eel/eel-gtk-extensions.c
@@ -80,36 +80,49 @@ eel_gtk_window_get_geometry_string (GtkWindow *window)
 static void
 sanity_check_window_position (int *left, int *top)
 {
+    GdkScreen *screen;
+    gint scale;
+    GdkDisplay *display;
+
     g_assert (left != NULL);
     g_assert (top != NULL);
-    GdkDisplay *display;
-    GdkWindow *root_window;
-    GdkRectangle workarea = {0};
 
-    /* Make sure the top of the window is on screen, for
-     * draggability (might not be necessary with all window managers,
-     * but seems reasonable anyway). Make sure the top of the window
-     * isn't off the bottom of the screen, or so close to the bottom
-     * that it might be obscured by the panel.
-     *
-     */
-    root_window = gdk_screen_get_root_window (gdk_screen_get_default());
-    display = gdk_display_get_default ();
-    gdk_monitor_get_workarea(gdk_display_get_monitor_at_window (display, root_window), &workarea);
-    *top = CLAMP (*top, 0, workarea.height - MINIMUM_ON_SCREEN_HEIGHT);
+    screen = gdk_screen_get_default ();
+    display = gdk_screen_get_display (screen);
+    if  (GDK_IS_X11_DISPLAY (display))
+    {
+        scale = gdk_window_get_scale_factor (gdk_screen_get_root_window (screen));
 
-    /* FIXME bugzilla.eazel.com 669:
-     * If window has negative left coordinate, set_uposition sends it
-     * somewhere else entirely. Not sure what level contains this bug (XWindows?).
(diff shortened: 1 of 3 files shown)


lukefromdc:
Split out from https://github.com/mate-desktop/caja/pull/1729 which uses the same functions to set the size of the usable icon area on the desktop

lukefromdc:
Strangely, I got a build failure on translations with this on the laptop unless --disable-nls was passed(Ithen the build finished as expected), maybe something specific to that machine. Debian travis build finished, Federa errored out in make distcheck

raveit65:
Distcheck did create the tarball, so it is fine.
It's a problem with deployment of gh-pages.
https://app.travis-ci.com/github/mate-desktop/caja/jobs/605290626#L5040
`Failed to push the build to github.com/mate-desktop/caja.git:gh-pages`
I guess it's a connection problem between travis-CI and github servers.
Starting the build again.

raveit65:
CI result is good now.

lukefromdc:
Anyone else want to review this?

raveit65:
As far i know we using `scale` only because of using Width/HeightOfScreen which is a xlib macros and can't be used with wayland.
Maybe it is possible to use GDK equivalents because GDK knows about the scaling factor?
At one place you're replacing `screen_width  = gdk_screen_get_width  (screen)` with `screen_width  = WidthOfScreen (gdk_x11_screen_get_xscreen (screen)) / scale;`
And you added the scale variable.
I am thinking i might go the other way around and trying to use the same code for X11 and wayland for getting the screen dimensions.
But maybe this isn't possible.

lukefromdc:
I can run a test build using the wayland code in all places, I will lhave to examine it closely for issues with navigation windows called with --geometry oversize and with the desktop edges

L-U-T-i:
@raveit65 
> I am thinking it might go the other way around and trying to use the same code for X11 and wayland for getting the screen dimensions. But maybe this isn't possible.

I've indicated in [#1728](https://github.com/mate-desktop/caja/issues/1728#issuecomment-1616552801) that it seems it could be enough just to not deduct margins (container->details->top_margin and container->details->bottom_margin in my case) from container_height for X11 (if it has to be done for wayland at all - or, it was just the remaining from the old code, and is already done elsewhere for the new code...), keeping all of the new (wayland) code. With just that patch (and without both #1729 and #1730) everything seems to work just fine for me in X11.

Maybe this way it would be much more simple to test and to maintain the code in the future?

lukefromdc:
If we can use the same codepath in x11 and wayland, it certainly simplifies testing

eel/eel-gtk-extensions.c
```diff
@@ -169,8 +184,23 @@ eel_gtk_window_set_initial_geometry (GtkWindow *window,
...
+        }
+        else
+        {
+            GdkRectangle geometry = {0};
+            GdkMonitor *monitor;
+            monitor = gdk_display_get_monitor (display, 0);
+            gdk_monitor_get_geometry(monitor, &geometry);
+            screen_width = geometry.width;
+            screen_height = geometry.height;
+        }
```
raveit65, line 203: Before this change we have 2 lines of (well working) code.
I do not get why this should blow up, only because of an idea to use at all places the same old xlib macros.
Do we have crashes in x11 at this place?
Does anyone have a stacktrace which point to this 2 lines?
I recall in 2019 when all this xlib macros were added by a deprecation fix PR, that we get warned by a gnome dev not to do this, because it is only usable for x11.
I won't review it because i am thinking the idea of this PR is wrong.
Anyway, i don't block it when other devs from @mate-desktop/core-team  think this is an excellent idea.

lukefromdc:
OK, I will close this

lukefromdc:
Going forward, the general case gdk code is likely to be better maintained

Issue #1740, opened 2023-08-23
Travis-CI Debian build is broken with master


raveit65:
#### Actual behaviour

Travis-CI Debian:testing build is broken with master
https://app.travis-ci.com/github/mate-desktop/caja/jobs/608465207#L26915
```
====================================

   caja 1.27.1: src/test-suite.log

=====================================

# TOTAL: 1

# PASS:  0

# SKIP:  0

# XFAIL: 0

# FAIL:  1

# XPASS: 0

# ERROR: 0

.. contents:: :depth: 2

FAIL: check-caja

================

running caja_self_check_file_utilities

running caja_self_check_file_operations

running caja_self_check_directory

(caja:43086): GLib-GIO-CRITICAL **: 10:35:13.829: GFileInfo created without standard::sort-order

Trace/breakpoint trap (core dumped)

FAIL check-caja (exit status: 133)
```
#### MATE general version

master
#### Package version

> 1.27
#### Linux Distribution
Debian testing

#### Link to bugreport of your Distribution (requirement)
not releasesd

Note: fedora build works fine

lukefromdc:
That set of warnings came in with glib 2.77.1 and may date back to glib2.77.0. I was assuming it was a glib issue but neither Nautilus nor Nemo throw these warnings so there should be one or more commits in their history relevent to this

lukefromdc:
See
https://github.com/mate-desktop/caja/pull/1743
the warnings do not appear in Nemo or Nautilus, fixing this was literally a 3 line fix across two files

raveit65:
https://github.com/mate-desktop/caja/pull/1743 is merged.

PR #1743, opened 2023-08-24
fix gfile sort and symlink warnings w glib 2.76 or later

lukefromdc:
*Use code borrowed from Nemo

### DIFF
--- a/libcaja-private/caja-directory-async.c
+++ b/libcaja-private/caja-directory-async.c
@@ -3807,7 +3807,7 @@ static gboolean is_trusted_system_desktop_file (GFile *file)
         return FALSE;
     }
 
-    target = g_file_info_get_symlink_target (info);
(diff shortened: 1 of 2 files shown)


lukefromdc:
Fix
https://github.com/mate-desktop/caja/issues/1740
`
caja --check` finishes clean with this applied over glib2.76 or later (Debian Sid is shipping 2.77.1 at this moment)

raveit65 (approved):
LGTM,
debian build in travis-ci is working again.

lukefromdc:
We may want to redo the caja release (to 1.27.2)  to include this due to all the log spam the gfile issue creates

raveit65:
Sure i can do, but in which (runtime) logs do caja spam without this commit?

raveit65:
Ok, got it, 1.27.1 can't be build on debian:testings systems.

lukefromdc:
A normal build of 1.27.1 on my end w my default build options will finish and will run, but when run from terminal you will see more than a full terminal screen's worth of warnings. After this commit those are gone

PR #1750, opened 2023-11-22
fix building with libxml 2.12.0

raveit65:
```
caja-application.c:1557:11: error: implicit declaration of function ‘xmlReadMemory’; did you mean ‘xmlInitMemory’? [-Werror=implicit-function-declaration]
1557 |     doc = xmlReadMemory (data, strlen (data), NULL, "UTF-8", 0);
|           ^~~~~~~~~~~~~
|           xmlInitMemory
caja-application.c:1557:9: warning: assignment to ‘xmlDocPtr’ {aka ‘struct _xmlDoc *’} from ‘int’ makes pointerfrom integer without a cast [-Wint-conversion]
1557 |     doc = xmlReadMemory (data, strlen (data), NULL, "UTF-8", 0);
|         ^
```
Building caja with commit from PR for fedora-rawhide at koji buildserver works fine.
https://src.fedoraproject.org/rpms/caja/blob/rawhide/f/caja_0001-fix-building-with-libxml-2.12.0.patch
https://koji.fedoraproject.org/koji/taskinfo?taskID=109410799

### DIFF
--- a/src/caja-application.c
+++ b/src/caja-application.c
@@ -32,6 +32,7 @@
 #include <unistd.h>
 #include <fcntl.h>
 
+#include <libxml/parser.h>
 #include <libxml/xmlsave.h>
 #include <glib/gstdio.h>
 #include <glib/gi18n.h>


lukefromdc:
Did xmlReadMemory get moved to a different file in libxml for 2.12, or have we been incorrectly reading a reference to it in another file for years? The former would be a breaking change calling for  a new major version, latter woud be on us.

lukefromdc (approved):
Again, no problems building with libxml 2.9 with this, though we need a test with 2.12.0

cwendling:
> Did xmlReadMemory get moved to a different file in libxml for 2.12, or have we been incorrectly reading a reference to it in another file for years? The former would be a breaking change calling for a new major version, latter woud be on us.

I didn't move, but [they changed which headers include which others](https://discourse.gnome.org/t/libxml2-2-12-0-released/18063/1), meaning some headers used to bring others and they stopped.  Agreed that it would have been worth a major release because it seems to basically break everybody, but well.  OTOH we kind of were not including the header we should have, but the fact that it worked kind of made it part of the API.

cwendling (approved):
LGTM

raveit65:
>> Again, no problems building with libxml 2.9 with this, though we need a test with 2.12.0

Fedora-rawhide use libxml-2.12.0. For this reason i posted the link with a sucessfull build at fedora build-server.
https://src.fedoraproject.org/rpms/caja/blob/rawhide/f/caja_0001-fix-building-with-libxml-2.12.0.patch
https://koji.fedoraproject.org/koji/taskinfo?taskID=109410799

Looks like i have to write a wiki entry. How to release a package release with travis-ci.

PR #1754, opened 2023-12-09
wayland desktop: suppress spatial window warning

lukefromdc:
*When under wayland, silence `caja_spatial_window_show: assertion 'location != NULL' failed` warning
*The startup sequence is not the same with wlroots and the location isn't ready yet 
*Using gtk_widget_show instead of gtk_widget_realize in caja-desktop-window.c generates this warning


raveit65 (approved):
LGTM,
didn't found the time to test it under wayland.

PR #1755, opened 2023-12-10
Wayland desktop: do not attempt to take over background image/color handling from the compositor

lukefromdc:
Do not attempt to draw the background in wayland, as it starts but reverts to the compositor's background image in about a second. This seems to be the compositor taking over. We have no root window to draw the background on in wayland. In wayland, the compositor normally manages the desktop background, drawing it on the screen without even defining it as a window.

Also, connect the right-click menu item "Change Desktop Background" to a useful error dialog telling users to change the background in the compositor's background manager.

No change in desktop icon rendering, this affects the background only.

### DIFF
--- a/eel/eel-background.c
+++ b/eel/eel-background.c
@@ -449,6 +449,15 @@ init_fade (EelBackground *self)
         return;
     }
 
+    GdkDisplay *display = gdk_screen_get_display (gdk_screen_get_default());
+
+    /*do not attempt to manage desktop background on wayland
+     *as the compositor handles this
+     */
+
+    if (!(GDK_IS_X11_DISPLAY (display)))
+        return;
+
     mate_background_preferences = g_settings_new ("org.mate.background");
     do_fade = g_settings_get_boolean (mate_background_preferences,
                                       MATE_BG_KEY_BACKGROUND_FADE);
--- a/src/file-manager/fm-desktop-icon-view.c
+++ b/src/file-manager/fm-desktop-icon-view.c
@@ -42,6 +42,7 @@
 #include <eel/eel-glib-extensions.h>
 #include <eel/eel-gtk-extensions.h>
 #include <eel/eel-vfs-extensions.h>
+#include <eel/eel-stock-dialogs.h>
 
 #include <libcaja-private/caja-desktop-icon-file.h>
 #include <libcaja-private/caja-directory-background.h>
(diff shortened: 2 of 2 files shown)


lukefromdc:
Even if we can get caja to take over and hold the background(posssible if Caja's background was crashing, not possible if the compositor was drawing over or redirecting it somewhere), the compositor's background image would then show when caja starts or restarts and would be different unless set by the user both in caja and in the compositor to the same image or color. Thus, yield to the compositor's management of the background when in wayland. This gives a consistant background.

Previously, Caja and the compositor's background manager (.e.g wf-background) were set to different backgrounds, Caja's would show for about a second, then the compositor's image would show.

Sunderland93:
> Also, connect the right-click menu item "Change Desktop Background" to a useful error dialog telling users to change the background in the compositor's background manager.

I would still prefer native support for changing wallpapers in Caja, since it would not break the workflow familiar to MATE X11 users (setting a background image from Caja or Eye of MATE instead of a separate wf-background or swaybg setting)

lukefromdc:
I will experiment to see if this is possible. With current git master it comes up for
about a second before either crashing (which we could fix) or more likely being
overrriden by the compositor (more difficult to fix).

Some way of allowing Caja to call WCM->shell->Background->General if wayfire
is detected and similar config managers for other common compositors would also
work.

lukefromdc:
Also note that clicking the background and selecting "Change Desktop Background" works by callijng mate-settings-daemon->appearance.
Few or one of the m-s-d plugins as they currently stand can run under wayland, mate-settings-daemon is an xsettings manager so we'd
need an alternate way to do this. Possibly a bare-bones background manager, advantage being this is compositor-independent but might
be more than I can code up.

What we'd need would be a way to run just the background portion of m-s-d->appearance in standalone mode so it could write to dconf.
Caja would then read the background from dconf just as it does in x11. 

A problem with trying to get the rest of m-s-d running in wayland is this: much of the work an xsetting manager does in x11 is included
in the compositor in wayland (yes, with all the fragmentation that comes with that). Still worse, wayfire and probably many other compositors
read the GNOME not the MATE settings from dconf, presumably as these are the schemas most likely to already have been installed.

I could submit a feature request to the wayfire devs for the ability to select which set of gsettings prefs to read though.

lukefromdc:
Closing as I have found a way to easily render the background using GTK and css, and a way to get the background filename into that CSS. WIP right now as for the moment I've only got images supported, caja in wayland isn't listening for a change of background yet (but will apply it on a restart) and I need to add to caja running in wayland the dialog to change the background.

Turns out we can even fade one background into another in GTK with css, though that's down the road

lukefromdc:
Now working on background management at
https://github.com/mate-desktop/caja/tree/wayland_desktop_background
So far can show the background, and can change it from dconf-editor or "set as wallpaper" but must then use the right-click "Change Desktop Background" menu item to apply it. Next up is a dialog there sufficient at least to take filenames, and hopefully I can get it to listen for the gsettings change by also invoking the background read/apply function from a callback

lukefromdc:
I now have this to the point that the desktop background can be changed immediately from dconf-editor or the wallpaper ("set as wallpaper") caja extension as well as from a GtkFileChooserButton in the Change Desktop Background dialog.

Code was loosely based on an old fallback routine for desktop handling in the noncomposited case before we got the gtk 3.21 desktop redraw issues fixed right during the 1.16 cycle. I had to move the actual drawing back to the toplevel desktop window as that's the one GtkLayerShell is supposed to keep on the bottom. Note that on wayland we cannot draw on the root window because it does not exist.

Issue #1623, opened 2022-04-03
Data transfer estimates and bandwidths wrong when queuing transfers


mxmlnkn:
#### Expected behaviour

Data transfer estimates and bandwidths should at least somewhat mirror what the user can deduce from the transferred size progress (x GB of y GB).

#### Actual behaviour

When queueing multiple transfers, some transfers will only run after an hour or so because they waited for other transfers. However, the bandwidth calculation and time estimate seem to include the hour of wait time in the average bandwidth calculation leading to vastly underestimated bandwidths of e.g. 1 MB/s instead of 200 MB/s.

#### Steps to reproduce the behaviour

1. Copy large files to some location
2. Copy another large file to some location
3. The second copy should be queued in the Caja "File Operations" dialog
4. Wait for the first transfer to finish and the second to begin

#### MATE general version

I'm using Caja on XFCE 4.16.1

#### Package version

1.26.0-0ubuntu1

#### Linux Distribution

Ubuntu 21.10

PR #1758, opened 2023-12-31
caja-file-operations: fixes the operations timer handling

### LINKED ISSUE #1420: File Operations dialog incorrectly displays throughput speed when files are queued
#### Expected behaviour

display real copy speed in octet/s  since the beginning of the copy of the current file proceeded. 

#### Actual behaviour

display an average speed deducted from time since the beginning of the copy of the first file in the pool, instead of the beginning of copy of the current file been proceeded. 

copy a file to destination, then add another file to copy to the same destination so the operations is queued, then notice the difference in displayed throughput in the file operations window when the second file is copied.
 
#### MATE general version
1.24.0

#### Package version
caja           1.24.0-1     amd64

#### Linux Distribution
ubuntu 20.04 LTS

basicmaster:
Essentially fixes the bug that for queued file operations the wrong transfer rate is shown.

To reproduce the issue/check the fix, start a (larger) file copy/move operation and then start a second one, while the first one is still running.

Fixes #1420 and #1623.

### DIFF
--- a/libcaja-private/caja-file-operations.c
+++ b/libcaja-private/caja-file-operations.c
@@ -5232,6 +5232,8 @@ move_job (GIOSchedulerJob *io_job,
 		goto aborted;
 	}
 
+	g_timer_start (job->common.time);
+
 	memset (&transfer_info, 0, sizeof (transfer_info));
 	move_files (job,
 		    fallbacks,
--- a/libcaja-private/caja-file-operations.c
+++ b/libcaja-private/caja-file-operations.c
@@ -1873,7 +1873,7 @@ trash_files (CommonJob *job, GList *files, guint *files_skipped)
 	for (l = files;
 	     l != NULL && !job_aborted (job);
 	     l = l->next) {
-        caja_progress_info_get_ready (job->progress);
+        caja_progress_info_get_ready (job->progress, job->time);
 
 		file = l->data;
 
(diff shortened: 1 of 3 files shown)


lukefromdc (approved):
With this applied, what I found most noticable was that queuing other files for copying now seems to give accurate bandwidth indication, and the waiting file shows "0 seconds" and zero bandwidth as it is not being processed yet. This is a lot better than what we used to get. Ideally the "0 seconds" would be hidden but users will probably know that's a value that can be ignored. Approving and calling for a 2nd review

basicmaster:
The shown estimate for a queued operation indeed doesn't make sense. The reason is that actually a non-zero transfer rate is [one condition to *not* show the estimate](https://github.com/mate-desktop/caja/blob/8bb93666b35ee6277071d1b5ab4d52ae57d06975/libcaja-private/caja-file-operations.c#L3083). I fixed this in #1759, aligning to the delete operation case.

lukefromdc:
Any updates on this or the alternate? Can this one be closed since we have https://github.com/mate-desktop/caja/pull/1759 ?

basicmaster:
This and #1759 both affect file operations. However they are not alternative solutions, but address different aspects. So they are needed both.

From my side, this MR is ready to be merged (however there still has been no second review).

basicmaster:
Is there maybe an update on this?

lukefromdc:
I will pull this, squash, rebase and then merge to avoid any git surprises

lukefromdc:
Merged with
https://github.com/mate-desktop/caja/commit/ed41589e35a534cb47ea7b6dcc685fef2e7ca578

lukefromdc:
Will cherrypick to 1.28 if no suprises down the road from this

PR #1759, opened 2024-01-02
caja-file-operations: fix estimate for queued copy

basicmaster:
Fixes the condition for showing an estimate of the remaining duration in case a copy operation is queued, aligning to the delete operation case.

### DIFF
--- a/libcaja-private/caja-file-operations.c
+++ b/libcaja-private/caja-file-operations.c
@@ -3079,8 +3079,8 @@ report_copy_progress (CopyMoveJob *copy_job,
 		transfer_rate = transfer_info->num_bytes / elapsed;
 	}
 
-	if (elapsed < SECONDS_NEEDED_FOR_RELIABLE_TRANSFER_RATE &&
-	    transfer_rate > 0) {
+	if (elapsed < SECONDS_NEEDED_FOR_RELIABLE_TRANSFER_RATE ||
+	    transfer_rate <= 0) {
 		char *s;
 		/* Translators: %S will expand to a size like "2 bytes" or "3 MB", so something like "4 kb of 4 MB" */
 		s = f (_("%S of %S"), transfer_info->num_bytes, total_size);
--- a/libcaja-private/caja-file-operations.c
+++ b/libcaja-private/caja-file-operations.c
@@ -1495,7 +1495,7 @@ report_delete_progress (CommonJob *job,
 			TransferInfo *transfer_info)
 {
 	int files_left;
-	double elapsed;
+	double elapsed, transfer_rate;
 	gint64 now;
 	char *files_left_s;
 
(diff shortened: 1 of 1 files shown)


lukefromdc (commented):
I applied this by hand on top of the other PR. Built fine and no crash, but a long copy job once started tended to continue reporting zero bytes transfered until done or nearly so

libcaja-private/caja-file-operations.c
```diff
@@ -3074,21 +3074,17 @@ report_copy_progress (CopyMoveJob *copy_job,
...
 		/* Translators: %S will expand to a size like "2 bytes" or "3 MB", so something like "4 kb of 4 MB" */
 		s = f (_("%S of %S"), transfer_info->num_bytes, total_size);
 		caja_progress_info_take_details (job->progress, s);
 	} else {
+		double transfer_rate;
 		int remaining_time;
 		char *s;
 
+		transfer_rate = transfer_info->num_bytes / elapsed;
 		remaining_time = (total_size - transfer_info->num_bytes) / transfer_rate;
```
cwendling, line 3088: This is dangerous, yet nothing really new to this code. if either `elapsed` or `transfer_rate` is 0, it'll crash.
Admittedly, `elapsed` will not be 0 *if `SECONDS_NEEDED_FOR_RELIABLE_TRANSFER_RATE` is greater than 0*, which is gonna be the case but is not necessarily a good thing to depend on.

`transfer_rate` not being 0 puzzles me a tad more, because it means we have to have transferred *something* before this can be safely called.

What about simply changing the condition from `&& transfer_rate > 0` to `|| transfer_rate <= 0`? This fixes the zero-division, and might even have been the intention.

lukefromdc: Somehow I didn't see that. I didn't get a crash of all of caja but DID get a general failure of the progress indicator to function. In general, any input that could be zero cannot be divided by without checking for and avoiding the zero case.

basicmaster: Indeed `SECONDS_NEEDED_FOR_RELIABLE_TRANSFER_RATE` could be changed to zero at a later time, so this shouldn't be assumed to be non-zero. Being a fan of minimally invasive fixes, I reverted the original fix (feel free to discard the original fix and the revert when merging) and instead just changed the condition, as proposed.

As the possible division-by-0 also affects the delete case, I added a fix for this in a separate commit.

faulesocke:
Fixing this properly should be easier: Just reset `elapsed` to zero when the job actually starts. Also reset it when it is manually unpaused by the user (which might be the same code path, haven't looked into it).

libcaja-private/caja-file-operations.c
raveit65: Ok, this definetly needs to be squashed during merge.
To much revert and fixes of fixes commits.
@basicmaster 
with using git rebase -i you can edit/delete, sort, etc. very easy
Still wondering why coder don't know this strong git command....

lukefromdc: Could just be new to it: I had to learn these things one at a time myself.

basicmaster: I'm aware of interactive rebases but didn't want to do a forced push here, as I didn't know whether/how GitHub handles this in a MR. Thus I pushed the revert commit and the improved solution commits, and asked you to discard them later.
As GitHub apparently copes with a force push, I just discarded the initial solution (and its revert) and rebased the branch on current `master`.

basicmaster:
@faulesocke:
You actually refer to the changes in #1758, aren't you?

The case to pause a job after it started (e.g. when copying multiple files) has also to be considered here. In that case you definitely need to stop/continue the timer, otherwise you would again start at zero time and get a wrong estimate.

libcaja-private/caja-file-operations.c
basicmaster: I was wondering if there is anything else needed to resolve (and in the end to merge) this. Have you been able to check the rebased solution in the meantime?

lukefromdc:
Any updates on this?

basicmaster:
From my side, this MR is ready to be merged. As far as I can see, I answered/addressed all feedback, but I'm of course happy to do further adjustments, if needed.

libcaja-private/caja-file-operations.c
basicmaster: @cwendling Can this conversation be resolved?

basicmaster:
Is there maybe an update on this?

lukefromdc:
This goes into git master which is 1.29, if there are no nasties down the road it can be cherrypicked to 1.28

lukefromdc:
Merged in 
https://github.com/mate-desktop/caja/commit/bfa6b650c732ff9edddd1537b14bdab39819fc45

PR #1766, opened 2024-02-03
caja-file-operations: fix caption/button in destination error dialogue

basicmaster:
For certain operations, the destination is checked. If an error occurs, the user is asked how to proceed using a dialogue. So far this dialogue was hard-coded to only consider the copy operation.

This fix also considers the other affected operations and generalizes the continue button text.

### DIFF
--- a/libcaja-private/caja-file-operations.c
+++ b/libcaja-private/caja-file-operations.c
@@ -152,6 +152,8 @@ typedef struct {
 } SetPermissionsJob;
 
 typedef enum {
+	OP_KIND_CREATE,
+	OP_KIND_LINK,
 	OP_KIND_COPY,
 	OP_KIND_MOVE,
 	OP_KIND_DELETE,
@@ -188,7 +190,7 @@ typedef struct {
 #define DELETE_ALL _("Delete _All")
 #define REPLACE_ALL _("Replace _All")
 #define MERGE_ALL _("Merge _All")
-#define COPY_FORCE _("Copy _Anyway")
+#define CONTINUE_FORCE _("Continue _Anyway")
 
 NotifyNotification *unmount_notify;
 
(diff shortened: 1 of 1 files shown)


basicmaster:
The mentioned dialogue for a move operation:

Before ("kopieren" translates to "to copy"):
grafik

After:
grafik

lukefromdc (commented):
The text change in the button works, though translations would presumably have to be re-synced before the middle button would be translated.

I don't see any easy way to test (or really an application for) this in the "move" case as such errors as out of space cannot exist in that case. NOT sure how copy-on-write filesystems handle this though. I am using EXT4 for basically everything. When permissions are the issue,a force option does not exist. Creating a new file created a 0 byte file, and would fail only with an absolutely full filesystem that cannot handle the metadata. Continuing in such a case would surely corrupt the entire fs.

The two button error dialog for trying to copy into a root-owned directory as user is unchanged, create, make link, etc are greyed out when permissions is the issue and that seems unchanged, as it should be

basicmaster:
A real-world use case is e.g. to copy (or move) a larger file to a different file system (e.g. to a USB drive), which doesn't have enough space left to store the file.

My screenshots show the attempt to move a file to a RAM drive (for which caja always shows 0 bytes free space). Everytime I want to copy or move a file to it, I get the shown error message and have to click on the center button to (successfully) copy/move a file to it.
The following line in `/etc/fstab` is used to create/mount the RAM drive:

    ramfs   /ramdisk     ramfs   defaults,mode=777        0       0

raveit65:
I could reproduce the issue with creating a ramdisk on the fly https://www.linuxbabe.com/command-line/create-ramdisk-linux , and i can confirm that PR fixes the wrong text message.

raveit65 (approved):
LGTM

lukefromdc (approved):
From what I understand a file cannot be moved to a different filesystem, only copied there and deleted from the source directory. If caja supports doing that as a "move" operation than yes, this is needed. It works so approving. The dialog worked fine when I tested attempting to copy a big video file into a bootable flash drive's boot partition

basicmaster:
> From what I understand a file cannot be moved to a different filesystem, only copied there and deleted from the source directory. If caja supports doing that as a "move" operation than yes, this is needed.

This is indeed what caja is doing here when moving a file. Usually [GLib's `g_file_move`](https://docs.gtk.org/gio/method.File.move.html) (called by caja) already takes care of this.

PR #1771, opened 2024-03-25
wayland background: use mate-appearance-properties if we can

lukefromdc:
*In the wayland session, we now have mate-settings-daemon with GDK_BACKEND=x11 
*We need it to control theming in xwayland apps
*So we can now use mate-appearance-properties to set the desktop background when it is running

*To support standalone use of caja in any wayland session  when mate-settings-daemon is not running or when mate-control-center is not installed, we use the standalone wayland dialog in those cases

### DIFF
--- a/src/file-manager/fm-desktop-icon-view.c
+++ b/src/file-manager/fm-desktop-icon-view.c
@@ -717,7 +717,16 @@ action_change_background_callback (GtkAction *action,
     /*Get the new background and switch to it in wayland*/
     if (GDK_IS_WAYLAND_DISPLAY (gdk_display_get_default()))
     {
-        wayland_bg_dialog_new ();
+        /*We can use the appearance capplet with some versions of mate-control-center
+         *in which the appearance capplet works in wayland
+         *Try it first, and fall back to the standalone dialog if it fails
+         */
+        GError *error = NULL;
+
+        g_spawn_command_line_async ("mate-appearance-properties --show-page=background",
+                                          &error);
+        if (error != NULL)
+            wayland_bg_dialog_new ();
     }
     else
 #endif


lukefromdc:
No need to update anything else to use it so hoping this will be OK for 1.28 as wayland is considered experimental, nothing changes for the production x11 session, and it doesn't change any dependencies due to the fallbacks. 

If not, bumping version to 1.29 for wayland development is also fine with me.

lukefromdc:
Latest force-push loses the system() calls, uses g_spawn_command_line_async() instead, and uses the GError variable to determine directly whether we have a version of mate-appearance-properties that can run in wayland directly or not.

Takes advantage of
https://github.com/mate-desktop/mate-control-center/pull/749
but does not depend on it because if mate-appearance-properties fails, we just fall back to the standalone wayland dialog

lukefromdc:
Bumped version in master to 1.29 so we can use this whether or not we want to include it in the next 1.28 release. In this case, nothing breaks if used with an older version of mate-control-center (predating https://github.com/mate-desktop/mate-control-center/pull/749) but the next PR opened by anyone could have a hard dependency on another, so I figured time to bump version in master. Many projects do this as soon as they release a major version

lukefromdc:
Anyone want to have a look at this one?

lukefromdc:
Still waiting for a review on this, it's basically three lines of code plus comments to allow using mate-appearance-properties for background setting on wayland if we have the 1.29/git master version installed. If not, we fall back to the standalone background setting dialog.

Purpose is to get the same UI for setting the desktop background in wayland as in x11

lukefromdc:
https://github.com/ilya-fedin can you test this?

mark-herbert42:
> @ilya-fedin @mark-herbert42 can you test this?

Tried on my laptop before and after patch - works perfectly. Settings daemon is running in the background in X11 mode.

lukefromdc:
Merging as we have a successful test by someone else in wayland and the x11 session is untouched.

PR #1787, opened 2024-08-14
wayland: ensure windows can be moved and resized  if compositor is using CSD

lukefromdc:
Ensure that nagivation and spatial windows can be moved and resized when running under wayland with CSD enabled, note that CSD is the default in many compositors and forced in some. 

Also note that in wayfire, using CSD will pick up the GTK theme's decorations which will look like they do for windows that do not use headerbars (such as all MATE apps) in Mutter. This gives far better support for the MATE themes than firedecor or similar SSD decorators for wayfire. We read the port of the metacity theme to GTK and without a GtkHeaderBar we get normal titlebar theming and dimensions unless the theme itself has an issue.

### DIFF
--- a/src/caja-window.c
+++ b/src/caja-window.c
@@ -632,17 +632,6 @@ caja_window_set_initial_window_geometry (CajaWindow *window)
 }
 
 static void
-caja_window_constructed (GObject *self)
-{
-    CajaWindow *window;
-
-    window = CAJA_WINDOW (self);
-
-    caja_window_initialize_bookmarks_menu (window);
-    caja_window_set_initial_window_geometry (window);
-}
-
-static void
 caja_window_set_property (GObject *object,
                           guint arg_id,
                           const GValue *value,
@@ -742,6 +731,12 @@ caja_window_constructor (GType type,
 
     slot = caja_window_open_slot (window->details->active_pane, 0);
     caja_window_set_active_slot (window, slot);
+    /*We can now do this here instead of in a separate constructed function
+     *and we need to because the separate constructed function causes the
+     *window to be un-draggable/un-resizable with the mouse in wayland
+     */
+    caja_window_initialize_bookmarks_menu (window);
+    caja_window_set_initial_window_geometry (window);
 
     return object;
 }
@@ -2170,7 +2165,6 @@ caja_window_class_init (CajaWindowClass *class)
     GtkBindingSet *binding_set;
 
     G_OBJECT_CLASS (class)->constructor = caja_window_constructor;
-    G_OBJECT_CLASS (class)->constructed = caja_window_constructed;
     G_OBJECT_CLASS (class)->get_property = caja_window_get_property;
     G_OBJECT_CLASS (class)->set_property = caja_window_set_property;
     G_OBJECT_CLASS (class)->finalize = caja_window_finalize;


lukefromdc:
There is a "metacity decorator" for wayfire but at least Debian doesn't ship it and it's an early alpha right now.

src/caja-window.c
```diff
@@ -2170,7 +2185,18 @@ caja_window_class_init (CajaWindowClass *class)
     GtkBindingSet *binding_set;
 
     G_OBJECT_CLASS (class)->constructor = caja_window_constructor;
-    G_OBJECT_CLASS (class)->constructed = caja_window_constructed;
+#ifdef HAVE_WAYLAND
+    /*For some reason if we use the constructed() class method
+     *in wayland with the compositor set to or forcing CSD
+     *we lose the ability to drag the window by the titlebar
+     */
```
cwendling, line 2192: I would guess then that some of the things done in `constructed()` are done too late for some compositors.  If that's the case, can't we simply do it earlier for everyone, instead of having that weird setup?

src/caja-window.c
```diff
@@ -2170,7 +2185,18 @@ caja_window_class_init (CajaWindowClass *class)
...
+    GdkDisplay *display = gdk_display_get_default();
+    if (GDK_IS_X11_DISPLAY (display))
```
cwendling, line 2194: This should be the exact opposite test as in constructor(), so `! GDK_IS_WAYLAND_DISPLAY (display)`.  OK, it's likely not gonna matter anytime soon, but it's conceptually more correct, and could cause headaches if at some point there's more that just X11 vs. Wayland (does anybody remember the MIR backend? :smile: )

lukefromdc:
I will test moving those funcrions unconditionally. The use of a "constructed" block is according to the documentatiion meant for cases where something must br delayed, so my guess is that a long time ago this was needed. The libunique to GtkAppication port years ago used code from Nautilus 3.0, so my guess is in that era there was an issue.

The reason I split the move of this out in a non-x11 only block was to minimize changes in the X11 session given limites testers. Simpler code and one less runtime check if we can do this unconditionally of course. 

I will be able to test that later today, if no surprises in x11 I will update this PR accordingly. If that acts up I can chexk for wayland explicitly instead.

src/caja-window.c
lukefromdc: That is exactly what the latest force-push does: removes the constructed() class method entirely and moves setting initial geometry and bookmarks to the constructor() class

src/caja-window.c
lukefromdc: This is no longer a factor as we now move this code to the constructor() function unconditionally

lukefromdc:
Latest force-push unconditionally removes the constructed() class and moves setting up bookmarks and initial geometry to the constructor() class unconditionally. No change in wayland behavior from the earlier version of this. A test in x11 also showed the bookmarks loading as before, and initial window geometry whether stored from last run or specified on the command line with the `-g` flag working the same as before. It thus appears we no longer need the ` constructed() `class at all. 

Removed it entirely: this simplifies the code, gives us one less include, and one less runtime check to deal with

lukefromdc:
The Travis error is
https://github.com/mate-desktop/caja/issues/1790
which is another gem error on deployment that seems to be MATE-wide. As it is the deploy code affects only the Federa build. Debian build without the deploy code completed fine.

PR #1789, opened 2024-08-29
build: remove configure dependency on perl

eli-schwartz:
It is totally unused. In 2010, nautilus removed a perl script used to generate enums:
https://gitlab.gnome.org/GNOME/nautilus/-/commit/75a03a440ebff23ccbf8674ca4a0e9f6475ebbb8

Before then, perl was needed to build. After this commit, nautilus still had messy code to check for perl availability but did nothing with it. Across multiple forks and reinitialized git repositories that dropped all commit history, it turns out that caja still has that messy code too.


lukefromdc (approved):
No problems on this end. Build went fine, runs same as before, and only references to "perl" I could find with a pluma string search were in the work "properly" so definatly unused.

Very simple, will merge this. Thanks!

eli-schwartz:
No problem. :) I noticed it because somehow Gentoo triggered a rebuild of this package after the perl 5.40 migration and I was trying to figure out what actually used it. Lo and behold.... :)

PR #1805, opened 2025-03-11
wayland/window closing: avoid a crash on gtk_widget_destroy

lukefromdc:
* Do not attempt to unref or destroy a child of a container

*In GTK 3 at least, GtkContainers automatically destroy child widgets when destroyed

*Trying to do this with glib 2.82 or later causes sporadic crashes on closing a navigation window at least in wayland.

### DIFF
--- a/libcaja-private/caja-icon-container.c
+++ b/libcaja-private/caja-icon-container.c
@@ -4416,7 +4416,10 @@ destroy (GtkWidget *object)
     /* destroy interactive search dialog */
     if (container->details->search_window)
     {
-        gtk_widget_destroy (container->details->search_window);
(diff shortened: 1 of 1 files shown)


lukefromdc:
I have been chasing this crash for months and finally found it. Not a single crash since I installed caja with this change, and I looked for any memory leak behavior by repeatedly opening and closing windows while monitoring RAM consumption in mate-system-monitor.

I'd like to have another set of eyes on this before merge if possible

lukefromdc:
Merging as I've had no trouble with this and newer glib versions require it at least on wayland and possibly on x11 as well

PR #1810, opened 2025-04-03
Various small accessibility improvements adding missing or insufficient labels

cwendling:
This replaces #1568.  It gives mostly the same results, but in what I believe a better way:
* [Properties window: add accessible description to “Execute” check box](https://github.com/mate-desktop/caja/pull/1568/commits/e830ac40e711b8ea9bc396c49666fc540f83f65d)]: this seems to be a hack for an Orca bug that I could not reproduce, so nothing similar is added here
* [Properties window: add accessible label to icon button](https://github.com/mate-desktop/caja/pull/1568/commits/2042ef4b03daaccdff71e8a647630afcf6c59a32) and [Properties window: add accessible name & description to icon button](https://github.com/mate-desktop/caja/pull/1568/commits/128d6dcc1420411819ed7c9240d665855386329d) are implemented differently: 4abe18dbd3aeb1f609fde72a1f534bc5a829f6f2 for the icon/button, and a8e36f3bfe154237b00853e2804a50afe5531f30 for the emblems (which also fixes the property browser, see below)
* [Properties window: ensure that the “Others” label can be read](https://github.com/mate-desktop/caja/pull/1568/commits/894b6a4518f6adc9558e571141aea7cf12a5d6a6) is implemented differently, in a way that also improves the other permission combo boxes: 30e3cf5dfbfd138a84a2a29ba3483a48f0cb04a4
* [Property browser: add accessible labels](https://github.com/mate-desktop/caja/pull/1568/commits/6023ae04d7044877203ba521cca0557f770504cf) is fixed properly and more thoroughly by fixing the accessible implementation of the related widgets: a8e36f3bfe154237b00853e2804a50afe5531f30
* [Notes viewer sidebar: allow keyboard navigation with Shift-Tab](https://github.com/mate-desktop/caja/pull/1568/commits/693434b47626d8b021b24117cdbe50344c5fdd30): I don't see any reason for this, one can navigate with <kbd>Ctrl+Tab</kbd> or <kbd>Ctrl+Shift+Tab</kbd> already in there, like in other text views.

Basically this fixes the issues identified in #1568, improving some results, trying to use generic solutions were possible, and not trying to workaround outdated Orca bugs.  Tested with Orca 48.

### DIFF
--- a/eel/eel-labeled-image.c
+++ b/eel/eel-labeled-image.c
@@ -2277,22 +2277,76 @@ eel_labeled_image_accessible_init (EelLabeledImageAccessible *accessible)
  * GTK accessible parent.  CheckButton and ToggleButton accessible objects
  * share the same type as GTK uses the same there. */
 
+typedef GtkButton EelLabeledImageButton;
+typedef GtkButtonClass EelLabeledImageButtonClass;
+
+G_DEFINE_TYPE (EelLabeledImageButton,
+               eel_labeled_image_button,
+               GTK_TYPE_BUTTON)
+
 static void
-eel_labeled_image_button_class_init (GtkWidgetClass *klass)
+eel_labeled_image_button_class_init (EelLabeledImageButtonClass *klass)
 {
-    gtk_widget_class_set_accessible_type (klass, eel_labeled_image_button_accessible_get_type ());
+    gtk_widget_class_set_accessible_type (GTK_WIDGET_CLASS (klass), eel_labeled_image_button_accessible_get_type ());
 }
 
 static void
-eel_labeled_image_toggle_button_class_init (GtkWidgetClass *klass)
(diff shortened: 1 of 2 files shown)


lukefromdc (approved):
Seems to work, Orca seems to give rational output even if the details take a while to speak. In the file properties dialog for an image, note that "image tab" is spoken but the contents of that tab (provided by an extension) are not. That's probably an issue for a caja-extensions PR though.

Someone more familiar with what Orca would ideally be saying should test this, but as of right now I have Orca reading rational output from caja in the mate-wayland session. Hats off to the Orca devs for getting it working in wayland...

cwendling:
I got a positive testing off-PR for the changes here, so I'm gonna merge this.  I also got some comments on other details that I might try to address separately.

Issue #1549, opened 2021-09-03
Failed to set text from markup due to error parsing markup


MelvilleTheatre:
#### Expected behaviour
Create a directory that contains an ampersand in the directory name.
Now go to that directory using caja and see what shows up in /var/log/messages when you enter that directory.

#### Actual behaviour
/var/log/messages says: Failed to set text 'a & b' from markup due to error parsing markup: Error on line 1: Entity did not end with a semicolon; most likely you used an ampersand character without intending to start an entity - escape ampersand as &amp;
This is very minor since everything seems to work right.  It just logs an error.

#### Steps to reproduce the behaviour
mkdir 'a & b'
tail -f /var/log/messages
Use caja to enter that directory.

#### MATE general version
1.24.1

#### Package version
caja-1.22.3-3.el8.x86_64

#### Linux Distribution
Centos 8

#### Link to bugreport of your Distribution (requirement)
Not part of the official Centos distribution.  This came from https://copr.fedorainfracloud.org/coprs/stenstorp/MATE/

PR #1819, opened 2025-05-03
Fix markup parsing error for paths with '&' in places sidebar

banad60:
**Fix for Markup Parsing Error with '&' in Paths in Caja**

**Problem**
Caja displays a GTK warning in ~/.xsession-errors when bookmark paths contain special characters like & (e.g., /home/user/WORX/MA&PA). This causes a markup parsing error in the "Places" sidebar.

**Reproduction**
1. Add a bookmark for a path with & (e.g., /home/user/WORX/MA&PA).
2. Hover the mouse over the bookmark in the sidebar.
3. Check ~/.xsession-errors for the error message:

> (caja:XXXX): Gtk-WARNING **: Failed to set text '/home/user/WORX/MA&PA' from markup due to error parsing markup: Error on line 1: Entity did not end with a semicolon; most likely you used an ampersand character without intending to start an entity - escape ampersand as &amp;

**Solution**
In src/caja-places-sidebar.c, the tooltip text in the add_place function is escaped with g_markup_escape_text before being inserted into the GtkListStore. This prevents parsing errors caused by special characters like &.

**Changes**
src/caja-places-sidebar.c: Escape tooltip text with g_markup_escape_text.
bookmarks_check_popup_sensitivity: Added check for NULL popup menu to avoid Gtk-CRITICAL error.

**Tests**
Locally tested on Ubuntu 24.04.2 with Caja 1.29.0.
Added bookmark with &, no GTK warnings in ~/.xsession-errors.

**Known Issues**
Xlib errors (ignoring invalid extension event 161 and 146) occur with tooltips. These are non-disruptive and likely related to the compositing manager (e.g., picom). They can be investigated later.

**Reference**
Relates to Issue #1549.

### DIFF
--- a/src/caja-places-sidebar.c
+++ b/src/caja-places-sidebar.c
@@ -80,6 +80,7 @@ typedef struct
     gboolean  drag_data_received;
     int       drag_data_info;
     gboolean  drop_occured;
+    gboolean  selection;
 
     GtkWidget *popup_menu;
     GtkWidget *popup_menu_open_in_new_tab_item;
@@ -173,7 +174,16 @@ static void  check_unmount_and_eject                   (GMount *mount,
         gboolean *show_unmount,
         gboolean *show_eject);
 
-static void bookmarks_check_popup_sensitivity          (CajaPlacesSidebar *sidebar);
+static void
+bookmarks_check_popup_sensitivity (CajaPlacesSidebar *sidebar)
+{
+    g_return_if_fail (CAJA_IS_PLACES_SIDEBAR (sidebar));
+    
+    if (sidebar->popup_menu != NULL) {
+        gboolean has_selection = (sidebar->selection != NULL);
+        gtk_widget_set_sensitive (sidebar->popup_menu, has_selection);
+    }
+}
 
 /* Identifiers for target types */
 enum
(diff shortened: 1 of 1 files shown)


raveit65:
> Note: The Travis CI build (#275081104) fails in the Fedora environment with the error No package 'libtool-ltdl-devel' found. This issue appears related to the Travis CI configuration (.travis.yml) or build scripts, not the changes in this PR. The build was successful locally on Ubuntu 24.04.2, as noted in the PR description. Could the maintainers review the Fedora dependencies?

Package still exist in fedora
https://koji.fedoraproject.org/koji/buildinfo?buildID=2641628

banad60:
Thanks for the feedback and for confirming that libtool-ltdl-devel exists in Fedora (Koji #2641628). The Travis CI build (Build #275081104) failure in Fedora with No package 'libtool-ltdl-devel' found seems to be a configuration issue in .travis.yml or the build scripts, unrelated to this PR's changes. The build succeeded locally on Ubuntu 24.04.2, as noted in the PR description.

Additionally, I noticed a Gtk-CRITICAL error in my local logs for mate-panel (gtk_widget_destroy: assertion 'GTK_IS_WIDGET (widget)' failed). This appears unrelated to this PR, as it occurs in a different module and does not affect Caja's functionality with this fix. Could the maintainers review the Fedora setup in Travis CI?

Let me know if further clarification or testing is needed!

lukefromdc:
I will definately need others on the team to help with that. I am not on Fedora
locally (and don't have a landline to download installers over) and don't have
a lot of understanding of how to fix travis problems.

banad60:
Thanks for the update, @lukefromdc! I understand the Travis CI issues (Fedora and Debian) are tricky and appreciate the team looking into them. My local build on Ubuntu 24.04.2 works fine, as noted in the PR. Happy to run additional tests or provide more details if needed. Let me know how I can assist!

lukefromdc:
Just confirmed the error message

lukefromdc (commented):
The code works, but why are you cutting out so many existing comments?
We need to keep those for maintainability in the future

banad60:
Thank you for confirming the bug and pointing out the issue with the removed comments. I agree that keeping the code well-documented is important, so I’ll work on restoring the comments to make everything clear again.

I’ll update the pull request with the restored comments in the next few days. It might take a little time, but I’ll let you know once it’s ready. Thanks for your patience and support!

lukefromdc:
Thanks, it gets here when it gets here

banad60:
It's finally here

lukefromdc (commented):
Just ran into a new problem and confirmed it happens only with this PR applied. If I right-click on any item in the places sidebar, the context menu comes up with all items greyed out and is unreponsive. Even to close it sometimes requires closing the entire window.

Left-click works as expected, setting the directory for the open window to the clicked item, Middle click opens it in a new tab, but right click fails.

Tested in the wayland session but this probably will also happen in the x11 session or at least the 100%  greyed out entries in the context menu would be expected.

The correct filename of the test folder is read when hovering over it so that part works

Making this a comment not a formal request for changes in case anything happens to me with the current mess in the US

lukefromdc:
Holding on the github-actions workflow approval as we already know we have a problem that needs to be fixed

banad60:
I can also confirm the error in the right-click menu under X11 with LightDM: gray, non-navigable menu items. I'll investigate this; it's probably related to the first commit.

cwendling (changes requested):
This is not reviewable.  Don't include seemingly random changes in a pull request.
The *meaningful* change here seem to simply be the `g_markup_escape_text()` call (which probably make sense indeed), the rest might or might not be interesting but it's definitely not related and must not be part of the same PR.  Also, such changes need be explained in the commit message.

luigifab:
It looks like it's working!

PR #1822, opened 2025-05-18
Try to use github action

yetist:
Try to use github action build


lukefromdc:
Let me know when these are merge-ready, this is not exactly something I can test with local builds on bare metal

yetist:
I think this PR is ready now, please review it, thanks.
The release test is here: https://github.com/yetist/caja/releases

@mate-desktop/core-team

yetist:
Currently, `gh release create --generate-notes` is used to generate release notes, which look like this: https://github.com/elastic/opbeans-node/releases

If we want to customize the release notes, we can use the `gh release create --notes-from-tag` command, but we need to write the content in the comment of the created tag, for example:

https://github.com/python-kasa/python-kasa/tags (Click the ellipsis to view the commit message of the tag)
https://github.com/python-kasa/python-kasa/releases (Release note generated from the tag commit message)

Which one do we prefer?  @lukefromdc @raveit65

lukefromdc (approved):
I'm fine with either approach to generating release notes. Nothing here I know how to test myself but see we have a sucessfull test release. I also agree with moving to a 1.30 release due to the soname mess in mate-desktop, we will need to fix that first before releasing 1.30 for real.

Approving based on the test results at https://github.com/yetist/caja/releases

oz123 (approved):

lukefromdc:
I have no idea how that works. If we have access to the credentials for that we should be 
able to reconfigure it. We did not always use travis

PR #1841, opened 2025-07-22
file: Fallback to fast-content-type if content-type is not set

### LINKED ISSUE #1840: caja developer with IDE open, can you have a look at this? PNG file mimetype on samba share.
If you are active on the caja codebase, could you have a look at this?

https://bugs.launchpad.net/ubuntu-mate/+bug/2077315

I've tried to make the research as complete as possible.

cwendling:
The G_FILE_ATTRIBUTE_STANDARD_CONTENT_TYPE attribute doesn't have to be always set. See https://gitlab.gnome.org/GNOME/gvfs/-/merge_requests/68 for more details. In that case, Caja fallbacks to the "application/octet-stream" type, which causes issues when opening the files. Let's fallback to the "standard::fast-content-type" attribute instead to fix issues when opening such files.

Imported from Nautilus commit: https://gitlab.gnome.org/GNOME/nautilus/-/commit/0e5978035b0fc87c91d7b93ed79c64d51b6d6825

Fixes #1840.

@sijskes could you give this a spin and see if it fixes it?  It's simply an import of the Nautilus fix, but it might just do it.

### DIFF
--- a/libcaja-private/caja-file.c
+++ b/libcaja-private/caja-file.c
@@ -2460,6 +2460,9 @@ update_info_internal (CajaFile *file,
 	}
 
 	mime_type = g_file_info_get_content_type (info);
+	if (mime_type == NULL) {
+		mime_type = g_file_info_get_attribute_string (info, G_FILE_ATTRIBUTE_STANDARD_FAST_CONTENT_TYPE);
+	}
 	if (eel_strcmp (file->details->mime_type, mime_type) != 0) {
 		changed = TRUE;
 		g_clear_pointer (&file->details->mime_type, g_ref_string_release);


lukefromdc (commented):
I didn't get any problems from this, built and ran fine, but I don't think I have any files that can give this a functional test and I don't use Samba as I don't have any Windows machines or macs

sijskes:
> It looks good (Caja 1.26.4). But I'm not sure how to see the difference (I've already seen _application/octet-stream_, but I can't remember where).

Enable column mime-type in caja.
Enable list mode. look at PNG files.

When local file, it shows image/png
When samba file, it should show the same or similar, like application/x-extension-png

Probably the whole mate sourcecode should be checked for G_FILE_ATTRIBUTE_STANDARD_CONTENT_TYPE handling.
Because Eye of mate EOM also crashes on opening remote PNG files.

sijskes:
crosslinking issues: https://github.com/mate-desktop/eom/issues/360

lukefromdc:
> > It looks good (Caja 1.26.4). But I'm not sure how to see the difference (I've already seen _application/octet-stream_, but I can't remember where).
> 
> Enable column mime-type in caja. Enable list mode. look at PNG files.
> 
> When local file, it shows image/png When samba file, it should show the same or similar, like application/x-extension-png
> 
> Probably the whole mate sourcecode should be checked for G_FILE_ATTRIBUTE_STANDARD_CONTENT_TYPE handling. Because Eye of mate EOM also crashes on opening remote PNG files.

Is this working correctly for you? If so we can merge this

lukefromdc:
I am on Debian unstable myself but cannot really advise as I have no VM experience and no landline so
no experience playing with distros. Maybe we call this done and merge it?

lukefromdc:
Can anyone else test this for function?

cwendling:
@sijskes can't you test this with your regular machine that shows the issue?  Testing a local Caja isn't the easiest of all, but it's not the end of the world either, you don't necessarily have to replace your system's version.

Otherwise maybe I can try this at some point, but not having any Samba share at hand it'll require me to figure some stuff out, so don't hold your breath.

luigifab:
I tried with sshfs but I think there is no problem, and I don't have access to any Samba directory.

vkareh (approved):
This fixes the issue for me on an ssh mount - thanks!

PR #1855, opened 2025-10-02
file-operations: Use correct filename when restoring from Trash

### LINKED ISSUE #1854: Trash "Restore" not using .trashinfo name
Was writing my own code to trash files according to XDG.  I noticed while the file is placed in its correct directory when restored. Its file name is the file name it had in ".local/share/Trash/files" which included suffix additions due to multiple files with the same name in ".local/share/Trash/files".

### Original file
* /some/path/foobar.txt
### $HOME/.local/share/Trash/files
* foobar.txt
* **_foobar.txt.1_**
* foobar.txt.2
### Actual Restored file
* /some/path/foobar.txt.1
### Expected Restored file
* /some/path/foobar.txt

Is this correct? Caja Version 1.26.3

vkareh:
When restoring files from trash, use the original filename from the .trashinfo file instead of the internal trash filename.

Adapted from Nautilus:
- https://gitlab.gnome.org/GNOME/nautilus/commit/ecee8be8
- https://gitlab.gnome.org/GNOME/nautilus/commit/0d0d95d2

Fixes #1854

### DIFF
--- a/libcaja-private/caja-file-operations.c
+++ b/libcaja-private/caja-file-operations.c
@@ -3393,6 +3393,28 @@ get_target_file (GFile *src,
 		char *basename;
 
 		basename = g_file_get_basename (src);
+
+		/* if file is being restored from trash make sure it uses its original name */
+		if (g_file_has_uri_scheme (src, "trash")) {
+			GFileInfo *info;
+
+			info = g_file_query_info (src,
+						  G_FILE_ATTRIBUTE_TRASH_ORIG_PATH,
+						  0, NULL, NULL);
+
+			if (info) {
+				const char *orig_path;
+
+				orig_path = g_file_info_get_attribute_byte_string (info, G_FILE_ATTRIBUTE_TRASH_ORIG_PATH);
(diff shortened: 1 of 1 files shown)


lukefromdc (commented):
I just tested this, and what I got was this and was same on master and with this PR applied: the internal filename in the trash was only visible by copying the file (not the file contents) to the clipboard and dropping into Pluma, which renders the path and true filename. On restore, the filename shown after putting two files with identical filenames but different paths (thus different internal names in the trash) retained the changed names, with one file going from Test111.mp4 to Test111.2.mp4 in the trash-and retaining that name on restore. Again, this PR didn't change this behavior here. Tested twice to ensure I had installed the build and restarted caja.

This is with a local build of gtk3-classic version 3.24.51 and current glib as provided by Debian Unstable. Running under wayland but that should have no effect on this PR

mbkma (approved):
This looks right. Using the original path metadata from the trash info to recover the filename matches the expected restore behavior, and keeping the fallback handling around the attribute lookup makes the change safe.

PR #1869, opened 2026-01-29
ci: remove redundant GH CLI installation step

mbkma:
sorry, for atril I accidently pushed this directly to master (https://github.com/mate-desktop/atril/commit/a6a92da563e1bdb6e6520b5ec5f67a7dcef4056a). I can revert it if it is not wished. Also maybe we should set it so that no one can push directly to master by default.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,11 +24,6 @@ jobs:
       - name: Build the source code
         run: .github/workflows/builds.sh
 
-      - name: Install GH CLI
-        uses: dev-hanz-ops/install-gh-cli-action@v0.2.1
-        with:
-          gh-cli-version: 2.39.1
-
       - name: Create github release
         run: |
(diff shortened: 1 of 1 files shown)


vkareh:
Now we should submit another PR to see how the pipelines behave...

mbkma:
seems to work https://github.com/mate-desktop/caja/actions/runs/21495705126

raveit65:
@yetist  did the test in his own repo https://github.com/mate-desktop/mate-desktop/pull/647#issuecomment-2892720625

mbkma:
btw. we can also get the sha256sum back if you want :) https://github.com/mbkma/libmateweather/blob/e13034d0a45ae2270ff5e6831e5771befffcc197/.github/workflows/release.yml#L20

mbkma:
hm no, I did not find a way to do that. but maybe i build the tarball on my local machine and then upload it to the release.

Issue #1871, opened 2026-02-15
Thumbnails Get Smaller on Clicking


SchrodingersCpp:
Mouse clicking (with any button) on thumbnails (at least, tested with image files) makes them smaller. Looks like thumbnails match actual size of their images.

```
Debian 13 (trixie) 64-bit
Kernel Linux 6.12.69+deb13-amd64 x86_64
MATE 1.26.2
Caja 1.26.4
```

![Image](https://github.com/user-attachments/assets/098d19e5-350e-4e3a-8489-d327c59ca39f)

PR #1873, opened 2026-02-23
caja-file: fix thumbnails shrinking when loaded on select

vkareh:
Small images were loading thumbnails at their native pixel dimensions after being selected, instead of being scaled to the icon view's display size. This means that when clicking or selecting those images, they visually shrunk.

Fixes https://github.com/mate-desktop/caja/issues/1871

### DIFF
--- a/libcaja-private/caja-file.c
+++ b/libcaja-private/caja-file.c
@@ -4611,13 +4611,8 @@ caja_file_get_icon (CajaFile *file,
 			h = gdk_pixbuf_get_height (raw_pixbuf);
 
 			s = MAX (w, h);
-			/* Don't scale up small thumbnails in the standard view */
-			if (s <= cached_thumbnail_size) {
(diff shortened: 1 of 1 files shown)


lukefromdc:
I should be able to test this one tonight, the issue is an old nuisiance here.

lukefromdc:
For some reason I was just unable to duplicate the issue, should not be affected by wayland vs x11. Using caja from git master, wondering if recent glib changes affected this

vkareh:
The easiest way I've found to duplicate this is opening caja on `/usr/share/icons/hicolor/16x16/apps/` in a new session and then clicking a random icon and moving with the keyboard arrows. This consistently triggers the issue for me, where the icons show up initially at whatever size caja is configured for, then go down to 16px. I'm on a HiDPI display (scale=2) in case this matters.

lukefromdc:
OK to merge this without being able to duplicate the original issue anymore, or should we wait for someone who still has that problem to test it?

PR #1874, opened 2026-02-26
icon-view: Display hidden files with transparency

### LINKED ISSUE #1872: Hidden folders/files must be different color (or transparent)
Image

I think it would be nice to support a bit transparent color on hidden items similar to, Pantheon, Dolphin, Thunar..

vkareh:
Render hidden files with some transparency to visually distinguish them from regular files. Files become fully opaque when selected or highlighted.

Fixes #1872

### DIFF
--- a/libcaja-private/caja-icon-canvas-item.c
+++ b/libcaja-private/caja-icon-canvas-item.c
@@ -48,6 +48,7 @@
 #include "caja-file-utilities.h"
 #include "caja-global-preferences.h"
 #include "caja-icon-private.h"
+#include "caja-file.h"
 
 #define EMBLEM_SPACING 2
 
@@ -1912,6 +1913,7 @@ caja_icon_canvas_item_draw (EelCanvasItem *item,
     GdkPixbuf *emblem_pixbuf;
     cairo_surface_t *temp_surface;
     GtkStyleContext *context;
+    gboolean render_as_hidden = FALSE;
 
     container = CAJA_ICON_CONTAINER (item->canvas);
     gboolean is_rtl;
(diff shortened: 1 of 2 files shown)


cwendling (commented):
I couldn't test yet, but looks fairly reasonable.

This said, shouldn't the text also be dimmed?  Especially in list/tree mode, having only the icon dimmed might not make so much of a difference, does it?

vkareh:
@cwendling 
> This said, shouldn't the text also be dimmed? Especially in list/tree mode, having only the icon dimmed might not make so much of a difference, does it?

Good question, but I'm not sure about that. The icon is a quick visual cue, whereas the text is meant to be read. Making it less opaque will affect contrast and therefore readability and accessibility. For comparison, nautilus doesn't seem to dim the text either.

mbkma:
I think I am doing something wrong, for me I can not see any transparency.

Screenshot at 2026-04-10 16-36-01

vkareh:
Because caja is running your desktop session directly, you might need to do something like:
```
$ killall caja
$ caja --no-desktop
```

lukefromdc (approved):
I could not tell if there were any new build warnings as I had to do a new build directory to test this on top of the other two PR's so lots of old warnings we always see. That said, the build finished, the rendering change seems elegant to me, and so does the actual behavior. Approving

lukefromdc:
Also note that I agree with @cwendling that dimming the text too could create issues for some users, whereas since this is the file manager its the text not the icon you have to read or have spoken to identify the directory you are looking for. Thus, dimming icons in adverse lighting should not make the files hard to read, but dimming text would.

PR #1878, opened 2026-03-07
ci: add stale-issues.yml

mbkma:
I would like to create  a new repo called `ci-templates`, where we can put things like this script. After 90 days of inactivity in an issue it writes a comment and after 30 more days of inactivity it closes the issue, unless it has some specific labels. Thoughts?

### DIFF
--- a/.github/workflows/stale-issues.yml
+++ b/.github/workflows/stale-issues.yml
@@ -1,0 +1,44 @@
+name: Stale issue management
+
+on:
+  schedule:
+    - cron: "0 2 * * *"  # run daily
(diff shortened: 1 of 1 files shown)


mbkma:
I understand your points, but I still think we should take action on this. With 500+ issues in Caja, it becomes very difficult to distinguish between critical bugs, minor issues, and reports that aren’t actually valid.

It might make sense to introduce clearer labeling (e.g., severity, reproducibility, scope) and to close issues that are out of scope, not reproducible, or otherwise not actionable. If any of those turn out to be important, they can always be reopened later.

PR #1879, opened 2026-03-07
allow detachable tabs

### LINKED ISSUE #454: Drag tabs to a new Caja window
It would be nice if the existing tab can be dragged & dropped to create a new Caja window (with the tab moved from the old Caja windows to the new one).

Similarly like it works in Mate terminal.

mbkma:
fixes #454

### DIFF
--- a/src/caja-navigation-window-pane.c
+++ b/src/caja-navigation-window-pane.c
@@ -33,6 +33,7 @@
 #include "caja-navigation-window-pane.h"
 #include "caja-window-private.h"
 #include "caja-window-manage-views.h"
+#include "caja-application.h"
 #include "caja-pathbar.h"
 #include "caja-location-bar.h"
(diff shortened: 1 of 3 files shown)


vkareh (commented):
This works really well, thanks!

vkareh (approved):

lukefromdc:
I just found that this works as intended in x11, but in wayland a detached tab must be dragged to any navigation window and to the toolbar, menubar, or window decoration to detach the tab to a new window. The desktop although a spatial not a navigation window does not work, nor does the portion of any other caja window where the icons are shown. Dragging to another app's window also doesn't work but that's really no surprise.

There is also an occasional but delayed crash after using this in both x11 and wayland, most of the time that does not occur. Nothing shows up in dmesg and when I ran it as root from the terminal I was  unable to invoke the crash

PR #1882, opened 2026-03-13
fm-list-view: Remember column widths per directory

### LINKED ISSUE #18: Column-width remembering issue in Caja
In Caja, when I have files listed in list-mode (ctrl-2), I prefer to set the width of the filename column. Every time I open a new tab to the same directory, the column width resets to the max-width for the widest file/directory name... is there any way to get it to remember the column widths (like Nautilus did)?
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/2121173-column-width-remembering-issue-in-caja?utm_campaign=plugin&utm_content=tracker%2F651521&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F651521&utm_medium=issues&utm_source=github).
</bountysource-plugin>

vkareh:
When the user manually resizes columns in List View, save the widths as per-directory metadata along with the column order and visibility. Saved widths are restored when revisiting the directory.

Data is stored in GIO for each directory as `caja-list-view-column-widths` using the format `[col1:width1, col2:width2, etc...]`. Clearing this data (e.g. clicking on `View -> Reset View to Defaults`) will clear the GIO metadata, so next time you view the directory it will auto-expand as normal.

You can view the metadata with `gio info -a "metadata::*" /path/to/directory`

Fixes #18


lukefromdc (approved):
This works, and not only saves column widths in the treeview but does so on a per-directory basis and across restarts of caja. The issue with "reset view to defaults" is confirmed, but is a lesser issue that as marked in the code comments can be dealt with later if necessary. Approving

PR #1889, opened 2026-05-19
location-bar: add trailing slash to displayed location path

### LINKED ISSUE #874: Ctrl+L Should Add Trailing Slash
Most other file browsers add a trailing slash to the end of the file location when using Ctrl+L to open the location bar. With the trailing slash subfolders can be typed in immediately without having to hit the slash key first and without it, those of us that have recently switched from another file manager frequently forget to add the slash and type in the location address incorrectly.

vkareh:
When pressing Ctrl+L to show the location bar, the displayed path now includes a trailing slash so users can immediately start typing a subfolder name without having to press the slash key first.

Fixes #874


mbkma (approved):
Small but sensible UX fix. Appending a trailing slash only in the displayed path makes the location bar clearer for directory paths without changing the underlying location handling.

PR #1890, opened 2026-05-19
preferences: add option to disable thumbnails in large directories

### LINKED ISSUE #1304: Option to disable previews on large directories
The file preview feature significantly slows down load times for giant directories. Since there is already a feature to count the items in a folder, it would be nice to be able to set a cut-off for showing the previews. "Only show in folders with less than x items" seems like a logical addition to the current "only show for files smaller than x mb."

vkareh:
Directories with many files have significant performance issues when generating thumbnails. This adds a preference that disables thumbnails when the directory exceeds the threshold.

Fixes #1304

Screenshot at 2026-05-18 23-09-31

### DIFF
--- a/data/caja-file-management-properties.ui
+++ b/data/caja-file-management-properties.ui
@@ -68,6 +68,32 @@
       </row>
     </data>
   </object>
+  <object class="GtkListStore" id="model11">
+    <columns>
+      <!-- column-name gchararray -->
+      <column type="gchararray"/>
+    </columns>
+    <data>
+      <row>
+        <col id="0" translatable="yes">100 files</col>
+      </row>
+      <row>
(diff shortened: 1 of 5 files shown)


lukefromdc (approved):
Works fine and no build time surprises though I was not monitoring warnings as they scrolled by. For all the volume of file changes, the actual core addition is relatively simple, the UI change for users to set the maximum number of files to thumbnail is neat, and the rest is supporting changes. Approving

Mupli:
Add disabled too!

vkareh:
@Mupli 
> Add disabled too!

You can already select "Show thumbnails: Never"
Screenshot at 2026-05-22 10-41-01

PR #1891, opened 2026-05-26
background: always redraw when monitor configuration changes

### LINKED ISSUE #352: Background placement corrupted after numerous hotplug monitor events
#### Expected behaviour
During a fresh start of mate, I have 3x 4k displays, and use a single 11520x2160 hi-def image across all 3 displays set as a span background in system settings that fills the entire 3x displays set up as a single display with nvidia settings.  Upon shutdown/poweroff of displays going to sleep, the desktop should awaken with the background in the same place, positioned across all 3 displays without a border and without scaling the image.

#### Actual behaviour
When the pc wakes up and the displays are powered up for use, randomly the desktop will change the background, scaled across 2 displays only in letterbox with borders top and bottom, and then starts the image again on the 3rd split in half in the same letterbox with the other half un-rendered as though on a 4th display.  It usually doesn't do this on the first, and sometimes second day of powering off all 3 displays, but eventually it'll awaken broken.

Entering system settings, appearance, and changing the background to something different, and back resets the image properly scaled across a 3 displays again.

These displays are Samsung 4k Televisions connected via HDMI 2.0, and do not do dpms shutdown events, so once screen saver and screen power down occurs, the displays are set to shut off after 15 minutes, typically with work-day cycle.  This causes linux and graphics to see all 3 displays removed via hotplug events, which causes a number of OS/DE quirks, but this in particular causes desktop background placement to go haywire after a few times (2x or more), and rather than full-screen fit the 11520x2160 image across all 3 displays, scrunches the image as described above

This has occurred in prior versions of mate for years, but filing an official report finally.

Boot PC with 3x (4k HDMI) displays connected to an nvidia 1070-class device, power off all 3 displays during display sleep several times during the course of use for an hour or more, powering the displays back on again, activate displays with mouse/keyboard, and at lock screen notice that large 11520x2160 is scaled down in letterbox across displays.  This last time this occurred within 2 days after reboot powering displays off 2 nights in a row.

#### MATE general version
Mate Desktop 1.20.4-1

#### Package version
[user@host ~]$ pacman -Q | grep mate
libmatekbd 1.20.2-1
libmatemixer 1.20.1-1
libmateweather 1.20.2-1
mate-applet-dock 0.87-1
mate-applet-streamer 0.4.0-1
mate-applets 1.20.2-1
mate-backgrounds 1.20.0-1
mate-calc 1.20.3-1
mate-common 1.20.0-1
mate-control-center 1.20.4-1
mate-desktop 1.20.4-1
mate-icon-theme 1.20.3-1
mate-icon-theme-faenza 1.20.0-1
mate-media 1.20.2-1
mate-menu 18.04.3-1
mate-menus 1.20.2-1
mate-netbook 1.20.1-1
mate-notification-daemon 1.20.2-1
mate-panel 1.20.3-1
mate-polkit 1.20.1-1
mate-power-manager 1.20.2-2
mate-screensaver 1.20.2-1
mate-sensors-applet 1.20.2-2
mate-session-manager 1.20.1-1
mate-settings-daemon 1.20.3-1
mate-system-monitor 1.20.1-1
mate-terminal 1.20.1-1
mate-themes 3.22.18-1
mate-user-guide 1.20.2-1

#### Linux Distribution
Arch Linux, 64bit, Nvidia 1070GTX graphics, nvidia 415.25-1 driver

vkareh:
The monitors-changed signal was handled by the same function as size-changed, which only redraws when the screen dimensions differ. When a monitor is removed without changing the virtual screen size, the desktop background was not repainted.

This adds a separate handler that always triggers a redraw, since a monitor layout change requires repainting regardless of whether the dimensions changed.

Fixes mate-desktop/mate-desktop#352

### DIFF
--- a/eel/eel-background.c
+++ b/eel/eel-background.c
@@ -642,6 +642,12 @@ screen_size_changed (GdkScreen *screen, EelBackground *background)
 }
 
 static void
+screen_monitors_changed (GdkScreen *screen, EelBackground *background)
+{
+    g_signal_emit (background, signals[APPEARANCE_CHANGED], 0);
+}
+
+static void
 widget_realized_setup (GtkWidget     *widget,
                        EelBackground *self)
 {
@@ -664,7 +670,7 @@ widget_realized_setup (GtkWidget     *widget,
         g_signal_handler_disconnect (screen, self->details->screen_monitors_handler);
     }
     self->details->screen_monitors_handler =
-        g_signal_connect (screen, "monitors-changed", G_CALLBACK (screen_size_changed), self);
+        g_signal_connect (screen, "monitors-changed", G_CALLBACK (screen_monitors_changed), self);
 
     if (GDK_IS_X11_DISPLAY (display))
     {


mbkma (approved):
This looks good. Handling monitor configuration changes separately from plain size changes is the right fix here, and always redrawing on 'monitors-changed' should cover the cases where the layout changes without a screen resize.

PR #1892, opened 2026-06-03
file-manager: fix double-free and relative path crash for submodule gitdir pointers

Inspirati:
## Problem

When Caja enumerates a directory whose subdirectories contain a `.git` *file*
(rather than a `.git` directory — as used by git submodules and linked
worktrees), `get_git_branch()` crashes intermittently via heap corruption.

The root cause is a **double-free**:

1. Inside `get_git_branch()`, when `.git` is a regular file the code calls
   `g_free(git_path)` on its own `const char *` parameter (to free the
   caller's allocation before reassigning the local variable).
2. The caller then calls `g_free(git_path)` again on the same pointer after
   `get_git_branch()` returns.

The resulting heap corruption is non-deterministic — it manifests as a crash
elsewhere in the process, not at the site of the double-free, which is why a
desktop code review found no obvious NULL-dereference path in the feature code.

A secondary issue: `g_build_filename()` does not canonicalise `../` components
in relative `gitdir:` paths. If the resolved path does not exist on disk
(e.g. an uninitialised submodule), `realpath()` returns NULL; without a guard
this would be passed into subsequent GLib calls.

## Fix

- Introduce a `resolved_gitdir` local variable to hold the heap-allocated
  resolved path. The `git_path` parameter (which belongs to the caller) is
  never freed inside the function.
- Use `realpath()` to canonicalise relative `gitdir:` paths containing `../`
  components. The result is copied into GLib-managed memory with `g_strdup()`
  and the `malloc()`-allocated `realpath()` buffer is freed with `free()`.
- Guard against a NULL return from `realpath()` — when the target path is
  absent on disk, `effective_gitdir` is set to NULL and the subsequent
  `g_file_test()` call is skipped cleanly.
- Add `#include <stdlib.h>` for `realpath()` and `free()`.

## Crash reproduction

Navigate to a parent directory in Caja icon view that contains a subdirectory
with a `.git` file whose `gitdir:` line holds a relative path
(e.g. `gitdir: ../../../../../.git/modules/...`). With the Display Git Branch
preference enabled, Caja crashes on directory enumeration. After this fix,
enumeration completes without crashing.

### DIFF
--- a/src/file-manager/fm-icon-container.c
+++ b/src/file-manager/fm-icon-container.c
@@ -22,6 +22,7 @@
    Author: Michael Meeks <michael@ximian.com>
 */
 #include <config.h>
+#include <stdlib.h>
 #include <string.h>
 
 #include <glib/gi18n.h>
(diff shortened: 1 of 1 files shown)


lukefromdc (commented):
Looking closely at this due to its AI origin. I don't see any obvious AI screwups here, but I don't know this part of the code well enough to follow the flow of it from just what we see here

vkareh:
The commit in this PR is included in #1893, which I reviewed earlier today. I think we should leave this here open for now until we decide whether we use one or the other (the fix is valid in either case, with some small changes)

PR #1893, opened 2026-06-03
file-manager: classify git worktrees and submodules with distinct icon labels

Inspirati:
## Summary

Extends the git branch display feature in Caja's icon view to visually
distinguish three types of git-tracked directories:

| Directory type  | Label          |
|-----------------|----------------|
| Standard repo   | `[branch]`     |
| Linked worktree | `[branch ⎇]`  |
| Submodule       | `[branch ⊂]`  |

The ⎇ symbol (U+2387, ALTERNATIVE KEY) is widely used by git prompt tools
to denote linked worktrees. The ⊂ symbol (U+2282, SUBSET OF) reflects that
a submodule is a repository contained within another.

The suffix is appended for both normal branch names and detached HEAD
labels (e.g. `[detached: abc1234 ⎇]`).

## Implementation

Classification is performed inside `get_git_branch()` after the gitdir
path has been fully resolved, by checking for the well-known path
components that git always uses:

```c
if (strstr (effective_gitdir, ".git/worktrees/") != NULL)
    gitdir_type = GITDIR_WORKTREE;
else if (strstr (effective_gitdir, ".git/modules/") != NULL)
    gitdir_type = GITDIR_SUBMODULE;
else
    gitdir_type = GITDIR_STANDARD;
```

A file-scoped `GitdirType` enum encodes the three cases. No changes are
required outside `fm-icon-container.c`.

## Relationship to PR #1892

This branch includes the bug fix from PR #1892 (double-free and
`realpath()` for relative `gitdir:` paths) as a prerequisite commit,
since the feature builds directly on the `effective_gitdir` variable
introduced by that fix. Maintainers may choose to merge #1892 first and
rebase this branch, or merge both together.

## Testing

A representative test tree covering all three cases can be created with:

```bash
# Standard repository
git init main-repo && cd main-repo
git checkout -b main
echo "test" > README.md && git add . && git commit -m "init"

# Linked worktree (produces .git file with absolute gitdir path)
git checkout -b feature-xyz && git checkout main
git worktree add ../feature-worktree feature-xyz

# Submodule (produces .git file with relative ../  gitdir path)
cd .. && git init container-repo && cd container-repo
git checkout -b main
echo "container" > README.md && git add . && git commit -m "init"
git -c protocol.file.allow=always submodule add ../main-repo submodule
git commit -m "Add submodule"
```

Navigate to the parent of `main-repo/` and `feature-worktree/` in Caja
icon view to see `[main]` and `[feature-xyz ⎇]` respectively. Navigate
into `container-repo/` to see `[main ⊂]` on the `submodule/` icon.

The submodule case (`gitdir: ../.git/modules/submodule`) also exercises
the relative-path crash fix from PR #1892 — if enumeration of
`container-repo/` completes without crashing, both fixes are working.

### DIFF
--- a/src/file-manager/fm-icon-container.c
+++ b/src/file-manager/fm-icon-container.c
@@ -22,7 +22,6 @@
    Author: Michael Meeks <michael@ximian.com>
 */
 #include <config.h>
-#include <stdlib.h>
 #include <string.h>
 
 #include <glib/gi18n.h>
@@ -327,17 +326,9 @@ get_git_branch (const char *git_path) {
                 gchar *relative = g_strstrip (contents + git_dir_prefix_len);
                 if (!g_path_is_absolute (relative))
                 {
-                    gchar *base   = g_path_get_dirname (git_path);
-                    gchar *joined = g_build_filename (base, relative, NULL);
+                    gchar *base = g_path_get_dirname (git_path);
+                    resolved_gitdir = g_canonicalize_filename (relative, base);
                     g_free (base);
-                    /* realpath resolves ../ components; returns NULL if path does not exist */
-                    char *rp = realpath (joined, NULL);
-                    g_free (joined);
-                    if (rp != NULL)
-                    {
-                        resolved_gitdir = g_strdup (rp);
-                        free (rp);
-                    }
                 }
                 else
                 {
--- a/src/file-manager/fm-icon-container.c
+++ b/src/file-manager/fm-icon-container.c
@@ -316,7 +316,7 @@ get_git_branch (const char *git_path) {
 
     if (g_file_test (git_path, G_FILE_TEST_IS_REGULAR))
     {
-        /* It's a file — submodule or worktree. Read it to resolve the real git dir. */
+        /* It's a file, thus we are probably dealing with a submodule, so read it to resolve the real git dir */
         gchar *contents = NULL;
         if (g_file_get_contents (git_path, &contents, NULL, NULL) && contents)
         {
@@ -357,9 +357,9 @@ get_git_branch (const char *git_path) {
     if (effective_gitdir != NULL && g_file_test (effective_gitdir, G_FILE_TEST_IS_DIR))
     {
         GitdirType gitdir_type;
-        if (strstr (effective_gitdir, ".git/worktrees/") != NULL)
+        if (strstr (effective_gitdir, "/.git/worktrees/") != NULL)
             gitdir_type = GITDIR_WORKTREE;
-        else if (strstr (effective_gitdir, ".git/modules/") != NULL)
+        else if (strstr (effective_gitdir, "/.git/modules/") != NULL)
             gitdir_type = GITDIR_SUBMODULE;
         else
             gitdir_type = GITDIR_STANDARD;
--- a/src/file-manager/fm-icon-container.c
+++ b/src/file-manager/fm-icon-container.c
@@ -22,6 +22,7 @@
    Author: Michael Meeks <michael@ximian.com>
 */
 #include <config.h>
+#include <stdlib.h>
 #include <string.h>
 
 #include <glib/gi18n.h>
@@ -301,13 +302,15 @@ fm_icon_container_get_icon_text_attribute_names (CajaIconContainer *container,
  **/
 static char *
 get_git_branch (const char *git_path) {
+    gchar *resolved_gitdir = NULL;
     gchar *head_content = NULL;
     gchar *branch_name = NULL;
     gchar *head_path = NULL;
+    const gchar *effective_gitdir;
 
     if (g_file_test (git_path, G_FILE_TEST_IS_REGULAR))
     {
-        /* It's a file, thus we are probably dealing with a submodule, so read it to resolve the real git dir */
+        /* It's a file — submodule or worktree. Read it to resolve the real git dir. */
         gchar *contents = NULL;
         if (g_file_get_contents (git_path, &contents, NULL, NULL) && contents)
         {
@@ -316,19 +319,38 @@ get_git_branch (const char *git_path) {
             {
                 size_t git_dir_prefix_len = strlen (git_dir_prefix);
                 gchar *relative = g_strstrip (contents + git_dir_prefix_len);
-                gchar *base = g_path_get_dirname (git_path);
-                g_free (git_path);
-                git_path = g_build_filename (base, relative, NULL);
-                g_free (base);
+                if (!g_path_is_absolute (relative))
+                {
+                    gchar *base   = g_path_get_dirname (git_path);
+                    gchar *joined = g_build_filename (base, relative, NULL);
+                    g_free (base);
+                    /* realpath resolves ../ components; returns NULL if path does not exist */
+                    char *rp = realpath (joined, NULL);
+                    g_free (joined);
+                    if (rp != NULL)
+                    {
+                        resolved_gitdir = g_strdup (rp);
+                        free (rp);
+                    }
+                }
+                else
+                {
+                    resolved_gitdir = g_strdup (relative);
+                }
             }
             g_free (contents);
         }
+        effective_gitdir = resolved_gitdir;   /* may be NULL if path absent on disk */
+    }
+    else
+    {
+        effective_gitdir = git_path;
     }
 
     /* Extract the current git branch name of the repository from file HEAD within .git directory */
-    if (g_file_test (git_path, G_FILE_TEST_IS_DIR))
+    if (effective_gitdir != NULL && g_file_test (effective_gitdir, G_FILE_TEST_IS_DIR))
     {
-        head_path = g_build_filename (git_path, "HEAD", NULL);
+        head_path = g_build_filename (effective_gitdir, "HEAD", NULL);
         if (g_file_get_contents (head_path, &head_content, NULL, NULL))
         {
             g_strstrip (head_content);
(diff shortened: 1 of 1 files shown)


src/file-manager/fm-icon-container.c
```diff
@@ -347,9 +353,20 @@ get_git_branch (const char *git_path) {
...
 
-    /* Extract the current git branch name of the repository from file HEAD within .git directory */
+    /* Classify the git context, then extract branch name from HEAD */
     if (effective_gitdir != NULL && g_file_test (effective_gitdir, G_FILE_TEST_IS_DIR))
     {
+        GitdirType gitdir_type;
+        if (strstr (effective_gitdir, ".git/worktrees/") != NULL)
+            gitdir_type = GITDIR_WORKTREE;
+        else if (strstr (effective_gitdir, ".git/modules/") != NULL)
+            gitdir_type = GITDIR_SUBMODULE;
```
vkareh: These could theoretically match the wrong thing, like cloning a bare `foo.git` that contains a `modules` directory, the type indicator would point to the wrong directory `foo.git/` instead of `foo.git/.git/`

You can compare like `"/.git/modules/"` (note the leading `/`) to make sure it only captures the actual `.git` directory.

src/file-manager/fm-icon-container.c
```diff
@@ -347,9 +353,20 @@ get_git_branch (const char *git_path) {
...
+        else
+            gitdir_type = GITDIR_STANDARD;
+
+        const char *suffix = (gitdir_type == GITDIR_WORKTREE) ? " ⎇" :
+                             (gitdir_type == GITDIR_SUBMODULE) ? " ⊂" : "";
```
vkareh, line 359: Would these render in all systems, or would it show a tofu box instead?

src/file-manager/fm-icon-container.c
```diff
@@ -316,19 +325,49 @@ get_git_branch (const char *git_path) {
...
+                    gchar *joined = g_build_filename (base, relative, NULL);
+                    g_free (base);
+                    /* realpath resolves ../ components; returns NULL if path does not exist */
+                    char *rp = realpath (joined, NULL);
+                    g_free (joined);
+                    if (rp != NULL)
+                    {
+                        resolved_gitdir = g_strdup (rp);
+                        free (rp);
+                    }
```
vkareh: This entire block can probably be replaced with
```c
gchar *base = g_path_get_dirname (git_path);
resolved_gitdir = g_canonicalize_filename (relative, base);
g_free (base);
```

src/file-manager/fm-icon-container.c
```diff
@@ -296,18 +297,26 @@ fm_icon_container_get_icon_text_attribute_names (CajaIconContainer *container,
...
+    gchar *resolved_gitdir = NULL;
     gchar *head_content = NULL;
     gchar *branch_name = NULL;
     gchar *head_path = NULL;
+    const gchar *effective_gitdir;
 
     if (g_file_test (git_path, G_FILE_TEST_IS_REGULAR))
     {
-        /* It's a file, thus we are probably dealing with a submodule, so read it to resolve the real git dir */
+        /* It's a file — submodule or worktree. Read it to resolve the real git dir. */
```
vkareh: Avoid unnecessary re-wording of things.

src/file-manager/fm-icon-container.c
```diff
@@ -22,6 +22,7 @@
    Author: Michael Meeks <michael@ximian.com>
 */
 #include <config.h>
+#include <stdlib.h>
```
vkareh: What are you using this for?

vkareh (changes requested):
Caveat: I do not like this feature being part of core caja. It should live in caja-extensions. That said, the code is good (with some change requests).

A few additional requests:
- Remove the Anthropic ad from the commit messages (or at least change to `Assisted-by: Claude Sonnet 4.6` to avoid linking the Claude account to MATE commits)
- Reword the commit messages for human readability (it currently is too verbose with too many internal code details, rather than a concise description of the problem and how it solves it - look at the few latest commits in https://github.com/mate-desktop/caja/commits/master for examples)

src/file-manager/fm-icon-container.c
Inspirati: I didn't know g_canonicalize_filename() exists, as it seems to have been introduced in later versions of glib "available since Version 2.58". As I also don't know which maximum version of glib the project is targeted at, so stdlib's realpath() is the safe option.

g_canonicalize_filename() does not appear anywhere else in the codebase.

src/file-manager/fm-icon-container.c
Inspirati: realpath()

src/file-manager/fm-icon-container.c
Inspirati: I do not know, and it is beyond my scope to find out. I went with it because it looks good compared to the alternative options considered. If there is downstream feedback about it being a problem, it can be revisited later. In possible cases of them not rendering, at least it should fail safe with the tofu box.

Run-time determination of the available character set is beyond the scope of the feature introductory effort.

src/file-manager/fm-icon-container.c
Inspirati: Incorporated.

src/file-manager/fm-icon-container.c
Inspirati: Reverted.

src/file-manager/fm-icon-container.c
vkareh: Caja already requires GLib 2.58.1 (https://github.com/mate-desktop/caja/blob/master/configure.ac#L6)

Using the GLib function means that we no longer need to include `stdlib.h` above, and the code becomes a lot smaller and easier to understand. Also `realpath` does not appear anywhere else in the codebase either, so that's not a valid argument in this case.

src/file-manager/fm-icon-container.c
vkareh: Got it. Once you replace `realpath` with `g_canonicalize_filename`, then this can be removed.

src/file-manager/fm-icon-container.c
vkareh: Sure, but I expect more due diligence than "out of scope" when it's actually the purpose of the feature you're submitting.

src/file-manager/fm-icon-container.c
Inspirati: Incorporated change.

src/file-manager/fm-icon-container.c
Inspirati: Removed. realpath() and free() no longer referenced.

src/file-manager/fm-icon-container.c
robert-dickenson: Until the possibility of rendering issues being reported, the current implementation should fail safe. If such is reported further investigations can be undertaken.

vkareh: > Until the possibility of rendering issues being reported, the current implementation should fail safe. If such is reported further investigations can be undertaken.

You're proposing a "fix after merge" solution to a "test before merge" problem.

No.

Inspirati: I have already tested it and it works perfectly. What I am proposing is that if an issue is identified, then it shall be investigated - just like with any other software artefact.

lukefromdc: If we do "fix after merge" or even "test after merge" with code AI played any role in, people will lose trust in MATE. Bad enough to do that with our own code but it we do that with AI we break faith with millions who do not trust AI.

vkareh: > I have already tested it and it works perfectly. What I am proposing is that if an issue is identified, then it shall be investigated - just like with any other software artefact.

Clearly.

I was going by what you said earlier:
> I do not know, and it is beyond my scope to find out.

If you've tested this in a variety of standard fonts and they all work consistently, then that's the "due diligence" part I'm referring to.

My original review question was:
> Would these render in all systems, or would it show a tofu box instead?

Seems like you have a reasonable answer to that question then. You've tested across a variety of setups and did not encounter tofu boxes, yes?

lukefromdc: What happens if an intentionally bad font is thrown at it? If I were an attacker that's one of the things I would try. "Tofu boxes" if this is what I think it is would be OK, a segfault would not be.

Inspirati: Correct. I'm just disclosing that upon looking at the list of distributions on the Caja GitHub page https://github.com/mate-desktop/caja - I have not tried on many of these. For example, I'm not about to try configuring a build system under Cygwin in order to test for these rendering as tofo boxes or not. Another option may be to have the strings mapped into the translation system, but then I would think this is not really an issue of language translation, but rather of font support on all possible systems. When I implemented the original feature, I ended up spending way more time on the string translations than I did on the underlying code. And even then, I wasn't able to cover all the currently supported languages.

lukefromdc: To test on all distros is not what I'm asking and is certainly beyond my own capacity with no landline and no VM experience. Rather I would favor testing with a variety known good and then some known bad fonts to see what happens. That should cover most cases.

Same basic idea as testing a passphrase handling gui with a good passphrase, then a bad one, and finally with a whole page of text cut and pasted in while running from terminal. The last looks for buffer overflows in that case.

lukefromdc:
I thought all translations were supposed to be externaly handled originally at Transifex now at Weblate

PR #1896, opened 2026-06-12
Solve slow copy of big folders

dodobyte:
This is a long standing problem I've been personally struggling for years. The issue was undostack, ```add_origin_target_pair``` is called once per file and ```g_list_append``` walks the whole list every call.
Other users also noticed that it starts fast and gets slow down as expected. 

The solution is already in the documentation of [g_list_append](https://docs.gtk.org/glib/type_func.List.append.html).

> g_list_append() has to traverse the entire list to find the end, which is inefficient when adding multiple elements. A common idiom to avoid the inefficiency is to use g_list_prepend() and reverse the list with g_list_reverse() when all elements have been added.

So I switched to ```g_list_prepend```, as far as I know there's no side effect of having undostack in reverse order but I added code to reverse it as the documentation points anyway.

### DIFF
--- a/libcaja-private/caja-undostack-manager.c
+++ b/libcaja-private/caja-undostack-manager.c
@@ -748,6 +748,9 @@ caja_undostack_manager_add_action (CajaUndoStackManager * manager,
     return;
   }
 
+  action->sources = g_list_reverse (action->sources);
+  action->destinations = g_list_reverse (action->destinations);
+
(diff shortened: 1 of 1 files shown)


dodobyte:
[here's the video showing the result](https://www.youtube.com/watch?v=sSL4NpuArBo)

vkareh (approved):

PR #1897, opened 2026-06-21
Feat: Support tags

RamiZidan:


src/caja-tags-viewer.c
```diff
@@ -0,0 +1,736 @@
...
+    CajaTagsViewer *viewer;
+
+    /* Single-file selection only (parity with Notes) */
+    if (files == NULL || files->next != NULL)
+    {
+        return NULL;
+    }
+
+    file = CAJA_FILE_INFO (files->data);
+    uri = caja_file_info_get_uri (file);
```
vkareh, line 679: This is unused

src/caja-tags-viewer.c
```diff
@@ -0,0 +1,736 @@
...
+            {
+                g_free (l->data);
+                details->tags = g_list_delete_link (details->tags, l);
+                break;
+            }
+        }
+
+        rebuild_chips (self);
+
+        if (!details->loading)
```
vkareh, line 190: I don't think `loading` can ever be true outside of the `load_callback` function, so this is an unnecessary check...

vkareh (changes requested):
Can you write a proper commit message and a PR description please? Also, how can this be tested?

PR #1904, opened 2026-08-23
fm-tree-model: compare removed mount root by URI

### LINKED ISSUE #1903: Unmounted drive is not removed from tree view
### Steps to Reproduce

1. Mount a USB drive (e.g. USB stick or microSD card in a USB reader)
2. Switch to any folder on that drive in tree view
3. Unmount the drive via popup menu

### Expected Behavior

The drive node in the tree view gets removed.

### Actual Behavior

The drive node is still present.

### Error Output

```shell

```

### Version

1.26.4

### Distribution

Debian 13

### Additional Context

If I omit step 2, the issue doesn't occur.

### Reproducibility

Every time

### Checklist

- [x] I have searched existing issues to avoid duplicates
- [x] I have provided all the requested information above

basicmaster:
Due to a race condition, the corresponding file entry may already have been deleted. So the node of the unmounted device would be created as a new file again here, leading to two different files having the same URI. This way the loop comparison would never match then.

Fixes #1903.

### DIFF
--- a/src/file-manager/fm-tree-model.c
+++ b/src/file-manager/fm-tree-model.c
@@ -1774,17 +1774,15 @@ void
 fm_tree_model_remove_root_uri (FMTreeModel *model, const char *uri)
 {
     TreeNode *node;
-    CajaFile *file;
 
-    file = caja_file_get_by_uri (uri);
     for (node = model->details->root_node; node != NULL; node = node->next)
     {
-        if (file == node->file)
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
While I was not able at the moment to invoke the race condition, this didn't create any issues with mounting or unmounting anything.  MTP devices unmount but keep showing and can be remounted. Hoping this can solve some problems with unmounting encrypted devices from caja

PR #1909, opened 2026-09-11
Add Meson support

Sunderland93:


lukefromdc:
Got this error on trying to configure with meson:
`mate-submodules/meson.build:1:0: ERROR: Second call to project()`

Also: why are we defaulting to false on introspection? There are some 3ed party python applets out there that use it including Ubuntu's advanced menu

